PR #43, opened 2016-02-21
extensions.txt: fix inclusion of the vkEnumerateInstanceExtensionProperties validity block

derhass:
The include:: directive itself was transcribed into the PDF, instead of the included block. This adds missing line breaks after that directive.


oddhack:
This was also edited in the 1.0.4 spec update just pushed to github; it solved this problem but in a different way (rewriting some of the array query language for increased spec consistency). Closing this unless something else is still wrong here.

PR #50, opened 2016-02-21
Fix some typos

philiptaylor:


Vtec234:
As per @oddhack's comment on #3, you should probably split this, and possibly #51 as well, into smaller PRs pertaining to just a single section.

Issue #6, opened 2016-02-17
XML misc enum issues


mdsitton:
The following section of vk.xml has some issues if one was wanting to use the file to generate a binding for non C/C++ languages. Because the language target may not support it properly. Also technically it seems the LL extended literals came with C99 and C++11 so this *could* be an issue even with the target language if you have some odd compiler maybe? idk.

Plus its just weird that these values would need to be computed in C99/C++11 to get the proper value for them....
```
<enums name="API Constants" comment="Misc. hardcoded constants - not an enumerated type">
        <!-- This is part of the header boilerplate -->
    <enum value="256"   name="VK_MAX_PHYSICAL_DEVICE_NAME_SIZE"/>
    <enum value="16"    name="VK_UUID_SIZE"/>
    <enum value="256"   name="VK_MAX_EXTENSION_NAME_SIZE"/>
    <enum value="256"   name="VK_MAX_DESCRIPTION_SIZE"/>
    <enum value="32"    name="VK_MAX_MEMORY_TYPES"/>
    <enum value="16"    name="VK_MAX_MEMORY_HEAPS"/> <!-- The maximum number of unique memory heaps, each of which supporting 1 or more memory types. -->
    <enum value="1000.0f" name="VK_LOD_CLAMP_NONE"/>
    <enum value="(~0U)" name="VK_REMAINING_MIP_LEVELS"/>
    <enum value="(~0U)" name="VK_REMAINING_ARRAY_LAYERS"/>
    <enum value="(~0ULL)" name="VK_WHOLE_SIZE"/>
    <enum value="(~0U)" name="VK_ATTACHMENT_UNUSED"/>
    <enum value="1"     name="VK_TRUE"/>
    <enum value="0"     name="VK_FALSE"/>
    <enum value="(~0U)" name="VK_QUEUE_FAMILY_IGNORED"/>
    <enum value="(~0U)" name="VK_SUBPASS_EXTERNAL"/>
</enums>
```

oddhack:
We did consciously restrict the Vulkan header to modern C/C++ compilers, after considerable polling of IHVs and developers supporting different platforms and compilers, The contortions we go through in the GL XML/headers to deal with 64-bit types were a strong discincentive.

[...]

PR #75, opened 2016-02-24
[WIP/RFC] Generator for Julia bindings

vchuravy:
This is my current attempt to generate Julia bindings from vk.xml. This is still in progress and before it is mainlined it needs to be better documented and cleaned-up.

One of the big issue I am facing right now are Union types that are used in Vulkan, especially `VkClearValue` (the two union members have different size).
As pointed out in https://github.com/KhronosGroup/Vulkan-Docs/issues/6#issuecomment-187889611 we should split these community provided bindings into their own file. 
CC: @oddhack 

The generated code and Julia bindings can be found at https://github.com/JuliaGPU/Vulkan.jl/pull/1


vchuravy:
We generate Julia code from the C headers, so this is no longer needed

PR #108, opened 2016-02-29
vk.xml: add structure to type[category='funcpointer'] elements

Kangz:
PTAL

These elements' structure now matches the structure of the ```<command>```
elements with a ```<proto>``` child and ```<param>``` children. This will make it
easier for third-party parser to understand function pointer types.

generator.py has been modified to understand the updated definitions.

There are no changes (apart from a newline) to vulkan.h. However typedef for function pointers in the docs has its VKAPI_PTR removed, which I think is an improvement as it matches the fact that Vulkan commands don't have the VKAPI_FOO in the docs.


Kangz:
cc @oddhack

doc/specs/vulkan/validity/protos/vkCmdCopyBuffer.txt
```diff
@@ -16,8 +16,8 @@ endif::doctype-manpage[]
 * This command must: only be called outside of a render pass instance
 * The value of pname:regionCount must: be greater than `0`
 * Each of pname:commandBuffer, pname:srcBuffer and pname:dstBuffer must: have been created, allocated or retrieved from the same sname:VkDevice
-* The sum of the pname:srcOffset and pname:size members of a given element of pname:pRegions must: be less than or equal to the size of pname:srcBuffer
-* The sum of the pname:dstOffset and pname:size members of a given element of pname:pRegions must: be less than or equal to the size of pname:dstBuffer
+* The sum of the pname:srcOffset and pname:copySize members of a given element of pname:pRegions must: be less than or equal to the size of pname:srcBuffer
```
Kangz: Note: this is not an intended modification, but leftovers from a previous commit that did not ```make full_install```

PR #142, opened 2016-03-13
Avoid double negation in phrase.

nsf:
To me it seems like double negation here is harder to understand. And reader is mostly interested in things being used and not the things being unused. But also mentioning that they are both used and ignored implies that you need to put elements corresponding to each attachment and not just used ones.


oddhack:
This should be fixed in the 1.0.12 update. We're still sorting through some arcane issues around contributor license agreements, so rather than accepting the PR I did something equivalent internally - hopefully this will be resolved shortly.

PR #144, opened 2016-03-16
Fix comments of VkResult constants.

diaphore:
`VK_ERROR_INITIALIZATION_FAILED` and `VK_ERROR_DEVICE_LOST` comments are inverted.


oddhack:

PR #150, opened 2016-03-22
Fix prototype guard in documentation

1ace:
This was changed before the release, so as far as everyone knows, this has always been the other way around: define `VK_NO_PROTOTYPES` when you do *not* want the prototypes.


oddhack:

1ace:
Would it help if I sent my patches as MR directly on gitlab.khronos.org? There shouldn't be any CLA issue there.

oddhack:
Yes, that would be better until we get the CLA in place.

PR #151, opened 2016-03-22
Correct typo in comment on VK_INCOMPLETE

expipiplus1:
`s/resul/result/`


oddhack:

PR #168, opened 2016-04-07
Fix fname:: -> fname: (Remove extra colon)

MajorBreakfast:
I assume that colon snuck in there by accident.


courtney-g:
+1

1ace:
There's an other one in [vkAllocateCommandBuffers.txt:28](https://github.com/KhronosGroup/Vulkan-Docs/blob/1.0/doc/specs/vulkan/man/vkAllocateCommandBuffers.txt#L28) :+1:

MajorBreakfast:
@1ace Thx for spotting this.

I did a quick text search and these are all occurances of `fname::`.

oddhack:

PR #173, opened 2016-04-10
update doc, change "host local, host visible" -> "host local, device …

ttyio:
…visible"


pimanttr:
-1 this was correct.

VkDeviceMemory is always device-visible. Host-local doesn't necessarily imply host-visible, this wording was to indicate we're talking about host-visible memory.

ttyio:
thanks for the explanation. closing

PR #180, opened 2016-04-15
vk.xml: add structure to type[category='funcpointer'] elements

Kangz:
(reopening the #108 PR as I messed up my Github-fu)

These elements' structure now matches the structure of the <command>
elements with a <proto> child and <param> children. This will make it
easier for third-party parser to understand function pointer types.

generator.py has been modified to understand the updated definitions.

There are no changes (apart from a newline) to vulkan.h. However typedef for function pointers in the docs has its VKAPI_PTR removed, which I think is an improvement as it matches the fact that Vulkan commands don't have the VKAPI_FOO in the docs.

@oddhack PTAL, this was rebased on top of 1.0.10 and AFAIK my employer (Google) has a CLA in place.

### DIFF
--- a/doc/specs/vulkan/funcpointers/PFN_vkAllocationFunction.txt
+++ b/doc/specs/vulkan/funcpointers/PFN_vkAllocationFunction.txt
@@ -7,7 +7,7 @@ ifdef::doctype-manpage[]
 ["source","{basebackend@docbook:c++:cpp}"]
 endif::doctype-manpage[]
 ------------------------------------------------------------------------------
-typedef void* (VKAPI_PTR *PFN_vkAllocationFunction)(
+typedef void* (*PFN_vkAllocationFunction)(
     void*                                       pUserData,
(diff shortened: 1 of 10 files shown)


krOoze:
@Kangz It makes sense in the `vk.xml` (`funcptr` attribute already should imply VKAPI_PTR).

`VKAPI_FOO` were recently documented in the spec.

As an user I would like to still have `VKAPI_PTR` decorated function pointers in the spec, so I can declare functions I pass to it via them correctly.

TobiasHector:
Looks like this is from way back before we added the CLA - @Kangz are you still interested in pushing this forward? If you wanted to sign the CLA and rebase it, I'd be happy to review the change.

Kangz:
Thanks for the follow-up. I don't need this change anymore and will not push it forward. If someone wants to re-use the code they should feel free to do so, the CLA should be ok since my company (Google) is a Khronos member.

PR #181, opened 2016-04-16
fix typo in vkGetPhysicalDeviceQueueFamilyProperties manpage

Marqin:
fix two typos in vkGetPhysicalDeviceQueueFamilyProperties manpage:
transsfer -> transfer
an -> and


oddhack:

Marqin:
No problem, thanks for your reply :smile:

PR #186, opened 2016-04-20
Fix invalid parameter reference in vkEnumeratePhysicalDevice description

Overv:
The number of devices written by `vkEnumeratePhysicalDevices` is returned using the integer pointed to by `pPhysicalDeviceCount`, but the description states that the number is written to `pPhysicalDevices`.


oddhack:

Issue #26, opened 2016-02-19
Device memory type ordering


philiptaylor:
Section 10.2 (Device Memory) says:

> Memory types are ordered in the list such that X is assigned a lesser memoryTypeIndex than Y if X≤Y according to the partial order.

That implies X is always assigned a lesser memoryTypeIndex than X (because X <= X), which is impossible. It should use a strict partial order instead, i.e. say X is assigned lesser than Y if X < Y (which is equivalent to saying X <= Y && !(Y <= X) using the earlier definition of '<=').

> Note that the list of all allowed memory property flag combinations above satisfies this partial order

No it doesn't. E.g. "VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT" is a subset of "VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT | VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT" but comes later in that list.

Rearranging the list doesn't sound like a good solution, because then the slower non-DEVICE_LOCAL ones would be preferred over the DEVICE_LOCAL ones. Maybe the definition of the order should be changed to ignore DEVICE_LOCAL in the bits comparison (and rely on the performance clause to prefer the DEVICE_LOCAL ones).

> Given two memory types X and Y, the partial order defines X≤Y if:

I think mathematically it's a preorder, not a partial order, under the widely-accepted definitions of those terms (e.g. http://mathworld.wolfram.com/PartialOrder.html, http://mathworld.wolfram.com/Preorder.html).

You could have two distinct memory types X != Y (e.g. they have different heapIndexes, or the same heapIndex and just different memoryTypeIndex) that have the same propertyFlags and the same performance, so that X <= Y and Y <= X. A partial order is defined to have antisymmetry, i.e. X <= Y && Y <= X implies X = Y, which is not true in this case. A preorder is a partial order without antisymmetry, which means two elements can be equivalent but non-equal, which is what we want here.

...But maybe that's possibly being too pedantic about terminology.

jeffbolznv:
> Rearranging the list doesn't sound like a good solution, because then the slower non-DEVICE_LOCAL ones would be preferred over the DEVICE_LOCAL ones.

[...]

PR #194, opened 2016-04-25
Suggestions for ch10.2.0 Device Memory

krOoze:
I suggest some text changes/rewrite to the ch. 10.2.0 Device Memory (as a PR):

Non-semantic changes:

1) Random grammar and formatting

2) Make example in first paragraph more straightforward with elementary Vulkan primitives (the original also seemed factually wrong in places; e.g. does UBO really have to be in "on-device" memory?)

3) Removed unnecessary brandishing of obscure math describing the memory type list order. Tried something equally formal, but more readable.

4) Made a note into a Note

5) Updated example. Function names seem to be lower-case in the spec. Hopefully made it bit
more readable. Turned the acompanying note into the example content and inline comments.

6) Removed redundant `VkDeviceMemory` intro

7) Break out specifying language out of  a Note

8) More straightforward description of `memoryTypeIndex`

9) Fix pdf formatting of the memory types property list

<hr/>

Semantic changes:

1) Make explicit there is a <=> relationship between `DEVICE_LOCAL` heap and `DEVICE_LOCAL`  memory type

2) Make >4GB runtime error report mandatory. Otherwise it basically says "do not pass >= 4GB, because possible undefined behavior".

3) Removed `VUID-vkAllocateMemory-device-00636`. Behavior is deterministic returning `VK_ERROR_TOO_MANY_OBJECTS`

4) Moved `allocationSize` VU (depends on non-local state).

5) Changed `allocationSize`. Not sure what the original means, so improvized a new one saying alloc size must be <=  the whole size of the heap. Should potentially be removed altogether (looks more like something better covered by `VK_ERROR_OUT_OF_DEVICE_MEMORY`)?

6) Added VU restricting `memoryTypeIndex` to only the memory types of the device.


philiptaylor:
I think that with your phrasing:

> [...] it holds, that X occupies a lower index position than Y, if [C1] or if [C2]

it's unclear whether it means "(X is lower than Y) implies (C1 or C2)" or "(C1 or C2) implies (X is lower than Y)" or both.

There may be different ways of phrasing it in plain English that would be clear enough. But I think it's difficult to find exactly the right phrasing, and plain English can be quite fragile - a future edit might accidentally change the way people interpret the sentence and the error would be hard to spot.

A mathematical definition has a much better chance of being unambiguous and not open to interpretation. I think the downside is that a mathematical definition really needs to be a precisely correct mathematical definition (hence my complaints in #26 about partial order vs preorder etc), which can make it quite complicated, and if readers aren't familiar with set theory then I can imagine it's hard to read.

(I have no idea what level of mathematical background is expected from the specification's audience, but I suspect set theory isn't a common requirement for graphics programming.)

> 2) Wouldn't it be better/clearer to remove the "or equal" and consequently the ∧ ¬(Y ≤ X)?

Yeah, I think it should work if you make that change and say it's defining a strict partial order "X < Y" instead of a preorder "X ≤ Y", and that might be a bit simpler.

> ( Even my AMD impl. seems to have it sorted wrong - or I horribly misunderstood it)

AMD seems to use the same order as the list of valid propertyFlags in the pre-1.0.9 spec, so I'd guess they copied it from that, but that list was wrong, so they violate the order requirement. NVIDIA also seems to use the same order as that wrong list, but (perhaps by luck) their chosen types happen to be compatible with the order requirement. I think that suggests the problem was primarily caused by the spec being self-contradictory (fixed in 1.0.9), not by the implementers misunderstanding the maths.

> 6) Description of the code block of memory type find is a lie (unless I am horribly wrong about the preorder of course)

I think the bit about "(i.e. a memory type with the fewest additional bits set)" is technically wrong, since it implies counting the number of bits - it should be more like "(i.e. a matching memory type that is not a strict superset of any other matching memory type, and does not have worse performance than any other memory type with exactly the same bits)" (or remove that attempt at clarification entirely since it's making things more confusing).

krOoze:
> it's unclear whether it means "(X is lower than Y) implies (C1 or C2)" or "(C1 or C2) implies (X is lower than Y)" or both.

My version uses the same form of implication as is everywhere around the spec. It is pretty canonical form generaly everywhere else, really. "If A then B". Or equivalently "B, if A".

"Both ways" would be "iff".  As for (X<Y) → (C1∨C2), I don't see how it could be interpreted that way. I need you to be more specific.

> and plain English can be quite fragile

> A mathematical definition has a much better chance of being unambiguous and not open to interpretation.

Agreed. Though I WAS aiming for a mathematical definition (don't be mistaken by the apparent plesantness and lack of integrals and tensors). If it's not, then it is an unintentional failure on my part.

I just tried to downgrade it to elementary school math (and perhaps make it more terse at the same time - less surface area for things to be/stay wrong or be broken in the future.

> AMD seems to use the same order as the list of valid propertyFlags in the pre-1.0.9 spec.

The driver IS 1.0.8, so it makes half-sense. That they skipped the math altogether (and blindly used the list order, which has lower spec power of a "note"), seems to me even worse a symptom of unreadability.

> I think the bit about "(i.e. a memory type with the fewest additional bits set)" is technically wrong, since it implies counting the number of bits

Yes, I meant exactly that. Plain wrong, not just technicaly :) Entry with more additional bits may be before the one with less. No minimization is occuring ( e.g. DEV_LOC requested, DEV_LOC | H_V | H_C = +2 bits may be before DEV_LOC | LAZY = +1 bit ).

Also I was careful with the "not detrimental to the behaviors", due to what I see now is another oversight in the spec I should just report instead.

krOoze:
@oddhack This PR brings up some correctness issues with the spec. Should I break them up (and out) and post in Issues, or is this sufficient?

Food for thought: Would it be worthwhile to standardize "if" and "only if" similarly as the "must" and others to eliminate any remaining doubt?

philiptaylor:
> "If A, then B". Or equivalently "B if A".

I don't think those are clearly equivalent. In a normal English context (where you aren't expecting precise mathematical language), it'd be reasonable to say [something like](https://www.google.com/search?q="n+is+even+if+it") "n is even if it is a multiple of 2" which should be interpreted as a complete definition of the term 'even', i.e. as an if-and-only-if relation. That's different to a phrase like "if n is a multiple of 2, then it is even", which (at least to me) does not sound like a natural English way of defining a term, and therefore sounds more like a mathematical statement (therefore defining a one-way relation). It's a subtle distinction but I think it makes an important difference here.

> Yes, I meant exactly that. Plain wrong, not just technicaly :)

Well, it sort of has the right spirit in that the code will prefer 'smaller' (in some vague sense) sets of bits. But I agree, technically wrong is wrong :-)

krOoze:
Admittedly, I am not an Englishman, but I never considered "if" in your way in any context formal or informal.

> >"If A, then B". Or equivalently "B if A".

>I don't think those are clearly equivalent.

To me, those two versions do sound identical (apart from aesthetics). 

I looked up the grammar (for the last commits) and it says the comma means the sentence is reordered. I deduct, there is no difference of semantics.

>  it'd be reasonable to say [something like](https://www.google.com/search?q=%22n+is+even+if+it%22) "n is even if it is a multiple of 2" which should be interpreted as a complete definition of the term 'even'

Sorry, it does really sound to me as a one-way relation. The googled pages seem to also define the set complement (odd numbers) and so too treat this first "if" as an incomplete description.

It **would** be unpolite though, to keep the reader guessing/deciphering, whether the "if" also happen to be exhaustive, without good reason (it's implementation-dependent in our case == good reason). So I suggested to consider standardisation.

It is little bit too late to second-guess the "if"ses. The spec is full of them. I don't think rewriting it all in logic formulas would help anything. Most (all?) of the usages are quite simple, and this one is not so bad either.

> Well, it sort of has the right spirit in that the code will prefer 'smaller' (in some vague sense) sets of bits.

Only for the alternatives with subset relation.

Well, if the description of the order is made more readable (without sacrificing the exactness of course), then perhaps the example description can be shortened (or removed altogether)...?

oddhack:
We're finally ready to review this, now that the CLA signed and we've tweaked our internal process some. Could you bring the PR up to date with changes since then?

krOoze:
@oddhack Updated.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -446,9 +446,9 @@ endif::VK_EXT_validation_cache[]
 [[memory-device]]
 == Device Memory
 
-Device memory is memory that is visible to the device, for example the
-contents of opaque images that can: be natively used by the device, or
-uniform buffer objects that reside in on-device memory.
+_Device memory_ is a memory that is visible to the device. For example the
```
TobiasHector: "memory" is a mass noun, so the article "a" should not be added.

TobiasHector: I don't think these should be split into two sentences - the example sort of clarifies the first sentence. I'd maybe put a semi-colon between them though, rather than a full stop or a comma as was previously there.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
-    performance (as determined in an implementation-specific manner).
-
-Memory types are ordered in the list such that X is assigned a lesser
-pname:memoryTypeIndex than Y if [eq]#(X {leq} Y) {land} {lnot} (Y {leq} X)#
-according to the preorder.
-Note that the list of all allowed memory property flag combinations above
-satisfies this preorder, but other orders would as well.
-The goal of this ordering is to enable applications to use a simple search
-loop in selecting the proper memory type, along the lines of:
+The array pname:memoryTypes must: be ordered in such a way that for each of
```
TobiasHector: nit: I'd probably just start this sentence "pname:memoryTypes must: be..." - "The array" is just extra unnecessary verbage, and is a bit jarring as it's inconsistent with the rest of the spec.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
+its distinct memory types X and Y it holds that X occupies a lower index
```
TobiasHector: Elsewhere in the spec we've tagged identifiers like this with bold (e.g. `*X*`) so that it stands out as an identifier. Obviously an existing problem, but would be good if you could fix it.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
+position than Y if:
```
TobiasHector: Perhaps it would make more sense to say this as:

```
For any two array elements *X* and *Y* returned in pname:memoryTypes, *X* must: be at a lower index than *Y* if:
```

Mostly it's just shorter, which usually aids readability. (Though your paragraph is still way better than the previous one IMO)

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
+
+  * either the pname:propertyFlags of X (interpreted as a set of raised flags) is a
+    strict subset of the pname:propertyFlags of Y.
```
TobiasHector: I would perhaps word this as:

```
Either the set of bit flags returned in the pname:propertyFlags member of *X* is a strict subset of the set of bit flags returned in the pname:propertyFlags member of *Y*
```

Saying "interpreted as a set of raised flags" is a bit confusing - would much prefer the two sets be explicitly stated.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
+    implementation-specific manner).
+
+[NOTE]
+.Note
+====
+There might be more than one conformant way to order given set of memory
+types. Notice that the list of all allowed memory property flag
+combinations above is written in one such of several possible ways that do
+satisfy the prescribed order.
+====
```
TobiasHector: I'm not sure what this note is attempting to clarify. Your new spec text makes the situation a lot clearer - the only reordering that can take place is that of memory types with equivalent propertyFlags. This note seems to be suggesting there are other options and actually makes things less clear IMO... unless I'm missing something?

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
+
+The purpose of such ordering is to enable applications to use a simple search
+loop to select the desired memory type along the lines of:
```
TobiasHector: nit: Perhaps this alternative sentence start?

```
This ordering is provided to enable applications to...
```

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
-// Find a memory type in "memoryTypeBits" that includes all of "properties"
-int32_t FindProperties(uint32_t memoryTypeBits, VkMemoryPropertyFlags properties)
-{
-    for (int32_t i = 0; i < memoryTypeCount; ++i)
-    {
-        if ((memoryTypeBits & (1 << i)) &&
-            ((memoryTypes[i].propertyFlags & properties) == properties))
-            return i;
+// Find a memory type in "memoryTypeBitsRequirement" that includes at least all of "requiredProperties"
+int32_t findProperties(const VkPhysicalDeviceMemoryProperties* pMemoryProperties, uint32_t memoryTypeBitsRequirement, VkMemoryPropertyFlags requiredProperties) {
```
TobiasHector: Existing problem, but it should be VkMemoryPropertyFlag**Bit**s

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
+    for (int32_t mi = 0; mi < pMemoryProperties->memoryTypeCount; ++mi) {
```
TobiasHector: Might just be me, but I don't really like "mi" - either use a full name as per the Vulkan API, or just "i". "mi" is a weird middle ground that isn't explained.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,57 @@ There must: be at least one memory type with the
...
 }
 
-// Try to find an optimal memory type, or if it does not exist
-// find any compatible memory type
+// Try to find an optimal memory type, or if it does not exist try fallback memory type
+extern VkDevice device;
+extern VkImage image;
+extern VkPhysicalDeviceMemoryProperties memoryProperties;
+VkMemoryPropertyFlags requiredProperties = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+VkMemoryPropertyFlags optimalProperties = requiredProperties | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
```
TobiasHector: The reason these weren't in the original is so that it could be copy+pasted into an app and "just work". These declarations remove that ability. If you want them noted (which is reasonable), I suggest you put them in a comment block.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -775,29 +772,29 @@ same memory object.
...
+the error ename:VK_ERROR_OUT_OF_DEVICE_MEMORY must: be returned.
 
 .Valid Usage
 ****
-  * [[VUID-vkAllocateMemory-device-00636]]
-    The number of currently valid memory objects, allocated from
-    pname:device, must: be less than
-    sname:VkPhysicalDeviceLimits::pname:maxMemoryAllocationCount
+  * pname:pAllocateInfo::pname:allocationSize must: be less than or equal to
+    sname:VkPhysicalDeviceMemoryProperties::pname:memoryHeaps[pname:pAllocateInfo::pname:memoryTypeIndex].pname:size
```
TobiasHector: nit: Should be `slink:`

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -775,29 +772,29 @@ same memory object.
...
+    as returned by flink:vkGetPhysicalDeviceMemoryProperties for physical
+    device parent of pname:device
```
TobiasHector: I appreciate the extra clarity by identifying the parent, but in practice we don't do this anywhere else in the spec so it feels out of place. I'm not too unhappy with keeping it if you really want it though, but then I'd suggest something like:

```
...for the slink:VkPhysicalDevice which pname:device was created from.
```

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -775,29 +772,29 @@ same memory object.
...
+  * pname:pAllocateInfo::pname:memoryTypeIndex must: be less than
+    sname:VkPhysicalDeviceMemoryProperties::pname:memoryTypeCount
+    as returned by flink:vkGetPhysicalDeviceMemoryProperties for physical
+    device parent of pname:device
```
TobiasHector: Comments on the above VU apply equally to this one.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -813,9 +810,9 @@ include::../api/structs/VkMemoryAllocateInfo.txt[]
   * pname:sType is the type of this structure.
   * pname:pNext is `NULL` or a pointer to an extension-specific structure.
   * pname:allocationSize is the size of the allocation in bytes
-  * pname:memoryTypeIndex is the memory type index, which selects the
-    properties of the memory to be allocated, as well as the heap the memory
-    will come from.
+  * pname:memoryTypeIndex is an index selecting a memory type from the
```
TobiasHector: nit: suggest "identifying" rather than "selecting" - that's what we use elsewhere in the spec.

doc/specs/vulkan/chapters/memory.txt
krOoze: >"memory" is a mass noun, so the article "a" should not be added.

Ah. Again, feels a bit naked without some quantifier or specifier... perhaps "is a memory type"?

>I don't think these should be split into two sentences

Right. Comma felt wrong to me. Maybe semi-colon or em-dash then.

doc/specs/vulkan/chapters/memory.txt
krOoze: Or was it changed to `[eq]`? Should probably be in the styleguide...

doc/specs/vulkan/chapters/memory.txt
krOoze: Was that before. I did not like that because it make the sentence start with lower-case.

TobiasHector: Yea it's a bit weird, but it's an identifier so it's fine - we do this all over the place in the spec.

doc/specs/vulkan/chapters/memory.txt
krOoze: I suppose that makes sense. "be at index" feels informal.

> (Though your paragraph is still way better than the previous one IMO)

That was the initial point. Then it got out of hand...

doc/specs/vulkan/chapters/memory.txt
TobiasHector: I usually use [eq] if I'm going to do any actual equations, but if it's just using it to identify stuff in a body of text I'll just bold it. Tagging it with "[eq]##" just feels too heavyweight IMO. Either is better than just plain text though :)

doc/specs/vulkan/chapters/memory.txt
TobiasHector: > Ah. Again, feels a bit naked without some quantifier or specifier... perhaps "is a memory type"?

Saying "type" will confuse it with... well.. memory types. I don't think "...is memory that is visible to the device." is particularly weird though - it's grammatically correct.

I mean honestly, the problem is probably that the sentence is circular:

> Device memory is ... **memory** visible to the **device**

...but I think we need a proper memory model before we can fix that aspect of it. For now the best we can do is make sure it's grammatically accurate and adds some amount of information, I think.

doc/specs/vulkan/chapters/memory.txt
krOoze: Sure.
I wonder if it is necessary to explain. It is a name of the type after all.

doc/specs/vulkan/chapters/memory.txt
TobiasHector: > "be at index" feels informal.

Yea true, there's similar wording elsewhere in the spec that's equally awkward, not sure what a better wording would be though...? Well, maybe:

```
...be placed at a lower index within pname:memoryTypes than *Y*...
```

Not sure if that's better but maybe?

doc/specs/vulkan/chapters/memory.txt
krOoze: It tries to say the order is not unique. I.e. the same list of memory properties can be ordered validly more than one way.
The utility is to warn applications against expecting exactly the order in the list above.

TobiasHector: Oh because if for two sets of memory types and one is not a subset of the other, then the ordering between them is arbitrary? Right?

If that's it, maybe you could call that out in this note? Think that would help.

doc/specs/vulkan/chapters/memory.txt
krOoze: nit: what about "... enables ...".

doc/specs/vulkan/chapters/memory.txt
krOoze: I don't think so. More than one is allowed. Usually the API uses `*Flags` in that case.

doc/specs/vulkan/chapters/memory.txt
TobiasHector: As in "This ordering enables applications to..."?

Yes, less words meaning the same thing is always better :)

doc/specs/vulkan/chapters/memory.txt
TobiasHector: Ugh you're right, I totally flipped in my head which way around they were - whoops.

doc/specs/vulkan/chapters/memory.txt
krOoze: Means `m_i` which is tricky in cammel-case style.
Probably full name. The already long `pMemoryProperties->memoryTypes[mi].propertyFlags` could use its own name and line anyway.

doc/specs/vulkan/chapters/memory.txt
krOoze: Sure.

doc/specs/vulkan/chapters/memory.txt
krOoze: IMO `*link` should be everywhere. So useful.

doc/specs/vulkan/chapters/memory.txt
krOoze: You don't? E.g. `VkBufferCreateInfo`:

>If sharingMode is VK_SHARING_MODE_CONCURRENT, each element of pQueueFamilyIndices must be unique and must be less than pQueueFamilyPropertyCount returned by either vkGetPhysicalDeviceQueueFamilyProperties or vkGetPhysicalDeviceQueueFamilyProperties2KHR for the physicalDevice that was used to create device

Should probably copy-past it thought to match...

doc/specs/vulkan/chapters/memory.txt
krOoze: Do you? Can't find anything exept two occurences, which are struct output from the Implementation though.

krOoze: Probably mean "determines"?

doc/specs/vulkan/chapters/memory.txt
TobiasHector: Yea the style guide *should* actually say that now - but people still miss it. The only time you don't use it now is when you're within the thing you'd be linking to (i.e. the description of VkPhysicalDeviceProperties would have "sname:VkPhysicalDeviceProperties".)

doc/specs/vulkan/chapters/memory.txt
TobiasHector: Not *usually*. That one's a bit weird :P

Change the wording though and I'm happy :)

doc/specs/vulkan/chapters/memory.txt
TobiasHector: No I definitely meant identify. Again we don't always use it, and it's not always in the context of an index, but it's in the spec. Search for "identif" and you should find a number of uses.

doc/specs/vulkan/chapters/memory.txt
krOoze: Should have perhaps been `Bit` without "s", but what you gonna do now that the spec is released....

krOoze:
Updated w/ suggestions.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,70 @@ There must: be at least one memory type with the
...
+    pname:propertyFlags member of *Y*.
+  * or the pname:propertyFlags members of *X* and *Y* are equal, and *X*
+    belongs to a memory heap with greater performance (as determined in an
+    implementation-specific manner).
+
+[NOTE]
+.Note
+====
+The ordering rules above could allow more than one possible way to order the
+same set of memory types.
```
TobiasHector: Hmm, I kind of wonder if we should make this more referred to the statement above. Perhaps replacing the first sentence here with something like:

> There is no ordering requirement between memory types where the neither set of bit flags is a subset of the other, allowing more than one possible way to order the same set of memory types in this case.

?

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,58 +552,70 @@ There must: be at least one memory type with the
...
-    for (int32_t i = 0; i < memoryTypeCount; ++i)
-    {
-        if ((memoryTypeBits & (1 << i)) &&
-            ((memoryTypes[i].propertyFlags & properties) == properties))
-            return i;
+// Find a memory in `memoryTypeBitsRequirement` that includes all of `requiredProperties`
+int32_t findProperties(const VkPhysicalDeviceMemoryProperties* pMemoryProperties,
+                       uint32_t memoryTypeBitsRequirement,
+                       VkMemoryPropertyFlags requiredProperties) {
+    for (uint32_t mem_index = 0; mem_index < pMemoryProperties->memoryTypeCount; ++mem_index) {
```
TobiasHector: I don't care too much, so I won't block on this change, but it's odd to see mem_index (using an underscore) and everything else being camel case. Perhaps just use "memoryIndex"? Seems more consistent.

TobiasHector:
Couple extra bits of feedback, then I think we're good to go with this one.

doc/specs/vulkan/chapters/memory.txt
krOoze: Alright, I tried something along these lines.

doc/specs/vulkan/chapters/memory.txt
krOoze: Nah, just forgot which style I am writing in.
It was "mem" to fit the screen width. I am trying to shorten it now by making another constant instead...

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -500,48 +500,48 @@ different properties.
...
 characteristics, they may: all include
 ename:VK_MEMORY_HEAP_DEVICE_LOCAL_BIT.
-In a unified memory architecture (UMA) system, there is often only a single
+In a unified memory architecture (UMA) system there is often only a single
 memory heap which is considered to be equally "`local`" to the host and to
 the device, and such an implementation must: advertise the heap as
 device-local.
 
-
+[[memory-device-bimask-list]]
```
TobiasHector: Oh. should this be "bi**t**mask"???

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,57 +552,74 @@ There must: be at least one memory type with the
...
-  * the memory property bits set for X are a strict subset of the memory
-    property bits set for Y. Or,
-  * the memory property bits set for X are the same as the memory property
-    bits set for Y, and X uses a memory heap with greater or equal
-    performance (as determined in an implementation-specific manner).
+[NOTE]
+.Note
+====
+There is no ordering requirement between *X* and *Y* that are not in a
+subset relation, which potentially allows more than one possible way to
```
TobiasHector: "...that are not in a subset relation" reads weirdly to me - how about "...where neither is a subset of the other"?

krOoze: Dunno, it is a mouthful.
It also does not make that much sense literally. **If** there was a requirement "where neither is a subset of the other", then there would be zero possible orders.

I.e.
If I added requirement:
- or the `propertyFlags` members of **X** and **Y** are neither subset of the other

I would not be able to sort (**a**,**b**) array, because **a** has to be before **b** and at the same time **b** has to be before **a**.

It probably needs to be reworded differently for the extra word-smithing points...

doc/specs/vulkan/chapters/memory.txt
krOoze: Indeed.

TobiasHector:
I'm happy enough with these changes, think they're a useful addition, but would like others to at least skim over it and ideally recheck the changes to the code example.

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,57 +552,75 @@ There must: be at least one memory type with the
...
+That potentially allows more than one possible way to order the same set of
+memory types.
+Notice that the
+<<memory-device-bitmask-list,list of all allowed memory property flag combinations>>
+is written in the required order.
+But if instead
+ename:VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT was before
+ename:VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT |
+ename:VK_MEMORY_PROPERTY_HOST_COHERENT_BIT,
+such list would also be in the required order.
```
janharaldfredriksen-arm: "such a list" - or perhaps "the list"

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -552,57 +552,75 @@ There must: be at least one memory type with the
...
 }
 
-// Try to find an optimal memory type, or if it does not exist
-// find any compatible memory type
+// Try to find an optimal memory type, or if it does not exist try fallback memory type
+// `device` is the VkDevice
+// `image` is the VkImage that requires memory to be bound
+// `memoryProperties` properties as returned by vkGetPhysicalDeviceMemoryProperties
+// `requiredProperties` are the property flags that must be present
+// `optimalProperties` are the property flags that are preferred by application
```
janharaldfredriksen-arm: "by the application"

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -774,29 +790,30 @@ same memory object.
...
 ****
-  * [[VUID-vkAllocateMemory-device-00636]]
-    The number of currently valid memory objects, allocated from
-    pname:device, must: be less than
-    sname:VkPhysicalDeviceLimits::pname:maxMemoryAllocationCount
+  * pname:pAllocateInfo\->pname:allocationSize must: be less than or equal
+    to
+    slink:VkPhysicalDeviceMemoryProperties::pname:memoryHeaps[pname:pAllocateInfo\->pname:memoryTypeIndex].pname:size
+    as returned by flink:vkGetPhysicalDeviceMemoryProperties for the
+    slink:VkPhysicalDevice that pname:device was created from.
```
krOoze: If I may direct your attention, I wasn't sure about this VU:
 - not sure if it matches the intent of the original VU below
 - not sure if it is not handled by `VK_ERROR_OUT_OF_DEVICE_MEMORY` instead

TobiasHector: I do think it matches the intent of the removed VU.

I also think it is subtly different to the error message and worth preserving, since removing it might constitute a (albeit minor) functionality change.

TobiasHector: (It's *possible* that someone has set their allocator up to only handle allocations of sizes less than that, and runs into undefined territory due to e.g. integer overflows - unlikely - but technically possible)

PR #208, opened 2016-05-05
Fix styleguide build

krOoze:
Styleguide build was missing output directory existence test (and so make fails if build first).


krOoze:

PR #211, opened 2016-05-09
Faster MathJax configuration

krOoze:
Experimented with Mathjax configuration and this renders faster. Kept the chunked version unmodified for now though(as it is in short, quick to render pages)
My baseline time for the original (for page reload) is 75 s.

Changes:
- Removed MathML input(`MML` config and `content-mathml.js`), since only TeX math is used by the spec

- Removed `orMML` output. MML is almost never chosen and the script doing this automaticaly seems to be deprecated

- Swiched to `CHTML` which sheds about 10 s. Tradeoff is support for some old browsers.
Alternative could be `SVG` with another ~3 s improvement. Fastest would be the `MathML` setting (20 s off baseline), but they claim browser support is unreliable. `Preview` version is similarly fast.

- Disabled fast preview (renders preview, then the final version). It takes almost as much time as the final version and so not worth it. It also jerks the page by being replaced by the (slightly larger) final version and hides further MathJax % progress and uses CPU (so e.g. scrolling is slow/jerky), so not very usable. Sheds ~20 s

- Disabled AssistiveMML (supposedly adds better alternative code for readers). Sheds ~7 s. It can be re-enabled (as the other configurations) in a RMB context menu (the browser seems to remember these user settings).

Ended up with 35 s time.

### DIFF
--- a/doc/specs/vulkan/config/docbook-xsl/xhtml.xsl
+++ b/doc/specs/vulkan/config/docbook-xsl/xhtml.xsl
@@ -17,14 +17,16 @@
 
 <!-- Add MathJax <script> tags  to document <head> -->
 <xsl:template name="user.head.content">
-    <script type="text/x-mathjax-config">
-        MathJax.Hub.Config({
-            MathML: { extensions: ["content-mathml.js"] },
-            tex2jax: { inlineMath: [['$','$'], ['\\(','\\)']] }
-        });
-    </script>
-    <script type="text/javascript"
-        src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
-    </script>
+  <script type="text/x-mathjax-config">
+  MathJax.Hub.Config({
+    tex2jax: {
+      inlineMath: [ ['$','$'], ['\\(','\\)'] ]
+    },
+    "fast-preview": { disabled: true },
+    "AssistiveMML": { disabled: true }
+    });
+  </script>
+  <script type="text/javascript" src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_CHTML">
+  </script>
 </xsl:template>
 </xsl:stylesheet>
--- a/doc/specs/vulkan/config/mathjax.js
+++ b/doc/specs/vulkan/config/mathjax.js
@@ -1,8 +1,11 @@
-    <script type="text/x-mathjax-config">
-	MathJax.Hub.Config({
-	    MathML: { extensions: ["content-mathml.js"] },
-	    tex2jax: { inlineMath: [['$','$'], ['\\(','\\)']] }
-	});
-    </script>
-    <script type="text/javascript" src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
-    </script>
+  <script type="text/x-mathjax-config">
+  MathJax.Hub.Config({
+    tex2jax: {
+      inlineMath: [ ['$','$'], ['\\(','\\)'] ]
+    },
+    "fast-preview": { disabled: true },
+    "AssistiveMML": { disabled: true }
+    });
+  </script>
+  <script type="text/javascript" src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_CHTML">
+  </script>


krOoze:
BTW I was also experimenting with KaTeX in https://github.com/krOoze/Vulkan-Docs/tree/KaTeX . Seems faster/immediate, but won't render everything, so not a fair comparison and not a direct replacement. Also it doesn't work correctly in the XHTML Transitional schema (only in the html version of the spec)...

oddhack:
This seems very promising - thanks! Will bring up for evaluation by the Vulkan working group.

I'm kinda hoping there's some way to pre-generate the equation HTML eventually but haven't explored that. We will be moving to HTML5 whenever we can transition to asciidoctor - asciidoc just doesn't support that as an output target.

TobiasHector:
@oddhack FYI there's a toolchain to generate SVG files from latexmath that has been integrated with AsciiDoctor, which I'm looking into along with the rest of the transition. I was only going to use it for the pdf output, but I can look into doing it for the html outputs too. Then it just means loading up SVGs when the page loads, which should be much quicker.

krOoze:
> I'm kinda hoping there's some way to pre-generate the equation HTML eventually but haven't explored that.

More than likely there is. I just couldn't stand it anymore and went after the low hanging fruit for the short term.

@TobiasHector Now that you brought it up, how's the migration to AsciiDoctor progressing anyway?

TobiasHector:
> @TobiasHector Now that you brought it up, how's the migration to AsciiDoctor progressing anyway?

It's getting there. It's going to be a little painful to transition due to some awkward incompatibilities that mean I'll have to touch multiple portions of the spec for single line fixes (mostly equations and tables) - but I can generate documents for most of the targets with 90% of the content there with some quick fixes. Equations are the main missing piece right now. Unfortunately the last 10% is always the hardest bit ;)

oddhack:
I checked this out on my machine (AMD FX-6300 hexacore, Debian 8, Chromium 50.0). I'm not entirely sure how to measure processing time other than counting from when I hit return in the address bar until the status updates from MathJax in the lower left stop updating, but if that's a good measure, the results actually seem to be worse with this change. About 7-8 seconds for the 1.0 spec with the old MathJax config, about 17 seconds with the new one. I thought maybe there was some caching effect taking place and tried reloading a couple of times, and then opening in a new Chromium user window, but there didn't seem to be much change. Any thoughts on what I'm seeing here? I believe another Khronos person tried on their machine with similar results.

oddhack:
The main observable difference is after the initial loading phase, I see a bunch of "Typesetting math: ##%" messages, which correspond to the extra time.

oddhack:
I was unfamiliar with KaTeX before seeing this. It does seem potentially promising. We use the align* environment a lot but I suspect that could be replaced by aligned.

The trickiest part for me was getting math imbedded in asciidoc to flow through properly to both MathJax-in-XHTML and through Docbook to dblatex. Now we are looking at moving to asciidoctor and it's a similar problem, with different details. They have a direct PDF generator but AFAICT the  hoops you have to jump through to hope to deal with math on that path are many and varied. So we would probably end up, at least initially, with asciidoctor->HTML5 and asciidoctor->Docbook->dblatex paths. Unfortunately part of the motivation for asciidoctor was to speed the build times, and since the PDF path would be essentially unchanged, this wouldn't help as much as we'd like.

ratchetfreak:
Chrome and Firefox both have a profiler tab in their dev tools both have the same keyboard shortcut  to access them (ctrl+shift+I).

TobiasHector:
> at least initially, with asciidoctor->HTML5 and asciidoctor->Docbook->dblatex paths. Unfortunately part of the motivation for asciidoctor was to speed the build times, and since the PDF path would be essentially unchanged, this wouldn't help as much as we'd like.

We should have 'direct', and more importantly *identical* paths for both, with math flowing through the "mathematical" ruby gem, which would generate svg files from latexmath, which are then included in the html, and rendered into the pdf. Ideally I'd like to ditch the docbook toolchain altogether, and it's looking like that shouldn't be a problem.

krOoze:
@oddhack I should perhaps disclose my mesurement environment and method.

I did it on Celeron G1610, Windows 10 x64, Firefox 46. Nothing fancy though to ensure science grade measurement - just my regular PC. Should have said so earlier. Good thing, you are testing different browsers and setups.

I loaded the page once (so it would cache or whatnot). Then I measured the time from clicking reload until the bold final (not the thin preview) equations appear. Again nothing fancy, just me clicking stopwatch on my phone. Did that three times, to make sure the times are consistent (they are, +-1 s). I can post the raw data, if interested.

krOoze:
(Additional) measurements:

| browser    | original [s] | proposed [s] |
|------------|--------------:|--------------:|
| Firefox 46 | 74           | 34           |
| IE 11      | 114          | 28           |
| Chrome 50  | 43           | 25           |

oddhack:
@TobiasHector AFAICT all the Ruby stuff for math-in-direct-PDF was in an incomplete and badly maintained state; it was not just a matter of 'gem install' to attempt to get it working, there were platform binary components which are going to be difficult to deploy (ala python-lxml, but at least that has prebuilt packages that work), and I eventually gave up on attempts to play with it for the time being.

oddhack:
@krOoze are you opening up the spec at an anchor that shows some visible math, in order to address the preview vs. final issue? Otherwise I'm not sure how to tell, since the start of the document doesn't have any math (and it would be rendered first anyway, presumably). If you're doing that, what anchor are you testing at?

krOoze:
@oddhack Yes, usually at ch. 2.8 (it's near the beggining so it won't jerk away). Firefox seems to start at the anchor, other browsers start at the beggining of the page until the script finishes and then jump(so you can wait for that too).

Through browser profiler as @ratchetfreak said might be more pro version of doing it...

It doesn't seem to matter which equation you looking at (they "render" the second version invisible and then just switch it on by css I think - which is relatively immediate).

I think the "loading" includes parsing of that script and downloading is only small part of it. As it happens to be, I do also have a [offline/local MathJax branch](https://github.com/krOoze/Vulkan-Docs/tree/offline_mathjax). Haven't tested it for speed though. Intended it only to make version that does not need internet.

TobiasHector:
@oddhack Correct, but it's not too painful to setup once you have a decent set of instructions (which I'm working on) - no more painful than setup of a latex toolchain IMO! Anyway if we want to discuss this more let's do it offline - but I am working through all the kinks.

krOoze:
I am just gonna close this.

Recent update (moving some of the math outside MathJax) improved loading times from maddening to psychologically manageable.

I guess there are also still maybe some changes coming.

Also in the meantime anyone can make these changes by right-click context menu on a MathJax element (Math Settings -> Math Renderer). The browser tends to remember (probably through cookies).

Issue #219, opened 2016-05-21
Figures and Equations missing (nubered) captions


krOoze:
1. Some Figures in specification are missing numbered captions. Same for block equations.
As a result they are missing in the **List of Figures** and **List of Equations** in the HTML spec.
Well... or vice-versa: some have them and should not (I could live without the List of X...).

2. The examples are not in **List of Examples**.

3. Well, there is one item in the **List of Examples**, which I think was supposed to be a regular Note block.

4. [x] Considering the above, should investigate the Tables too...

oddhack:
The LOF should be fixed in the next update.

The LOE exists because those particular equations are linked from elsewhere in the spec, and it's not obvious how to remove them from the LOE without removing the links as well. I don't think we're likely to change that (or to mark up every other block equation gratuitously just so they're all in the LOE, as that's an undesired side-effect).

The LOExamples should be removed.

Finally if you'll identify any tables missing from that list, I'll take a look at that.

I say "should" as I don't expect any issues in cleaning all this up, but sometimes the WG decides to do something slightly different.

krOoze:
Investigated the tables and found missing tables and some emergent issues (concrete issues/fixes are in the referenced PR).

Also all those Command Properties tables are missing there, but I assume that's intentional...

oddhack:
LOF fixed, LOE unchanged, and LOExamples removed all per above. I'll keep this open to take a look at the issues in your PR.

krOoze:
This is fixed in 1.0.15 update as per above comment. Table problems follow-up is in #330.

PR #221, opened 2016-05-23
List of Tables fixes

krOoze:
Fixed:
- ~~3~~ [since 1.0.21] 2 missing tables relating to issue #219
- misformated names with underline
- missing link styling for MathJax elements

Most of those are trivial fixes, except:

One of the tables re-explains VkResult values. I removed that altogether, because that seemed redundant to the Return Codes block and 2.6.2 chapter.

To make math in links underlined I added CSS rule for inlineMedia class and its descendants to inherit decoration style (hopefuly form \<a\>). I made some assumptions and the rule uses evil *, so if someone knows CSS well, I could use some advice on that...


krOoze:
Abandoned in favor of Issue.

PR #222, opened 2016-05-23
Fixing variable names in math

krOoze:
This fixes the variable names with operator-like contents rendered with
extra space like this:
`struct. member` and `struct :: member`.

Fixed by replacing \mathit with \textit.

Could also be worth a styleguide note? And/or replace all the \mathit in the spec?
EDIT: Or moving the parameter names out of the math environment (and keeping only the operators in) could be viable for the simpler cases.


krOoze:
Abandoned in favor of an issue.

PR #225, opened 2016-05-24
Covering Get*ProcAddr() NULL return cases

krOoze:
Fixes emergent issue in #212 

Rewording in such a way `Get*ProcAddr()` returns NULL in all reasonable
error cases. Additions to previous version:

1) General cleanup and more clear and enforcing language

2) vkGetInstanceProcAddr returns NULL even for supported non-enabled
extension functions

3) really returns NULL for all recoverable errors (except for `pName == NULL` which is still inValid Usage)

Requesting advice:
- Not sure if it should include `pName == NULL` case. It does imply it even in the old version.
- Not sure if it should include `instance == valid instance` case and requesting freestanding command and `instance == NULL` and requesting instance commands
- ~~Not sure if `vkGetDeviceProcAddr()` can apply to some instance extensions' commands.~~ R: it can
- Not sure if `vkGetInstanceProcAddr()` can apply to some device extensions' commands.
- Not sure if requesting its own pointer should conceptually return NULL or a valid pointer.

### DIFF
--- a/doc/specs/vulkan/chapters/initialization.txt
+++ b/doc/specs/vulkan/chapters/initialization.txt
@@ -17,7 +17,7 @@ command:
 include::../protos/vkGetInstanceProcAddr.txt[]
 
   * pname:instance is the instance that the function pointer will be
-    compatible with.
+    compatible with, or `NULL` for commands not dependent on any instance.
   * pname:pName is the name of the command to obtain.
 
 include::../validity/protos/vkGetInstanceProcAddr.txt[]
@@ -31,21 +31,28 @@ Vulkan commands as well; if this is done, then applications that use only
 the core Vulkan commands have no need to use fname:vkGetInstanceProcAddr.
 
 Function pointers to commands that don't operate on a specific instance can:
-be obtained by using this command with pname:instance equal to `NULL`. The
-following commands can: be accessed this way:
+be obtained by using this command with pname:instance equal to `NULL`. Only
+pointers to the following commands can: be obtained this way:
 
   * flink:vkEnumerateInstanceExtensionProperties
   * flink:vkEnumerateInstanceLayerProperties
   * flink:vkCreateInstance
 
-If pname:instance is a valid sname:VkInstance, function pointers to any
-commands that operate on pname:instance or a child of pname:instance can: be
-obtained. The returned function pointer must: only be called with a
-dispatchable object (the first parameter) that is a child of pname:instance.
+If pname:instance is a valid sname:VkInstance, then function pointer to any
+command from any of the following groups can: be obtained:
 
-If pname:pName is not the name of a core Vulkan command, or is an
-extension command for any extension not supported by any available layer or
-implementation, then fname:vkGetInstanceProcAddr will return `NULL`.
+  * core Vulkan commands
+  * commands of instance extensions enabled for pname:instance
+  * commands of device extensions enabled for a valid sname:VkDevice child
+    object of pname:instance
+
+Any other valid usage of fname:vkGetInstanceProcAddr command must:
+return `NULL`.
+
+If pname:instance is a valid sname:VkInstance and the command does return a
+valid (non-`NULL`) function pointer, then the returned function pointer
+must: only be called with a dispatchable object (the first parameter) being
+pname:instance or its child.
 
 ifdef::editing-notes[]
 [NOTE]
@@ -79,17 +86,27 @@ command:
 include::../protos/vkGetDeviceProcAddr.txt[]
 
   * pname:device is the logical device that provides the function pointer.
-  * pname:pName is the name of any Vulkan command whose first parameter
-    is one of
-  ** sname:VkDevice
-  ** sname:VkQueue
-  ** sname:VkCommandBuffer
-
-If pname:pName is not the name of one of these Vulkan commands, and is
-not the name of an extension command belonging to an extension enabled for
-pname:device, then fname:vkGetDeviceProcAddr will return `NULL`.
+  * pname:pName is the name of Vulkan device dependent command.
 
 include::../validity/protos/vkGetDeviceProcAddr.txt[]
+
+For any command which takes a first parameter of any of the following types:
+
+  * sname:VkDevice
+  * sname:VkQueue
+  * sname:VkCommandBuffer
+
+and at the same time being from any of the following groups function pointer
+can: be obtained:
+
+  * core Vulkan commands
+  * commands of instance extensions enabled for the sname:VkInstance parent
+    object of pname:device
+  * commands of device extensions enabled for pname:device
+
+Any other valid usage of fname:vkGetDeviceProcAddr command must:
+return `NULL`.
+
 
 [[initialization-instances]]
 == Instances
--- a/src/spec/vk.xml
+++ b/src/spec/vk.xml
@@ -2994,17 +2994,11 @@ maintained in the master branch of the Khronos Vulkan Github project.
             <proto><type>PFN_vkVoidFunction</type> <name>vkGetDeviceProcAddr</name></proto>
             <param><type>VkDevice</type> <name>device</name></param>
             <param len="null-terminated">const <type>char</type>* <name>pName</name></param>
-            <validity>
-            </validity>
         </command>
         <command>
             <proto><type>PFN_vkVoidFunction</type> <name>vkGetInstanceProcAddr</name></proto>
             <param optional="true"><type>VkInstance</type> <name>instance</name></param>
             <param len="null-terminated">const <type>char</type>* <name>pName</name></param>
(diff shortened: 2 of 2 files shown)


doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -43,9 +43,13 @@ commands that operate on pname:instance or a child of pname:instance can: be
 obtained. The returned function pointer must: only be called with a
 dispatchable object (the first parameter) that is a child of pname:instance.
 
-If pname:pName is not the name of a core Vulkan command, or is an
-extension command for any extension not supported by any available layer or
-implementation, then fname:vkGetInstanceProcAddr will return `NULL`.
+If pname:pName is not the name of one of these Vulkan commands, and is not
```
courtney-g: "these" is not clear. Is that a reference to the EnumerateInstance* and vkCreateInstance?
Shouldn't that be: If pname:pName is not the name of a core Vulkan command and is not ...

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -85,9 +89,12 @@ include::../protos/vkGetDeviceProcAddr.txt[]
   ** sname:VkQueue
   ** sname:VkCommandBuffer
 
-If pname:pName is not the name of one of these Vulkan commands, and is
-not the name of an extension command belonging to an extension enabled for
-pname:device, then fname:vkGetDeviceProcAddr will return `NULL`.
+If pname:pName is not the name of one of these Vulkan commands, and is not
```
courtney-g: Same here. "these" is not clear to me.

krOoze: Well, was there before... also follows DRY principle.
What do you sugest?

doc/specs/vulkan/chapters/initialization.txt
krOoze: Made to match `DeviceGetProc()` version. Though the situation is more complicated because it references two paragraphs above not just one....

doc/specs/vulkan/chapters/initialization.txt
courtney-g: Okay, here "these" refers to any Vulkan command whose first parameter is one of VkDevice, VkQueue or VkCommandBuffer. With the description  immediately above the statement "If pname:pName ..." it's okay.

doc/specs/vulkan/chapters/initialization.txt
courtney-g: Right, it's the intervening paragraph that makes "these Vulkan commands" fuzzier.

instance==NULL should only apply to EnumerateInstance* and CreateInstance.
instance== valid VkInstance applies to everything else (all core Vulkan commands as well as enabled instance extension functions.)

How about:
If pname:pName is not a core Vulkan command and is not the name of an enabled instance extension command, then fname:vkGetInstanceProcAddr must: return 'NULL'.

krOoze: Was there before and perhaps should be done to trivialy fix the immediate problem of supportted vs enabled. But IMHO it has these issues:

1) Redefines what is said above. For perfection would need a chapter rewrite so that all of that links to same definition of "core Vulkan command" and "extension command".

2) What about the case `instance == NULL` and getting core  command or extension command?

3) What about the case of `instance == valid` and getting freestanding command?

krOoze:
I am tempted to change it once more to something simpler approximately alike to:

> For any case not covered above the command must return `NULL`.

Would that make any sense?

courtney-g:
I've come to a similar conclusion. There is some related work happening in a Khronos thread. Once that settles out I'll update this PR with the updated language and left folks comment.

courtney-g:
I believe this is now fixed with the 1.0.18 spec release.

krOoze:
@courtney-g IMHO, not really. The branch is rebased so you can judge from diff yourself...

**Most importantly:**
In .18 only change was the VU for vkGet**Device**ProcAddr was removed.
But VU still present for vkGet**Instance**ProcAddr. This PR also fixes that and moves that to description section.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -17,7 +17,7 @@ command:
 include::../protos/vkGetInstanceProcAddr.txt[]
 
   * pname:instance is the instance that the function pointer will be
-    compatible with.
+    compatible with, or `NULL` for commands not dependent on any instance.
```
courtney-g: Okay, but that is covered in the Valid Usage.
> If instance is NULL, pName must be one of: vkEnumerateInstanceExtensionProperties, vkEnumerateInstanceLayerProperties or vkCreateInstance

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -31,21 +31,28 @@ Vulkan commands as well; if this is done, then applications that use only
...
+  * commands of device extensions enabled for a valid sname:VkDevice child
+    object of pname:instance
+
+Any other valid usage of fname:vkGetInstanceProcAddr command must:
+return `NULL`.
+
+If pname:instance is a valid sname:VkInstance and the command does return a
+valid (non-`NULL`) function pointer, then the returned function pointer
+must: only be called with a dispatchable object (the first parameter) being
+pname:instance or its child.
```
courtney-g: Could you provide some more rational for this change?
I'm not seeing an issue with the original text, would like to understand your perspective.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -31,21 +31,28 @@ Vulkan commands as well; if this is done, then applications that use only
...
-dispatchable object (the first parameter) that is a child of pname:instance.
+If pname:instance is a valid sname:VkInstance, then function pointer to any
+command from any of the following groups can: be obtained:
 
-If pname:pName is not the name of a core Vulkan command, or is an
-extension command for any extension not supported by any available layer or
-implementation, then fname:vkGetInstanceProcAddr will return `NULL`.
+  * core Vulkan commands
+  * commands of instance extensions enabled for pname:instance
+  * commands of device extensions enabled for a valid sname:VkDevice child
```
courtney-g: This is not correct.
Right now, the behavior for GetInstanceProcAddr when given a device function name implemented by the system (e.g. can live in the ICD, loader or layer) is to return a function pointer to trampoline code. The extension still needs to be enabled to function, but app would have a valid function pointer.

That was the intent of:
> If pname:pName is not the name of a core Vulkan command, or is an extension command for any extension not supported by any available layer or implementation, then fname:vkGetInstanceProcAddr will return `NULL`.

Maybe a table would be a more appropriate explanation.

sname:VkInstance | pname:pName | return value
------------ | ------------- | ------------
NULL | flink:vkEnumerateInstanceExtensionProperties | fp
NULL | flnk:vkEnumerateInstanceLayerProperties | fp
NULL | flink:vkCreateInstance | fp
NULL | * (any pname:pName not covered above) | NULL
instance | core Vulkan command | fp
instance | enabled instance extension commands for pname:instance | fp
instance | available device extension commands for pname:instance | fp
instance | * (any pname:pName not covered above) | NULL
invalid instance | * | undefined

doc/specs/vulkan/chapters/initialization.txt
krOoze: In this line the only point is so the description doesn't lie (if it is "the instance" then it can't be NULL -- so it must be stated too)

>  Currently Valid Usage says that if instance is NULL, pName can only be those three calls and everything else is "undefined"

Precisely, if it is in VU, then the case `instance==NULL` and not one of the three command names is undefined behavior.
I think (and I could be wrong) that the intention is to return NULL in that case too instead.
(Same for the "must not" VU except the oposite is undefined behavior.)

doc/specs/vulkan/chapters/initialization.txt
krOoze: Nice table!

What's "available extension" then? Some union of all supported extensions of all PhysDevices in `instance`? Or just registered extensions?

I assume invalid string (`NULL` pointer) is meant to be undefined too... And maybe some OS can return NULL on general failure too (though I removed mention of this in this version...)

doc/specs/vulkan/chapters/initialization.txt
krOoze: If you mean just this one paragraph:

I did it to separate the two topics sharing single paragraph on original L41. Turns out I needed to do it anyway to be able to list the cases as bullet list.

courtney-g: I'm liking the table form. Defines what happens for all situations and feels less wordy. I'd just replace all the descriptive text with the table.

courtney-g:
Posted #273 for review.

PR #227, opened 2016-05-24
VkDebugMarkerObjectNameInfoEXT instead of VkDeviceCreateInfo

SaschaWillems:
Example 1 of the VK_EXT_debug_marker specification was erroneously using a VkDeviceCreateInfo structure instead of VkDebugMarkerObjectNameInfoEXT  to set the debug object name for the image.


baldurk:
Whoops! My mistake there.

:+1:

krOoze:
@baldurk That nonsense is still in there. Probably forgot about this?

TobiasHector:
@SaschaWillems This looks like a no brainer, but since you've opened the PR we need to you accept the CLA before we can accept this - there's no merge conflicts so it should be straightforward to accept this once you do so.

oddhack:
Woops. I didn't notice this wasn't merging to the 1.0 branch, since the PR predated the single-branch model. I'll do that in the internal process for this week's release and push it back to github.

PR #228, opened 2016-05-29
Fix three figures still missing from LOF

krOoze:
relates to issue #219 

- Fix 3 figures missing from LOF.
- Also updating them to new style from 1.0.15 styleguide.
- tweaked source (used open format instead - don't have new Office, everything else renders the original differently or can't get good SVG out of it)
- changed print widths for esthetics
- tested with XHTML and PDF output


krOoze:

Issue #220, opened 2016-05-21
Eventually replace use of scalar dot operator with LaTeX / Unicode \times symbol.


krOoze:
1) ~~(2x) The chapter mentions **c** but then works with **f'** instead.~~ fixed in 1.0.15

2) ~~(2x) The equation uses cross/cartesian product symbol instead of scalar multiplication dot.~~ Resolution: specmaker prefers cross style. -> elsewhere dot multiplication is used incorrectly in styleguide and ch. 2.7.1

3) ~~(2x)~~

> [...]integer values with exactly b bits which are closest to the floating-point value r (where rounding to nearest is preferred).

~~"is preferred"? Is that a **must**("is preferred by the implementation") or a **should**("the specification would prefer implementation to do so")?~~ fixed in 1.0.15

4) ~~(2x)~~ (1x) - one case fixed in 1.0.15

> If r is equal to an integer, then that integer value is returned. In particular, if f is equal to 0.0 or 1.0, then f′ must be assigned 0 or 2b−1 , respectively.

Again avoiding the keyword. Is the first sentence **must** and the second an interesting note? Or is it **should** and the second sentence some minimal guarantee?

5)

> This equation is used everywhere that floating-point values are converted to signed normalized fixed-point, including when querying floating-point state and returning [signed??] integers, as well as for specifying signed normalized texture or framebuffer values using floating-point.

~~Should "signed" be in the place I made a note in the quote?~~ fixed in 1.0.15

6) ~~On some occasions in the spec max functions are written as: max{} instead of max(). (In ch.2.8.1, 24.6.3)~~ fixed in 1.0.15, though now I see why { was there - the renderer renders them too small for their content

oddhack:
Most of this will get cleaned up in the next update. The symbol for (2) is not the cross product symbol, it's the LaTeX \times symbol, and is used consistently (unless we missed a few cases), so that's not going to change.

krOoze:
I suppose that's one of those cultural things... We never use cross for regular multiplication (they teach it in elementary school, then later we are told it's wrong usage :) ).

Well, in that case ch.2.7.1 uses dot for multiplication.

oddhack:
[...]

PR #229, opened 2016-05-29
Fix small parentheses to fit content

krOoze:
Tweaking Latex math to make parentheses right size for the content. Related to issue  #220 .


krOoze:

PR #230, opened 2016-05-29
Fixing must: omitted in 1.0.15 change

krOoze:
as per remains in issue #220


krOoze:

PR #232, opened 2016-05-30
Fix issues found causing xhtml invalidity

krOoze:
Fixing some issues found using HTML validator:

1) Some anchors with names containing spaces
2) Leaking `xmlns""`
3) xhtml page send with wrong MIME

4) `cellspacing` and `cellpadding` used in CSS - PR #228 should accidentally fix that


krOoze:

PR #244, opened 2016-06-11
Make sure dates in header are english

krOoze:
`date` prints date in locale language
`-R` prints as RFC 2822 (which is, among other things, english)
(optionally) `-u` prints time in UTC, not locale time zone


krOoze:

Issue #259, opened 2016-06-19
Should vkCreateDevice (still) return VK_ERROR_LAYER_NOT_PRESENT?


krOoze:
Should `vkCreateDevice` (still) return VK_ERROR_LAYER_NOT_PRESENT after the device layer deprecation, as is stated in the Return Codes section?

BTW:
1) *_NOT_PRESENT is not explained there same way as is newly done for the `vkCreateInstance`.

There is a paragraph in `vkCreateInstance` description explaining when VK_ERROR_{LAYER|EXTENSION}_NOT_PRESENT is returned. I found it odd, the same is not done for `vkCreateDevice` for the now remaining VK_ERROR_EXTENSION_NOT_PRESENT (or possibly of course vice-versa why it **is** explained for `vkCreateInstance`)

2) maybe the `VkResult` descriptions "is not supported" is a bit misleading and so is the new "cannot be found". I think, it is supposed to be simply "failed to enable" instead...

The "is not supported" wording is in 2.6.2. Return Codes chapter for VK_ERROR_{LAYER|EXTENSION}_NOT_PRESENT. And the "cannot be found" is in the description section for `vkCreateInstance` command.
I simply found those wordings too specific for the purpose(and as such misleading). I thought, those errors are meant to be simply returned when the relevant `create*` command is unable to enable the specified layer/extension for whatever reason (not only for some "failure of finding" or "supporting", whatever those mean).

courtney-g:
Have made pull request !260 to address this. krOoze, Please verify as I'm not certain I've covered everything listed in the issue (BTW items).

krOoze:
@courtney-g Obvious fix. But no, not really addressing the "BTW" items at all (yet?). Was I not clear with their description, or why do you ask for verification? I will explain them more precisely in the original post...

courtney-g:
Updated patch should address issue #1.

As for:
> 2) maybe the VkResult descriptions "is not supported" is a bit misleading and so is the new "cannot be found". I think, it is supposed to be simply "failed to enable" instead...

I'd be curious to see what other interested parties think is a good description for the error code VK_ERROR_LAYER_NOT_FOUND.

1) (current spec) VK_ERROR_LAYER_NOT_PRESENT: A requested layer is not present or could not be loaded.

2) (proposed change, IIRC) VK_ERROR_LAYER_NOT_PRESENT: Failed to enable a requested layer.

[...]

PR #260, opened 2016-06-20
LAYER_NOT_PRESENT no longer valid for CreateDevice

courtney-g:
Fixes issue #259

[ChangeLog]Remove unused error code now that device layers have been
deprecated.


doc/specs/vulkan/chapters/devsandqueues.txt
```diff
@@ -347,6 +347,13 @@ include::../structs/VkDeviceCreateInfo.txt[]
 
 include::../validity/structs/VkDeviceCreateInfo.txt[]
 
+fname:vkCreateDevice creates a VkDevice object, then enables
+extensions requested by the application. If an extension
+is provided by a layer, the layer must: be specified at
```
krOoze: Hmm..., **must** it really? Or should it instead in that case return the **VK_ERROR_EXTENSION_NOT_PRESENT**.

Now I see the same thing is in the description `vkCreateInstance`.
And it is relevant to the BTW item 2 in issue #259 (is VK_ERROR_EXTENSION_NOT_PRESENT supposed to catch all problems or not?).

krOoze:
Oh, (I hate to BTW again :blush: but) BTW:
What about extension-extension dependency? Is failure to specify both extensions supposed to return VK_ERROR_EXTENSION_NOT_PRESENT too (and so should it be in the description paragraph too?)

doc/specs/vulkan/chapters/devsandqueues.txt
courtney-g: The statement is true. "If an extension is provided by a layer, the layer **must** be specified at CreateInstance time."
We've moved that out of the "Valid Use" section as that implied the app needed to verify that the requested extension is supported/exists prior to calling Create(Instance|Device). That's not necessary. The intent of that statement is to make it clear that an application needs to do both parts - enable the layer and the extension to use the extension. Verses having the loader automatically enable layers based on extensions requested.

courtney-g:
AFAIK, what happens with dependencies between layers is not specified.
What we've talked about in the working group is that the layer is responsible for checking that it's dependencies are satisfied before the vkCreate(Instance|Device) call can complete successfully. I would expect the layer to simply return one of the *_NOT_PRESENT errors. I would also recommend that such a layer log more details via the VK_EXT_debug_report extension callback so that a developer can figure out what's happening.

krOoze:
> AFAIK, what happens with dependencies between layers is not specified.

I was speaking of **extensions** not layers. E.g.VK_KHR_swapchain requiring VK_KHR_surface.

doc/specs/vulkan/chapters/devsandqueues.txt
krOoze: Oh... If I understood you correctly then that is a wrong usage of RFC "must" (for raising requirement purposes, **not** for stating truth).

courtney-g:
Ah, right, extensions. I'd expect the same logic to apply there. If an extension requires another extension it should check that the other extension has been enabled at Create(Instance|Device) time.

krOoze:
> Ah, right, extensions. I'd expect the same logic to apply there.

So currently unspecified too, I take it.

doc/specs/vulkan/chapters/devsandqueues.txt
courtney-g: Here's definition of "must":
This word, or the terms **required** or **shall**, mean that the definition is an absolute requirement of the specification.

It's still required that the layer must be specified at CreateInstance time to get the desired behavior. The application **must** explicitly enable functionality it wants to use.

courtney-g:
> So currently unspecified too, I take it.

As far as I know. I expect this will get addressed in the not too distant future as Vulkan extensions become available.

doc/specs/vulkan/chapters/devsandqueues.txt
krOoze: oook then, so it **is** a requirement, so it is not (can't be) handeled by VK_ERROR_EXTENSION_NOT_PRESENT return.

But in that case should be caught by validation layers (and can be easily so, I think) and as such that would be good inside VU. (IIRC statements was removed from VU because of other things, not because of this).

krOoze:
^ Sure, I just like pointing spec holes, no matter how small/unimportant... thanks for taking the time

krOoze:
This is merged in 1.0.19 so can be closed and branch deleted.

I can post my BTW points here and in #259 in a new issue more concisely and detached from this problem.

PR #271, opened 2016-06-30
Minor typo

damag:


krOoze:
@damag Ref pages are automatically generated since 1.0.20 update. So this is perhaps fixed. Close?

damag:
Yes, it seems the text has been updated anyways. Thanks.

Issue #265, opened 2016-06-25
Invalid externsync atribute of param tag


krOoze:
The xml schema Readme says `externsync` attribute of `<param>` must be either omitted or `="true"`.

Yet there are many occurences where it has different value (assumably some list of concrete objects needing ext. synchronization):

https://github.com/KhronosGroup/Vulkan-Docs/blob/eb9997129d/src/spec/vk.xml#L3108
https://github.com/KhronosGroup/Vulkan-Docs/blob/eb9997129d/src/spec/vk.xml#L3264
https://github.com/KhronosGroup/Vulkan-Docs/blob/eb9997129d/src/spec/vk.xml#L3662
https://github.com/KhronosGroup/Vulkan-Docs/blob/eb9997129d/src/spec/vk.xml#L3681
https://github.com/KhronosGroup/Vulkan-Docs/blob/eb9997129d/src/spec/vk.xml#L3683
https://github.com/KhronosGroup/Vulkan-Docs/blob/eb9997129d/src/spec/vk.xml#L3757
https://github.com/KhronosGroup/Vulkan-Docs/blob/eb9997129d/src/spec/vk.xml#L4578

TobiasHector:
@oddhack I believe you made this particular change - not me. I mean I can give it a go at documenting it if you like, but I'm not 100% sure what the semantics here are, so it might be guesswork :).

oddhack:
Hm, I've forgotten. If you happen to know the internal MR/issue that led to this change, maybe I can track down whatever happened. I vaguely recall getting improper VU statements and wanting to fix it but that's about it.

oddhack:
This should be fixed in the 1.0.32 spec update.

Issue #245, opened 2016-06-11
(again) Undesired unicode replacements (⇐ instead of ≤, → instead of ->)


krOoze:
Similar as in issue #239 ,
in the [new 1.0.16 changes to features.txt](https://github.com/KhronosGroup/Vulkan-Docs/commit/5de77cf663c84cc2edfa355c49319dca0a869eb2#diff-7e6dee71ddb41acf30a28931d6bc5033), it creates Unicode arrow ⇐ instead of ≤.
Do not use `<=` in that context!  :no_entry_sign:

UPDATE: still not fixed in 1.0.17. And BTW neither is the similar prematurely closed issue #239 .

UPDATE 2: also C arrow operators -> are replaced with Unicode → all over the place.

consult http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#text-replacement 
but that seems to be all.
Only found some inconsistent em dash usage (some with space around, some not). And in three cases the desired arrow is made using the latexmath instead of making use of the text replacement feature (or direct Unicode).

oddhack:
There are no ⇐ symbols in the 1.0.18 spec update, AFAICT.

No plans to do anything about the → issue. They are appearing in places where they're either perfectly fine, like the table row headings, or just explanations in VU statements - not code samples people can cut and paste. I believe the meaning is clear. There are lots of higher priorities so nobody is motivated to deal with minor markup issues of this sort.

Finally, nobody is at all eager to use non-ASCII Unicode characters in the spec sources at present. We've tried to do that previously and had trouble with differing results on the toolchains that create HTML and PDF output, respectively.

PR #272, opened 2016-07-01
Fixing stuff related to C -> operator

krOoze:
1) Reused existing function (from `HostSynchronizationOutputGenerator`) for "human readable" output of externsync attribute values

2) Escaped `->` so it is not replaced with unicode arrow

3) Added `pname:` to members after C dereferences ( only after `->` -- there are no relevant `.` cases I think) 

4) Replaced `>` with `&gt;` (not necessary but a XML recommendation)

5) Fixed `.` where `->` should be instead

6) Opportunisticaly fixed "created" (on Pool resource) to "allocated"

Fixes part of #245 
Perhaps 1) gives hint how externsync in #265 should be specified in the schema documentation

Known problem: The arrows are ugly in XHTML output (nice in PDF) due to chosen font. If I make it ++literal++, then it is nice, but the PDF output decides to break line in the exact middle of the arrow in one case. Haven't found the appropriate hack for this to work correctly...

### DIFF
--- a/src/spec/validitygenerator.py
+++ b/src/spec/validitygenerator.py
@@ -346,7 +346,8 @@ class ValidityOutputGenerator(OutputGenerator):
                 if 'latexmath:' in lengths[0]:
                     asciidoc += lengths[0]
                 else:
-                    asciidoc += self.makeParameterName(lengths[0])
+                    asciidoc += 'pname:'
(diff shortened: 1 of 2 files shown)


krOoze:
This seems to have been resolved in 1.0.28 by replacing all `->` with `::`

PR #273, opened 2016-07-01
GitHub pr225b get proc addr

courtney-g:
Follow-on from #225 
Use table to enumerate all cases for vkGetInstanceProcAddr and vkGetDeviceProcAddr

### DIFF
--- a/doc/specs/vulkan/chapters/initialization.txt
+++ b/doc/specs/vulkan/chapters/initialization.txt
@@ -72,18 +72,24 @@ command:
 
 include::../protos/vkGetDeviceProcAddr.txt[]
 
-  * pname:device is the logical device that provides the function pointer.
-  * pname:pName is the name of any Vulkan command whose first parameter
-    is one of
-  ** sname:VkDevice
-  ** sname:VkQueue
-  ** sname:VkCommandBuffer
-
-If pname:pName is not the name of one of these Vulkan commands, and is
-not the name of an extension command belonging to an extension enabled for
-pname:device, then fname:vkGetDeviceProcAddr will return `NULL`.
-
 include::../validity/protos/vkGetDeviceProcAddr.txt[]
+
+The table below defines the various use cases for fname:vkGetDeviceProcAddr
+and expected behavior for each case, in the order defined by the table below.
+
+.vkGetDeviceProcAddr behavior
+[width="80%",options="header",align="center"]
+|=====
+| pname:device | pname:pName | behavior
+| invalid device | * | undefined
+| device | `NULL` | undefined
+| device | core Vulkan command (1) | returns fp
+| device | enabled device extension command for pname:device | returns fp
+| device | * (not any row/case above) | returns `NULL`
+|=====
+
+1) pname:pName is the name of any Vulkan command whose first parameter
+    is one of sname:VkDevice, sname:VkQueue, or sname:VkCommandBuffer.
 
 [[initialization-instances]]
 == Instances
--- a/doc/specs/vulkan/chapters/initialization.txt
+++ b/doc/specs/vulkan/chapters/initialization.txt
@@ -17,7 +17,7 @@ command:
 include::../protos/vkGetInstanceProcAddr.txt[]
 
   * pname:instance is the instance that the function pointer will be
-    compatible with.
+    compatible with, or `NULL` for commands not dependent on any instance.
   * pname:pName is the name of the command to obtain.
 
 include::../validity/protos/vkGetInstanceProcAddr.txt[]
@@ -30,41 +30,35 @@ Loaders are encouraged to export function symbols for all other core
 Vulkan commands as well; if this is done, then applications that use only
 the core Vulkan commands have no need to use fname:vkGetInstanceProcAddr.
 
-Function pointers to commands that don't operate on a specific instance can:
-be obtained by using this command with pname:instance equal to `NULL`. The
-following commands can: be accessed this way:
+The table below defines the various use cases for fname:vkGetInstanceProcAddr
+and expected behavior ("fp" is function pointer) for each case. Each rule
+is tested in the order below.
 
-  * flink:vkEnumerateInstanceExtensionProperties
-  * flink:vkEnumerateInstanceLayerProperties
-  * flink:vkCreateInstance
+.vkGetInstanceProcAddr behavior
+[width="80%",options="header",align="center"]
+|=====
+| pname:instance | pname:pName | behavior
+| invalid instance | * | undefined
+| * | `NULL` | undefined
+| `NULL` | flink:vkEnumerateInstanceExtensionProperties | returns fp
+| `NULL` | flink:vkEnumerateInstanceLayerProperties | returns fp
+| `NULL` | flink:vkCreateInstance | returns fp
+| `NULL` | * (not any row/case above) | returns `NULL`
+| instance | flink:vkEnumerateInstanceLayerProperties | `NULL`
+| instance | flink:vkEnumerateInstanceExtensionProperties | `NULL`
+| instance | flink:vkCreateInstance | `NULL`
+| instance | core Vulkan command | returns fp (1)
+| instance | enabled instance extension command for pname:instance | returns fp (1)
+| instance | available device extension command for pname:instance | returns fp (1,2)
+| instance | * (not any row/case above) | returns `NULL`
+|=====
 
-If pname:instance is a valid sname:VkInstance, function pointers to any
-commands that operate on pname:instance or a child of pname:instance can: be
-obtained. The returned function pointer must: only be called with a
-dispatchable object (the first parameter) that is a child of pname:instance.
+1) The returned function pointer must: only be called with a dispatchable object
+(the first parameter) that is pname:instance or a child of pname:instance (
+    i.e. sname:VkPhysicalDevice, sname:VkDevice, sname:VkQueue, or
+    sname:VkCommandBuffer).
 
-If pname:pName is not the name of a core Vulkan command, or is an
-extension command for any extension not supported by any available layer or
-implementation, then fname:vkGetInstanceProcAddr will return `NULL`.
-
-ifdef::editing-notes[]
-[NOTE]
-.editing-note
-====
-(Jon, Bug 14886 / Gitlab issue 4) The WSI group tentatively agreed that the
-WSI extensions were special, and should get static entry points in link
-libraries and prototypes in +vulkan.h+, while future extensions would have
-to be dynamically loaded. If this decision is upheld by the group as a
-whole, it would probably be encoded in the previous paragraph, in the WSI
-extensions branch of the Specification.
-
-However, this decision hasn't been fully signed off on by the entire Vulkan
-WG yet AFAIK. Note that implementations typically won't support many of the
-WSI extensions, so ``static entry points'' doesn't relieve apps of the
-neccessity of runtime enabling and testing of each extension before using
-it.
-====
-endif::editing-notes[]
+2) An available extension is an extension function supported by any of the loader, ICD or layer.
 
 In order to support systems with multiple Vulkan implementations
 comprising heterogeneous collections of hardware and software, the function
--- a/doc/specs/vulkan/validity/protos/vkGetInstanceProcAddr.txt
+++ b/doc/specs/vulkan/validity/protos/vkGetInstanceProcAddr.txt
@@ -9,8 +9,6 @@ Valid Usage
 endif::doctype-manpage[]
 * If pname:instance is not `NULL`, pname:instance must: be a valid sname:VkInstance handle
 * pname:pName must: be a null-terminated string
-* If pname:instance is `NULL`, pname:pName must: be one of: fname:vkEnumerateInstanceExtensionProperties, fname:vkEnumerateInstanceLayerProperties or fname:vkCreateInstance
-* If pname:instance is not `NULL`, pname:pName mustnot: be fname:vkEnumerateInstanceExtensionProperties, fname:vkEnumerateInstanceLayerProperties or fname:vkCreateInstance
 ifndef::doctype-manpage[]
 ********************************************************************************
 endif::doctype-manpage[]
--- a/src/spec/vk.xml
+++ b/src/spec/vk.xml
@@ -2994,17 +2994,11 @@ maintained in the master branch of the Khronos Vulkan Github project.
             <proto><type>PFN_vkVoidFunction</type> <name>vkGetDeviceProcAddr</name></proto>
             <param><type>VkDevice</type> <name>device</name></param>
             <param len="null-terminated">const <type>char</type>* <name>pName</name></param>
-            <validity>
-            </validity>
         </command>
         <command>
             <proto><type>PFN_vkVoidFunction</type> <name>vkGetInstanceProcAddr</name></proto>
             <param optional="true"><type>VkInstance</type> <name>instance</name></param>
             <param len="null-terminated">const <type>char</type>* <name>pName</name></param>
-            <validity>
-                <usage>If pname:instance is `NULL`, pname:pName must: be one of: fname:vkEnumerateInstanceExtensionProperties, fname:vkEnumerateInstanceLayerProperties or fname:vkCreateInstance</usage>
-                <usage>If pname:instance is not `NULL`, pname:pName mustnot: be fname:vkEnumerateInstanceExtensionProperties, fname:vkEnumerateInstanceLayerProperties or fname:vkCreateInstance</usage>
-            </validity>
         </command>
         <command>
             <proto><type>void</type> <name>vkGetPhysicalDeviceProperties</name></proto>


doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,22 +30,30 @@ Loaders are encouraged to export function symbols for all other core
...
+| NULL | * (any pname:pName not covered above) | NULL
+| instance | core Vulkan command | fp (1)
+| instance | enabled instance extension commands for pname:instance | fp (1)
+| instance | available device extension commands for pname:instance | fp (1,2)
+| instance | * (any pname:pName not covered above) | NULL
+|=====
+
+1) The returned function pointer must: only be called with a dispatchable object
+(the first parameter) that is a child of pname:instance.
+    i.e. sname:VkPhysicalDevice, sname:VkDevice, sname:VkQueue, or sname:VkCommandBuffer
```
krOoze: What about *instance* itself - not just child (and thus VkInstance in the list).

Not sure if exhaustive list is needed here (it should perhaps be in the earlier chapter that defines "dispatchable object").

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,22 +30,30 @@ Loaders are encouraged to export function symbols for all other core
...
-implementation, then fname:vkGetInstanceProcAddr will return `NULL`.
+The table below defines the various use cases for fname:vkGetInstanceProcAddr
+and expected return value ("fp" is function pointer) for each case.
+
+.vkGetInstanceProcAddr behavior
+[width="80%",options="header",align="center"]
+|=====
+| pname:instance | pname:pName | return value
+| * | NULL | undefined
+| invalid instance | * | undefined
```
krOoze: It's "undefined behavior", not "undefined return value". (Perhaps a null-pointer dereference or a segfault it those two cases.)

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,22 +30,30 @@ Loaders are encouraged to export function symbols for all other core
...
+| NULL | flink:vkEnumerateInstanceExtensionProperties | fp
+| NULL | flink:vkEnumerateInstanceLayerProperties | fp
+| NULL | flink:vkCreateInstance | fp
+| NULL | * (any pname:pName not covered above) | NULL
```
krOoze: perhaps "not in any row/case above" instead. (To be a nitpicker, there is a `*` "pname above" too).

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,22 +30,30 @@ Loaders are encouraged to export function symbols for all other core
...
+| instance | core Vulkan command | fp (1)
```
krOoze: Obviously not one of the commands above.

I see a row priority (or `switch`-like behavior, if you will). Maybe stating at start the rows are evaluated in order would simplify things (even that "pname above" thing above.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -78,18 +86,23 @@ command:
...
-    is one of
-  ** sname:VkDevice
-  ** sname:VkQueue
-  ** sname:VkCommandBuffer
-
-If pname:pName is not the name of one of these Vulkan commands, and is
-not the name of an extension command belonging to an extension enabled for
-pname:device, then fname:vkGetDeviceProcAddr will return `NULL`.
-
-include::../validity/protos/vkGetDeviceProcAddr.txt[]
```
krOoze: accidentally deleted this line?

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,22 +30,30 @@ Loaders are encouraged to export function symbols for all other core
...
+|=====
+
+1) The returned function pointer must: only be called with a dispatchable object
+(the first parameter) that is a child of pname:instance.
+    i.e. sname:VkPhysicalDevice, sname:VkDevice, sname:VkQueue, or sname:VkCommandBuffer
+
+2) available extension is an extension function supported by any of the loader, ICD or layer.
 
 ifdef::editing-notes[]
 [NOTE]
```
krOoze: BTW is this Note even true anymore?
The prototypes are in the `vulkan.h` and WSI extensions have to be loaded by the Khronos loader (per ch. C.14. Accessing Extension Functions from Programs).

Also what does it refer to anymore by "previous paragraph"?

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -78,18 +86,23 @@ command:
...
-
-include::../validity/protos/vkGetDeviceProcAddr.txt[]
+The table below defines the various use cases for fname:vkGetDeviceProcAddr
+and expected return value for each case.
+
+.vkGetDeviceProcAddr behavior
+[width="80%",options="header",align="center"]
+|=====
+| pname:device | pname:pName | return value
+| NULL | * | undefined
```
krOoze: I would be careful with the `NULL`. Are you sure implementations won't use 0 as the parameter? Besides, it already falls into the "invalid device" case below anyway.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -17,7 +17,7 @@ command:
 include::../protos/vkGetInstanceProcAddr.txt[]
 
   * pname:instance is the instance that the function pointer will be
-    compatible with.
+    compatible with, or `NULL` for commands not dependent on any instance.
```
krOoze: BTW if I think about it, shouldn't it use the `VK_NULL_HANDLE` instead of just `NULL`?

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,22 +30,30 @@ Loaders are encouraged to export function symbols for all other core
...
+| pname:instance | pname:pName | return value
+| * | NULL | undefined
+| invalid instance | * | undefined
+| NULL | flink:vkEnumerateInstanceExtensionProperties | fp
+| NULL | flink:vkEnumerateInstanceLayerProperties | fp
+| NULL | flink:vkCreateInstance | fp
+| NULL | * (any pname:pName not covered above) | NULL
+| instance | core Vulkan command | fp (1)
+| instance | enabled instance extension commands for pname:instance | fp (1)
+| instance | available device extension commands for pname:instance | fp (1,2)
```
krOoze: Shouldn't be plural. `pName` is only one command.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,22 +30,30 @@ Loaders are encouraged to export function symbols for all other core
...
+| instance | * (any pname:pName not covered above) | NULL
+|=====
+
+1) The returned function pointer must: only be called with a dispatchable object
+(the first parameter) that is a child of pname:instance.
+    i.e. sname:VkPhysicalDevice, sname:VkDevice, sname:VkQueue, or sname:VkCommandBuffer
+
+2) available extension is an extension function supported by any of the loader, ICD or layer.
```
krOoze: Inconsistent formatting: the 1) starts with capital letter and ends without a `.` and 2) exactly vice-versa

doc/specs/vulkan/chapters/initialization.txt
courtney-g: Sure. We can use VK_NULL_HANDLE here.

doc/specs/vulkan/chapters/initialization.txt
courtney-g: I'll change the "return value" column to "behavior". The behavior can then be "undefined" or "returns fp", etc.

doc/specs/vulkan/chapters/initialization.txt
courtney-g: Not sure I follow the nit picky comment: "there is a * "pName above" too".
I see an "invalid instance | * (pName)" case above, but that would only apply for invalid instance use and, by its nature, would have undefined behavior.

doc/specs/vulkan/chapters/initialization.txt
krOoze: That should be everywhere for handle types...
It is used correctly on some other commands (in VU section). There must be some bug in the script translating the `optional` `vk.xml` attribute, ending up  `NULL` instead for this command...

doc/specs/vulkan/chapters/initialization.txt
courtney-g: Thanks, will restore the validity include.

doc/specs/vulkan/chapters/initialization.txt
krOoze: >[...] , but that would only apply for invalid instance use [...]

That's why it's nit-picky. The reader does not know about this "but". I. e. strictly speaking:

**pName**s covered above = { `NULL`, `*`, "vkEnumerateInstanceExtensionProperties", "vkEnumerateInstanceLayerProperties", "vkCreateInstance" }

"any **pName** not covered above" = complement of above = {}

doc/specs/vulkan/chapters/initialization.txt
courtney-g: I'll remove this note. This is now covered elsewhere in the spec.

doc/specs/vulkan/chapters/initialization.txt
courtney-g: I'll delete this line as suggested.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -17,7 +17,7 @@ command:
 include::../protos/vkGetInstanceProcAddr.txt[]
 
   * pname:instance is the instance that the function pointer will be
-    compatible with.
+    compatible with, or `VK_NULL_HANDLE` for commands not dependent on any instance.
```
krOoze: I done research on the `NULL` vs `VK_NULL_HANDLE`:

The generator script specifically chooses NULL for dispatchable handles:
https://github.com/KhronosGroup/Vulkan-Docs/blob/5de77cf663c84cc2edfa355c49319dca0a869eb2/src/spec/generator.py#L1465

So not sure which is correct anymore... I guess the original author wanted them to have pointer-like semantics.

Oh well, it should be consistent, so either that needs to return back to `NULL` **or** rest of the spec needs to conform to this change.

courtney-g: Okay, good to know. I will use NULL to be consistent.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,41 +30,32 @@ Loaders are encouraged to export function symbols for all other core
...
+| instance | core Vulkan command | return fp (1)
+| instance | enabled instance extension command for pname:instance | return fp (1)
+| instance | available device extension command for pname:instance | return fp (1,2)
+| instance | * (not any row/case above) | return VK_NULL_HANDLE
+|=====
+
+1) The returned function pointer must: only be called with a dispatchable object
+(the first parameter) that is pname:instance or a child of pname:instance (
+    i.e. sname:VkPhysicalDevice, sname:VkDevice, sname:VkQueue, or
+    sname:VkCommandBuffer).
```
krOoze: Still should contain VkInstance in the (exhaustive) list.

Still similar sentence/requirement missing for vkGet**Device**ProcAddr

(Still am inclined, the list should be in the chapter where "dispatchable handle" term is defined)

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,41 +30,32 @@ Loaders are encouraged to export function symbols for all other core
...
+and expected behavior ("fp" is function pointer) for each case. Each rule
+is tested in the order below.
+
+.vkGetInstanceProcAddr behavior
+[width="80%",options="header",align="center"]
+|=====
+| pname:instance | pname:pName | behavior
+| invalid instance | * | undefined
+| * | VK_NULL_HANDLE | undefined
+| VK_NULL_HANDLE | flink:vkEnumerateInstanceExtensionProperties | return fp
```
krOoze: return**s**? (Or is it like a C statement `return x;`?)

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -78,19 +69,25 @@ command:
...
+
+.vkGetDeviceProcAddr behavior
+[width="80%",options="header",align="center"]
+|=====
+| pname:device | pname:pName | behavior
+| invalid device | * | undefined
+| device | VK_NULL_HANDLE | undefined
+| device | core Vulkan command (1) | return fp
+| device | enabled device extension commands for pname:device | return fp
+| device | * (any pname:pName not covered above) | return VK_NULL_HANDLE
```
krOoze: "**pName** above" changed for vkGet**Instance**ProcAddr, but not here

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,41 +30,32 @@ Loaders are encouraged to export function symbols for all other core
...
+[width="80%",options="header",align="center"]
+|=====
+| pname:instance | pname:pName | behavior
+| invalid instance | * | undefined
+| * | VK_NULL_HANDLE | undefined
+| VK_NULL_HANDLE | flink:vkEnumerateInstanceExtensionProperties | return fp
+| VK_NULL_HANDLE | flink:vkEnumerateInstanceLayerProperties | return fp
+| VK_NULL_HANDLE | flink:vkCreateInstance | return fp
+| VK_NULL_HANDLE | * (not any row/case above) | return VK_NULL_HANDLE
+| instance | core Vulkan command | return fp (1)
```
krOoze: still "core Vulkan command" would allow names of the three independent ones above. (sorry to mislead before - rule order won't help this case because of the differing first column)

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,41 +30,32 @@ Loaders are encouraged to export function symbols for all other core
...
+The table below defines the various use cases for fname:vkGetInstanceProcAddr
+and expected behavior ("fp" is function pointer) for each case. Each rule
+is tested in the order below.
+
+.vkGetInstanceProcAddr behavior
+[width="80%",options="header",align="center"]
+|=====
+| pname:instance | pname:pName | behavior
+| invalid instance | * | undefined
+| * | VK_NULL_HANDLE | undefined
```
krOoze: I think pName (cstring, not handle) isn't subject for VK_NULL_HANDLE either way.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -78,19 +69,25 @@ command:
...
+and expected behavior for each case, in the order defined by the table below.
+
+.vkGetDeviceProcAddr behavior
+[width="80%",options="header",align="center"]
+|=====
+| pname:device | pname:pName | behavior
+| invalid device | * | undefined
+| device | VK_NULL_HANDLE | undefined
+| device | core Vulkan command (1) | return fp
+| device | enabled device extension commands for pname:device | return fp
```
krOoze: plural fixed above (for `InstanceProc`), but not here

doc/specs/vulkan/chapters/initialization.txt
courtney-g: Are you asking that instances of "instance" in the table should be "VkInstance"?

courtney-g: > (Still am inclined, the list should be in the chapter where "dispatchable handle" term is defined instead)

That's tough. Not sure there is a "best" place that doesn't duplicate the info, which I don't think we want to do. Of the two places, definition of kinds of handles Vulkan can generate, and description of how to "initialize" Vulkan, I think it fits the "initialize" chapter better.

krOoze: Nevermind, I read it as the list in parentheses belonging to whole "**instance** or a child of **instance**", not just child. (Which would then need `sname:VkInstance` after the "i.e." too.)

doc/specs/vulkan/chapters/initialization.txt
courtney-g: Good catch, I'll add rows for those with a valid instance.

doc/specs/vulkan/chapters/initialization.txt
krOoze: > "best" place

Could be in glossary too. Though I personally hate definitions being (exclusively) there.

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -30,41 +30,35 @@ Loaders are encouraged to export function symbols for all other core
...
+| pname:instance | pname:pName | behavior
+| invalid instance | * | undefined
+| * | NULL | undefined
+| NULL | flink:vkEnumerateInstanceExtensionProperties | returns fp
+| NULL | flink:vkEnumerateInstanceLayerProperties | returns fp
+| NULL | flink:vkCreateInstance | returns fp
+| NULL | * (not any row/case above) | returns NULL
+| instance | flink:vkEnumerateInstanceLayerProperties | undefined
+| instance | flink:vkEnumerateInstanceExtensionProperties | undefined
+| instance | flink:vkCreateInstance | undefined
```
krOoze: "undefined"? That's bit of a surprise.
Supposed to be `NULL` instead? If not, then the second `<usage>` in `<validity>`  needs to stay.

courtney-g: Okay, it can be NULL. I had checked the code for both the Android and Windows/Linux loader and it looked like the desktop loader did not define this case, but I was mistaken. Both loaders will return NULL when given a non-NULL VkInstance and one of the three global functions.

krOoze:
Some version of this got merged with 1.0.23.

courtney-g:
Resolved with spec update 1.0.27.

PR #278, opened 2016-07-11
1.0 vk khr display swapchain

hustwarhd:


oddhack:
This looks like some sort of push accident; in any event I can't determine the intended purpose. If you have a problem that needs to be resolved with the tree, please open an issue.

PR #315, opened 2016-08-06
Correct description of vkCmdDrawIndexedIndirect

HildarTheDorf:
vkCmdDrawIndexedIndirect incorrectly claims it is an indirect version of vkCmdDrawIndirect instead of vkCmdDrawIndexed.


krOoze:
Seems fixed in 1.0.26 update.

Issue #323, opened 2016-08-09
Inferior table rendering quality in HTML ref pages


krOoze:
Tables are rendered oddly in html man targets. They are without borders and contents are center aligned(asciidoc default is supposed to be top aligned). Pdf man seems fine. Is there perhaps some css problem?

krOoze:
This was probably fixed with the move to asciidoctor. Closing.

Issue #304, opened 2016-07-30
Editorial issues: Image and image view parameter compatibility table


philiptaylor:
https://www.khronos.org/registry/vulkan/specs/1.0/xhtml/vkspec.html#resources-image-views says:

> Table 11.1. Image and image view parameter compatibility requirements
> ...
> `subResourceRange` and `viewType` must be compatible with the image, as described in the table below 

The table is above, not below. (And maybe it should refer to the table by name/number in case somebody is reading the spec without hyperlinks.)

---

The "View parameters" column refers to arrayLayers, but I think it should be layerCount instead, since presumably it's implicitly referring to the view's subresourceRange.

---

The table says:

> arrayLayers = 6×N

for cube arrays. It doesn't say N>0, so that implies arrayLayers = 0 is compatible, which doesn't sound right since the others all require arrayLayers >= 1. (Could be solved by #303 - if you can't create a subresource range with 0 layers then you can't have N=0 here.)

---

Each cell in the table is sensibly spread over multiple lines in the PDF spec, but squashed onto a single line in the HTML spec, which makes it a little bit harder to read.

krOoze:
Also shouldn't the first column use the proper enums instead?
And what is the purpose of the "sampled" in the table (there's no difference where applicable)?

oddhack:
I'll take a look at this. Most of it is straightforward. Sometimes convincing the various asciidoc toolchains to respect line breaks in the same way is not straightforward.

oddhack:
Most of this will get addressed this week. The "purpose of MS/samples" question from @krOoze is because the "Image parameters" cell is different for the MS/non-MS cases. The line breaks inside cell issue is going to be relegated to the magical land of "we hope this will be cleaned up by asciidoctor when we finally manage to move to that" because none of us want to put a lot of time into figuring out what has to be tweaked. Or we might move over to the asciidoc HTML5 output format, which does seem to handle this properly, in the interim.

krOoze:
@oddhack Well, obviously, but if I am not blind only the sample count changes. There are other Image parameters that could be in the first and change the second column (but are not). The question is, why is it important to have that there (i.e. why is it important WRT the ImageView -- what does it change in the third column)?

philiptaylor:
[...]

PR #324, opened 2016-08-09
Fixing formating of SPIRV-Image-ImageView table

krOoze:
from issue #304

 - Fixing formating so elements are on seprate lines across all build targets.
 - Adjusting witdth so contents fit well in man target.
 - Encapsulating stuff in math environment for format consistency
 - Moving `N` definition to same line where used

(I noticed there may be some trouble with tables generally in html man #323)


krOoze:

PR #355, opened 2016-08-29
Add bitmask type specification for NV vendor tag enums

discosultan:
Not sure if intentional or not, but the following enums were missing type specification attribute:

- VkExternalMemoryHandleTypeFlagBitsNV
- VkExternalMemoryFeatureFlagBitsNV


discosultan:
The missing type specifications have been added in v1.0.26 of the spec.

PR #363, opened 2016-09-10
GL_KHR_vulkan_glsl.txt Fix texture lookup functions.

Loftilus:


Ruslanis:
Vulka

TobiasHector:
@Loftilus This looks like a straightforward fix, but we can't accept this unless you accept the CLA - if you can accept that we can get this PR done. Thanks!

PR #380, opened 2016-09-24
Correct refBegin for VkDebugReportFlagBitsEXT

FacticiusVir:
The refBegin entry for VkDebugReportFlagBitsEXT is incorrectly set to VkDebugReportCallbackCreateInfoEXT


courtney-g:
This will be fixed in the next spec update.

FacticiusVir:
Fixed in 1.0.31

PR #410, opened 2016-11-15
Fixed wrong grammar

nubok:
"we have switch to" -> "we have switched to"


TobiasHector:
This got incidentally fixed at some point, so closing this PR.

PR #458, opened 2017-02-13
Adding chapter name into the pdf page footer

### LINKED ISSUE #2: Vulkan Specification lacks certain niceties in PDF format
In the OpenGL specification PDF, you can look at the page header to see exactly what section you are in. The Vulkan PDF seems more like a printout from a web-browser; the header has the name of the document, but not the name of each section on each page. That makes it hard when you link from one area to another to tell exactly where you are.

Also, the bookmarks on the Vulkan PDF lack chapter names. So if you find a section header that says `8.4`, you have to count the items in the bookmarks to find out the name of Chapter 8. The OpenGL PDF puts the chapter numbers before the chapter names in the PDF bookmarks.

krOoze:
Adding chapter name into the pdf page footer.

Fixes #2 

The used theme is the [default theme from Asciidoctor master](https://github.com/asciidoctor/asciidoctor-pdf/blob/master/data/themes/default-theme.yml) with only change to the footer.

### DIFF
--- a/doc/specs/vulkan/config/themes/pdf-theme.yml
+++ b/doc/specs/vulkan/config/themes/pdf-theme.yml
@@ -1,0 +1,275 @@
+# Modified asciidoctor-pdf default-theme.yml adding chapter names to the page footer
+
+font:
+  catalog:
+    # Noto Serif supports Latin, Latin-1 Supplement, Latin Extended-A, Greek, Cyrillic, Vietnamese & an assortment of symbols
+    Noto Serif:
+      normal: notoserif-regular-subset.ttf
+      bold: notoserif-bold-subset.ttf
+      italic: notoserif-italic-subset.ttf
+      bold_italic: notoserif-bold_italic-subset.ttf
+    # M+ 1mn supports ASCII and the circled numbers used for conums
+    M+ 1mn:
+      normal: mplus1mn-regular-ascii-conums.ttf
+      bold: mplus1mn-bold-ascii.ttf
+      italic: mplus1mn-italic-ascii.ttf
+      bold_italic: mplus1mn-bold_italic-ascii.ttf
+    # M+ 1p supports Latin, Latin-1 Supplement, Latin Extended, Greek, Cyrillic, Vietnamese, Japanese & an assortment of symbols
+    # It also provides arrows for ->, <-, => and <= replacements in case these glyphs are missing from font
+    M+ 1p Fallback:
+      normal: mplus1p-regular-fallback.ttf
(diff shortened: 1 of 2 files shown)


doc/specs/vulkan/config/themes/pdf-theme.yml
```diff
@@ -0,0 +1,273 @@
...
+      #content: '{page-number}'
+      content: '{section-or-chapter-title} | {page-number}'
+      #content: '{document-title} | {page-number}'
+    #center:
+    #  content: '{page-number}'
+  verso:
+    #columns: $footer_recto_columns
+    left:
+      #content: $footer_recto_right_content
+      content: '{page-number} | {section-or-chapter-title}'
```
TobiasHector: Looking through the spec, since sections don't always align to page boundaries, section titles look weird when you have shorter-than-a-page sections (which happens frequently). So I think I'd rather this was simply {chapter-title}.

TobiasHector: (Same comment for the right hand side as well)

krOoze: Hmm, that sucks. It kills half the usefulness of this.
I will change it to `{chapter-title}` for now.
But it could use a reformat so that some sections have their own page. E.g. part of the "1.4. Normative References" is awkwardly cut of to separate page, or the "2.3. Object Model" section has only one line on a page before it is cut off.

TobiasHector: Yea asciidoctor-pdf has several bugs related to "do better page breaking", section headers barely scratches the surface 😅 

But yea for now I think this is probably the right choice.

krOoze: FYI, I pushed the change. Personally bit annoying it is saying "Chapter" everywhere....

TobiasHector: Eh I actually like that :P

TobiasHector:
LGTM @oddhack - feel free to merge it in.

TobiasHector:
Fixes #2

TobiasHector:
@oddhack Was trying to trick this into autoclosing #2 when it's accepted, but apparently you can't do that when it's already been referenced previously or something... anyway we should close #2 when this is accepted.

krOoze:
@TobiasHector How so? Seems to work on my end. Says it will autoclose the issue.

TobiasHector:
@krOoze Weird... well, hopefully it'll work then.

PR #482, opened 2017-04-06
resources.txt: Fix typo

jeremy-lunarg:


oddhack:
I fixed this in internal gitlab and it will get pushed out in the next public spec update - thanks for catching it. The current process is either MR in gitlab, or issue on github (for non-Khronos-members), but we still don't have a CLA setup and still can't accept PRs on github :-(

Issue #489, opened 2017-04-08
[Extension] Ambiguous "should" about PFN_vkDebugReportCallbackEXT return


krOoze:
Spec says about debug callback return :

> The callback returns a `VkBool32` that indicates to the calling layer if the Vulkan call **should** be aborted or not.

Which would imply (per RFC "should") that the return might be ignored, and implementation might choose whether to abort or not.
But it feels the "should" in this sentence is colloquial (and in intention means "is supposed to").

In next paragraph it says:

> If the application returns `VK_TRUE` from its callback and the Vulkan call being aborted returns a `VkResult`, the layer will return `VK_ERROR_VALIDATION_FAILED_EXT`.

Which seem to make a different assumption about it than the above.
It is especially obvious from it not saying what happens if it returns `VK_FALSE` and call is being aborted (which is a possibility based on the first quote).

oddhack:
This should be fixed in the 1.0.50 spec update.

Issue #483, opened 2017-04-07
[Specification] Ambiguities in vkCreateInstance description


krOoze:
I find this paragraph about `vkCreateInstance` ambiguous:

> `vkCreateInstance` creates the instance, then enables and initializes global layers and extensions requested by the application. If an extension is provided by a layer, both the layer and extension **must** be specified at `vkCreateInstance` time. If a specified layer cannot be found, no `VkInstance` will be created and the function will return `VK_ERROR_LAYER_NOT_PRESENT`. Likewise, if a specified extension cannot be found the call will return `VK_ERROR_EXTENSION_NOT_PRESENT`.

1) Does the sentence about layer provided extension means:
a) If I want the extension I also have to enable the layer (unless the extension is also provided by the Implementation).
or
b) If I want the layer I also have to enable the extension.
or
c) (Read literaly but probably non-sense:) Any layer that provides any extension has to be enabled.

2) The first sentence should probably use the established "instance" instead of "global".

3) It feels there should not be "must" about the layer provided extensions, but such case should also return either `VK_ERROR_LAYER_NOT_PRESENT` or `VK_ERROR_EXTENSION_NOT_PRESENT`. Because layers (and their provided extensions) may change at runtime, so the Application would be unable to guarantee that requirement (and so avoid the undefined behavior).

4) It feels like the creation order should be opposite; i.e. "first initializes layers and extensions and then creates instance". Because I would want the `pNext` extension of `VkInstanceCreateInfo` to also potentially affect the instance creation itself (as it does in reality with e.g. `debug_report` extension or layers validating the input from the command start).

courtney-g:
#3 is interesting. The app needs to specify the extension and the associated layer for things to work so I think the must: is justified. This seems in conflict to other places where valid use allows for the implementation to return an error. In this case, for things to work the application does need to specify both the instance layer and the extension. If either is not found, the application will get a well defined error.
But if an extension is solely implemented by a layer and the app leaves off one, they will not get what they want.

krOoze:
> This seems in conflict to other places where valid use allows for the implementation to return an error.

Since some spec version there should be no more cases where VU disallows something that is also handled be an error `VkResult`.

It is not a VU statement (though almost certainly it should be if it requires anything of the Application).

But the underlying problem is the same. RFC "**must**" is "an absolute requirement". Consequences of breaking such requirement should be largely unspecified(\undefined).

courtney-g:
[...]

Issue #478, opened 2017-03-21
[Extension][Question] How *exactly* debug flags work?


krOoze:
1) What is the exact purpose of `VK_DEBUG_REPORT_DEBUG_BIT_EXT`?
In the spec there is only written:

> `VK_DEBUG_REPORT_DEBUG_BIT_EXT` indicates diagnostic information from the loader and layers.

which pretty much describes most of the messages.

I consider this flag part of a hierarchy (of seriousness): error > warning > perf. warning > info > debug. 
Is this correct interpretation?

2) The callback and `vkDebugReportMessageEXT` seems to accept any combination of the flags, but they seem quite mutually exclusive. Is allowing this intentional?

If so how is it supposed to be used? What purpose e.g. `ERROR | WARNING` message serve?

Also how does the filter work in that case? E.g. if I specify `ERROR` in `VkDebugReportCallbackCreateInfoEXT::flags` will the callback be called for `ERROR | WARNING` message, or not?

3) There is no guarantee the flags will be non-zero for the callback.
Is this also intentional to allow by the spec?

courtney-g:
The DEBUG_BIT is not intended for valid usage checks. It's intended as a way to share "interesting" information with the application. I don't see it used in the validation layers at this time, but the desktop loader does generate DEBUG_BIT reports with information about what it's doing, e.g. search path it's using to look for manifest files, ICDs, etc. This can be useful when checking loader behavior.

Similarly, VK_DEBUG_REPORT_INFORMATION_BIT_EXT is used by the validation layers and loader to remark on specific actions such as memory tracking info or what ICD(s) are being loaded.

The flags are independent. It's expected that most folks are only interested in ERROR and WARNING and, maybe, PERF_WARNING. Callbacks will only be generated if the bitwise AND of the mask and the callback flag is non-zero. E.g.  if you only indicate VK_DEBUG_REPORT_WARNING_BIT_EXT you will only see notifications that have that bit set.

krOoze:
> The DEBUG_BIT is not intended for valid usage checks. It's intended as a way to share "interesting" information with the application.

So my assumption is correct that it is like `INFORMATION_BIT` (except perhaps less important or something)? Or is it meant to be only (non-error) info about internal workings of the loader and layers?

I still feel the spec description could be clearer. Currently if taken literaly it describes any kind of notification including `ERROR` ones (they are diagnostic, and they do come from the layers, don't they).

[...]

PR #493, opened 2017-04-17
WIP: debug report & vkCreateInstance wording update

### LINKED ISSUE #490: [Registry] Missing "_CALLBACK" for debug report object type
`VK_DEBUG_REPORT_OBJECT_TYPE_DEBUG_REPORT_EXT`
should be
`VK_DEBUG_REPORT_OBJECT_TYPE_DEBUG_REPORT_CALLBACK_EXT`
to match the object name `VkDebugReportCallbackEXT`

umm, actually
`VK_DEBUG_REPORT_OBJECT_TYPE_DEBUG_REPORT_CALLBACK_EXT_EXT`
right?

courtney-g:
Fixes #490, #489, #483, #478


doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -172,14 +172,21 @@ include::../api/protos/vkCreateInstance.txt[]
...
+enabled layer, fname:vkCreateInstance must: return
 ename:VK_ERROR_EXTENSION_NOT_PRESENT.
+After verifying and enabling the instance layers and extensions the
+sname:VkInstance object is created and returned to the application.
+If a requested extension is only supported by a layer, both the layer and
+extension need to be specified at fname:vkCreateInstance time.
+For example, if an application wants to enable extensionB that is only
+supported in layerB, the application needs to include "layerB" in
+sname:ppEnabledLayerNames and "extensionB" in the
+sname:ppEnabledExtensionNames list.
```
krOoze: - This whole thing above renders as a single large monolithic paragraph -> unesthetical and hard to read.
 - If I understand things, the layer supported extension case description is now redundant to the descriptions above. It could be a Note block instead (would also help with the big paragraph problem)

courtney-g:
> Revision number in appendix is not incremented.

Not sure which revision number you are referring to. I updated the one in vk.xml for the debug_report extension.

> there possibly should be VU that VkDebugReportCallbackCreateInfoEXT::pfnCallback must be a pointer to a valid function (see VkAllocationCallbacks for examples). Should IMO be generated as Implicit VU instead though,

Probably should be implicit. That requires generator script updates. I'll put in manual VU for now.

> Missing pAllocator descriptions for create and destroy callback command
got it

> The VUs for pAllocator seem outdated in debug_report (compared e.g. to the VkInstance versions)

The VU is same as everywhere else it's auto generated. If you don't think that's appropriate, please file issue on it.

> Possibly copy-paste error for many vkDebugReportMessageEXT parameter descriptions (e.g. "messageCode is a layer-defined value indicating what test triggered this callback." — I would assume it is somewhat app-defined in this case as is the test)

I'll change location and messageCode to be application defined values for this call.

> The synchronization here (due to the nature of this extension) seem non-obvious. E.g. can I call vkDestroyDebugReportCallbackEXT from the callback function? Or on different thread while some vkX command (which is subject that can cause a callback) has not yet returned?

In the description of vkDestroyDebugReportCallbackEXT the 'callback' parameter is defined to be "externally synchronized" which is:
External synchronization
  A type of synchronization required of the application, where parameters defined to be externally synchronized must not be used simultaneously in multiple threads.

For this call, that may not be enough. Even if the application were to ensure that only one thread is calling vkDestroyDebugReportCallbackEXT making such a call during the callback could result in undefined behavior. Generally modifying a list while iterating through the list is not a good thing to do. I'll look at adding a VU for that.

> It is sneakily implied by the "bitwise-AND" description, but it would potentially be nice to say in PFN_vkDebugReportCallbackEXT::flags that it won't ever be 0

Why? Not obvious to me how adding such a statement is "potentially nice". To me it's another must: to be validated.

> vkCreateDevice should have similar description as vkCreateInstance now has.

updated

could also fix random minor issues in debug_report

> Many undecorated or wrong usage of "may" across the debug_report extension text.

You'll need to point those out. A lot of the "may"'s refer to example / speculative behavior.

> Undecorated and misnamed DebugReportCallback in pUserData parameter desc. of PFN_vkDebugReportCallbackEXT

I think I got it.

> there is "This extension define" instead "defines" in appendix.

Got it

> undecorated "RenderPass load_op"

I think I got it.

Rest as line comments.

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -79,14 +66,18 @@ include::../api/enums/VkDebugReportFlagBitsEXT.txt[]
...
+sname:VkDebugReportCallbackCreateInfoEXT::pname:flags determine
 when that function is called.
-A callback will be made for issues that match any bit set in its flags.
+When an event is created, the debug report extension will do a bitwise
+AND of the event's ename:VkDebugReportFlagBitsEXT flags to each
+sname:VkDebugReportCallbackEXT object's flags.
+For each non-zero result the corresponding callback will be called.
 The callback will come directly from the component that detected the event,
 unless some other layer intercepts the calls for its own purposes (filter
 them in different way, log to system error log, etc.) An application may
```
krOoze: @courtney-g That's perhaps for internal discussion.

BUT IMO RFC/Terminology chapter does not define any difference between decorated and non-decorated word &mdash; it simply says "The key words [keywords] in this document are to be interpreted as [...]."

Therfore (IMO) failing to decorate the keywords are just an violation of chosen style of this spec.
And using them colloquially (whether decorated or not) is violation of the Terminology chapter.

krOoze:
> Not sure which revision number you are referring to.

Right, sorry. I mean this one specifically (and also the "last modified date" there too):
https://github.com/KhronosGroup/Vulkan-Docs/blob/170410-debug_report_issues/doc/specs/vulkan/appendices/VK_EXT_debug_report.txt#L13


> > The VUs for pAllocator seem outdated in debug_report (compared e.g. to the VkInstance versions)
> 
> The VU is same as everywhere else it's auto generated. If you don't think that's appropriate, please file issue on it.

Sorry, I must have misread something.

Nevertheless there seem to be a copy-paste error in `vkDestroyDebugReportCallbackEXT()` VU for it.
"If VkAllocationCallbacks were provided when **instance** was created [...]"
Should be `callback` instead od `instance`? Elsewhere in the spec it seems to always be the object being vkDestroyed (even for non-dispatchables like `VkFence`)

> > It is sneakily implied by the "bitwise-AND" description, but it would potentially be nice to say in PFN_vkDebugReportCallbackEXT::flags that it won't ever be 0
>
> Why? Not obvious to me how adding such a statement is "potentially nice". To me it's another must: to be validated.

Well, it just seemed like a surprising implication of the `&` (events with `0` flag won't pass through no matter the filter flag). Also someone reading just the `PFN_vkDebugReportCallbackEXT` definition, which comes later, might miss that info &mdash; other parameters seem to have described there what they can and cannot be (except perhaps `pLayerPrefix` which should say it's null-terminated string, to disallow `nullptr` there)

It should not require another validation. As an Vulkan output it is not a subject for layer validation.

> You'll need to point those out. A lot of the "may"'s refer to example / speculative behavior.

That's perhaps for your internal discussion with the Group.

I mean all of them undecorated ones.

IMO RFC/Terminology chapter does not define any difference between decorated and non-decorated word &mdash; it simply says "The key **words** [list of keywords] in this document are to be interpreted as [...]."

Therfore my thinking is:
Failing to decorate any keyword is just a violation of the chosen style guide for this spec.
And using them colloquially (whether decorated or not) is a violation of the Terminology chapter.

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -223,14 +231,11 @@ include::../api/protos/vkDebugReportMessageEXT.txt[]
   * pname:objectType is a elink:VkDebugReportObjectTypeEXT specifying the
```
krOoze: These still feel copy-pasted and not applicable to `vkDebugReportMessageEXT`.
Should be something like:

> * `instance` is the debug stream's `VkInstance`.
> * `flags` indicates the `VkDebugReportFlagBitsEXT` classification of this event/message.

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -223,14 +231,11 @@ include::../api/protos/vkDebugReportMessageEXT.txt[]
...
     associated with the event.
-  * pname:location is a component (layer, driver, loader) defined value that
-    indicates the _location_ of the trigger.
-    This is an optional value.
-  * pname:messageCode is a layer-defined value indicating what test
-    triggered this callback.
+  * pname:location is an application defined value.
+  * pname:messageCode is an application defined value.
   * pname:pLayerPrefix is the abbreviation of the component making the
     callback.
```
krOoze: Same problem. Should be something like "making this event/message" instead "making the callback".

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -172,14 +172,16 @@ include::../api/protos/vkCreateInstance.txt[]
...
+fname:vkCreateInstance verifies that the requested layers exist. If not,
+fname:vkCreateInstance will return ename:VK_ERROR_LAYER_NOT_PRESENT.
+Next fname:vkCreateInstance verifies that the requested extensions are
+supported (e.g. in implementation or enabled instance layers) and if
+any requested extension is not supported, fname:vkCreateInstance must: return
 ename:VK_ERROR_EXTENSION_NOT_PRESENT.
+After verifying and enabling the instance layers and extensions the
+sname:VkInstance object is created and returned to the application.
+If a requested extension is only supported by a layer, both the layer and
+extension need to be specified at fname:vkCreateInstance time.
```
krOoze: Would add " , for the creation to succeed." at the end to emphasize this is not a requirement. (Same for the `vkCreateDevice` copy)

Should be "and **the** extension"?

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -124,11 +128,15 @@ include::../api/funcpointers/PFN_vkDebugReportCallbackEXT.txt[]
   * pname:pMessage is a null-terminated string detailing the trigger
```
krOoze: Should perhaps also say pname:pLayerPrefix is a null-terminated string (i.e. it cannot be e.g. `nullptr` I assume)

courtney-g:
FYI - pushed updated CLs

doc/specs/vulkan/chapters/devsandqueues.txt
```diff
@@ -560,6 +560,14 @@ include::../api/protos/vkCreateDevice.txt[]
   * pname:pDevice points to a handle in which the created sname:VkDevice is
     returned.
 
+fname:vkCreateDevice verifies that the requested extensions are supported (e.g.
+in implementation or enabled instance layer) and if any requested extension is
+not supported, fname:vkCreateDevice must: return ename:VK_ERROR_EXTENSION_NOT_PRESENT.
+After verifying and enabling the extensions the sname:VkDevice object is created
+and returned to the application.
+If a requested extension is only supported by a layer, the
+layer needs to be specified at flink:vkCreateInstance time.
```
krOoze: should probably match the `vkCreateInstance` wording, i.e. the changed:
"(e.g. in the implementation or in any enabled instance layer)"
and
"[...] for the creation to succeed."

doc/specs/vulkan/chapters/initialization.txt
```diff
@@ -172,14 +172,17 @@ include::../api/protos/vkCreateInstance.txt[]
...
+fname:vkCreateInstance verifies that the requested layers exist. If not,
+fname:vkCreateInstance will return ename:VK_ERROR_LAYER_NOT_PRESENT.
+Next fname:vkCreateInstance verifies that the requested extensions are
+supported (e.g. in the implementation or in any enabled instance layer) and if
+any requested extension is not supported, fname:vkCreateInstance must: return
 ename:VK_ERROR_EXTENSION_NOT_PRESENT.
+After verifying and enabling the instance layers and extensions the
+sname:VkInstance object is created and returned to the application.
+If a requested extension is only supported by a layer, both the layer and
+the extension need to be specified at fname:vkCreateInstance time, for the
```
krOoze: Not an englishman, but shouldn't this be without `,` (because no verb)?

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -217,22 +223,19 @@ To inject its own messages into the debug stream, call:
...
-    this callback.
+  * pname:instance is the debug stream's sname:VkInstance.
+  * pname:flags indicates the ename:VkDebugReportFlagBitsEXT classification of
+    this event/message.
   * pname:objectType is a elink:VkDebugReportObjectTypeEXT specifying the
     type of object being used or created at the time the event was
     triggered.
-  * pname:object is object where the issue was detected.
+  * pname:object object where the issue was detected.
     pname:object may be ename:VK_NULL_HANDLE if there is no object
```
courtney-g: Yeah, that's an appropriate spot for can:

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -217,22 +223,19 @@ To inject its own messages into the debug stream, call:
...
-  * pname:messageCode is a layer-defined value indicating what test
-    triggered this callback.
-  * pname:pLayerPrefix is the abbreviation of the component making the
-    callback.
+  * pname:location is an application defined value.
+  * pname:messageCode is an application defined value.
+  * pname:pLayerPrefix is the abbreviation of the component making this
+    event/message.
   * pname:pMessage is a null-terminated string detailing the trigger
     conditions.
```
courtney-g: Right, still apply flags.

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -257,12 +260,18 @@ include::../api/protos/vkDestroyDebugReportCallbackEXT.txt[]
 
  * pname:instance the instance where the callback was created.
  * pname:callback the sname:VkDebugReportCallbackEXT object to destroy.
+   pname:callback is an externally synchronized object and cannot be used
+   on more than one thread at a time. This means that
+   fname:vkDestroyDebugReportCallbackEXT cannot be called when a callback
```
courtney-g: ok

doc/specs/vulkan/chapters/devsandqueues.txt
courtney-g: ok

doc/specs/vulkan/chapters/initialization.txt
courtney-g: ok

courtney-g:
CLs updated.

doc/specs/vulkan/chapters/devsandqueues.txt
```diff
@@ -560,6 +560,15 @@ include::../api/protos/vkCreateDevice.txt[]
   * pname:pDevice points to a handle in which the created sname:VkDevice is
     returned.
 
+fname:vkCreateDevice verifies that the requested extensions are supported (e.g.
+the in implementation or in any enabled instance layer) and if any requested extension is
```
krOoze: typo: "the in"

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -111,24 +113,28 @@ include::../api/funcpointers/PFN_vkDebugReportCallbackEXT.txt[]
     type of object being used or created at the time the event was
     triggered.
   * pname:object gives the object where the issue was detected.
-    pname:object may be ename:VK_NULL_HANDLE if there is no object
+    pname:object can: be ename:VK_NULL_HANDLE if there is no object
```
krOoze: Wrong spot. This one should probably be "may:" (because it is basically output from Vulkan).

doc/specs/vulkan/chapters/VK_EXT_debug_report.txt
```diff
@@ -217,27 +223,24 @@ To inject its own messages into the debug stream, call:
...
-    this callback.
+  * pname:instance is the debug stream's sname:VkInstance.
+  * pname:flags indicates the ename:VkDebugReportFlagBitsEXT classification of
+    this event/message.
   * pname:objectType is a elink:VkDebugReportObjectTypeEXT specifying the
     type of object being used or created at the time the event was
     triggered.
-  * pname:object is object where the issue was detected.
+  * pname:object object where the issue was detected.
     pname:object may be ename:VK_NULL_HANDLE if there is no object
```
krOoze: I meant "can:" here.

TobiasHector:
@courtney-g it looks like this got fixed internally a few months back with what appears to be similar, if not the same, changes. Not sure what happened, but think we can close this? Can you confirm and close if true?

PR #502, opened 2017-05-10
Fix references to VK_ERROR_DEVICE_LOST

dennishamester:
The specification calls it VK_DEVICE_LOST in several places.


TobiasHector:
@dennishamester Looks good to me - if you're able to accept the CLA, we should be able to accept this.

PR #506, opened 2017-05-15
add support 'external_memory_win32(KHX,NV)'

techlabxe:
Fixed not working properly when using extension 'external_memory_win32(NV, KHX)'
I was in trouble when using validation layer, so I added it easily.


TobiasHector:
This got incidentally fixed in the spec internally at some point, so closing.

PR #535, opened 2017-07-26
Fix use of FlagBits to Flags types.

e1ner:
Building an API binding for Rust through generation rather than manually typing made me aware of some irregularities in the use of the FlagBits rather than the corresponding Flags types for structure members and command arguments.

Assuming I understand the differentiation between FlagBits and Flags types correctly, as in the FlagBits types are only the enumerations of the available bits and the Flags types are the types containing the actually selected bits.
That's at least how i understand and have map it while generating the Rust types. The FlagBits are mapped to enumerations (with a u32 representation) and the Flags are mapped to the u32 type.

### DIFF
--- a/src/spec/vk.xml
+++ b/src/spec/vk.xml
@@ -736,7 +736,7 @@ private version is maintained in the 1.0 branch of the member gitlab server.
             <member><type>VkExtent3D</type>             <name>extent</name></member>
             <member><type>uint32_t</type>               <name>mipLevels</name></member>
             <member><type>uint32_t</type>               <name>arrayLayers</name></member>
-            <member><type>VkSampleCountFlagBits</type>  <name>samples</name></member>
+            <member><type>VkSampleCountFlags</type>  <name>samples</name></member>
             <member><type>VkImageTiling</type>          <name>tiling</name></member>
             <member><type>VkImageUsageFlags</type>      <name>usage</name><comment>Image usage flags</comment></member>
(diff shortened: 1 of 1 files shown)


krOoze:
Huh, the CLA works already. Nice.

Anyway, I think your understanding is not entirely correct.

`*Flags` and `FlagBits` are same type in C, so there's no type-system advantage there. But there is a semantic difference.

`*Flags` is used where bitwise-OR of the `*FlagBits` is expected. But sometimes the API expects it to be used as an classic enum. In such a case it accepts `*FlagBits` (as in only one flag). It is used in places where only one value/bit is **fundamentally** expected. By "fundamentally" I mean such places, where even if new enumerants are added in the future, the Vulkan command will still only accept (exactly) one bit.

The `VkSampleCountFlags` and `VkSampleCountFlagBits` is a good example. `VkImageCreateInfo` expects and will always expect only one bit (representing concrete amount of samples), i.e. accepts `VkSampleCountFlagBits`. But `VkSampleCountFlags` is still used as flag-set where some bitwise-OR of the `VkSampleCountFlagBits` e.g. represends supported sample count values in `VkPhysicalDeviceLimits`.

e1ner:
Ok, that I didn't really think of and makes absolutely sense.
In that case I'll just change the way I generate the bindings.

oddhack:
@krOoze is correct - when a FlagBits type appears, it's because only a single bit value can be accepted. So I'm going to close this PR.

PR #547, opened 2017-08-22
Clarify pointer autovalidity and threading

krOoze:
1) State more explicitly in VUs if "valid" pointer is required
It is bit unnecessary for VU statements that say "point to a valid". But is more clear for statements that say e.g. "`pInstance` **must** be a pointer to a `VkInstance` handle" (which, literaly, it is by definition &mdash; that's the static type of the pointer)

2) Implicit validity is quite explicit for some time. Update pointer section as such.

3) Sometimes pointer parameter can be ignored. Update implicit pointer validity section as such.

4) Changing to more forceful language, to clarify pointers need synchronizing too, which is similar to "externally synchronized" objects.

### DIFF
--- a/doc/specs/vulkan/chapters/fundamentals.txt
+++ b/doc/specs/vulkan/chapters/fundamentals.txt
@@ -268,10 +268,12 @@ frees.
 It is an application's responsibility to track the lifetime of Vulkan
 objects, and not to destroy them while they are still in use.
 
-Application-owned memory is immediately consumed by any Vulkan command it is
-passed into.
-The application can: alter or free this memory as soon as the commands that
-consume it have returned.
+[[fundamentals-objectmodel-lifetime-acquire]]
+The ownership of application-owned memory is immediately acquired by any
+Vulkan command it is passed into.
+Ownership of such memory must: be released back to the application at the
+end of the duration of the command, so that the application can: alter or
+free this memory as soon as all the commands that acquired it have returned.
 
 The following object types are consumed when they are passed into a Vulkan
 command and not further accessed by the objects they are used to create.
@@ -600,6 +602,16 @@ library) perform memory barriers as a part of mutual exclusion, so mutexing
 Vulkan objects via these primitives will have the desired effect.
 ====
 
+Similarly the application must: avoid any potential data hazard of
+application-owned memory that has its
+<<fundamentals-objectmodel-lifetime-acquire,ownership temporarily acquired>>
+by a Vulkan command.
+While the ownership of application-owned memory remains acquired by a
+command the implementation may: read the memory at any point, and it may:
+write non-code:const qualified memory at any point.
+Parameters referring to non-code:const qualified application-owned memory
+are not marked explicitly as _externally synchronized_ in the specification.
+
 Many object types are _immutable_, meaning the objects cannot: change once
 they have been created.
 These types of objects never need external synchronization, except that they
@@ -754,13 +766,13 @@ commands, which will silently ignore these values.
 [[fundamentals-validusage-pointers]]
 ==== Valid Usage for Pointers
 
-Any parameter that is a pointer must: either be a valid pointer, or if
-_explicitly called out in the specification_, can: be `NULL`.
-A pointer is valid if it points at memory containing values of the number
-and type(s) expected by the command, and all fundamental types accessed
-through the pointer (e.g. as elements of an array or as members of a
-structure) satisfy the alignment requirements of the host processor.
+Any parameter that is a pointer must: be a _valid pointer_ only if it is
+explicitly called out by a Valid Usage statement.
 
+A pointer is "`valid`" if it points at memory containing values of the
+number and type(s) expected by the command, and all fundamental types
+accessed through the pointer (e.g. as elements of an array or as members of
+a structure) satisfy the alignment requirements of the host processor.
 
 [[fundamentals-validusage-strings]]
 ==== Valid Usage for Strings
--- a/doc/specs/vulkan/chapters/memory.txt
+++ b/doc/specs/vulkan/chapters/memory.txt
@@ -60,13 +60,13 @@ include::../api/structs/VkAllocationCallbacks.txt[]
 .Valid Usage
 ****
   * [[VUID-VkAllocationCallbacks-pfnAllocation-00632]]
-    pname:pfnAllocation must: be a pointer to a valid user-defined
+    pname:pfnAllocation must: be a valid pointer to a valid user-defined
     tlink:PFN_vkAllocationFunction
   * [[VUID-VkAllocationCallbacks-pfnReallocation-00633]]
-    pname:pfnReallocation must: be a pointer to a valid user-defined
+    pname:pfnReallocation must: be a valid pointer to a valid user-defined
     tlink:PFN_vkReallocationFunction
   * [[VUID-VkAllocationCallbacks-pfnFree-00634]]
-    pname:pfnFree must: be a pointer to a valid user-defined
+    pname:pfnFree must: be a valid pointer to a valid user-defined
     tlink:PFN_vkFreeFunction
   * [[VUID-VkAllocationCallbacks-pfnInternalAllocation-00635]]
     If either of pname:pfnInternalAllocation or pname:pfnInternalFree is not
--- a/src/spec/validitygenerator.py
+++ b/src/spec/validitygenerator.py
@@ -348,9 +348,9 @@ class ValidityOutputGenerator(OutputGenerator):
             if (lengths[0]) == 'null-terminated':
                 asciidoc += 'a null-terminated '
             elif (lengths[0]) == '1':
-                asciidoc += 'a pointer to '
+                asciidoc += 'a valid pointer to '
             else:
-                asciidoc += 'a pointer to an array of '
+                asciidoc += 'a valid pointer to an array of '
 
                 # Handle equations, which are currently denoted with latex
                 if 'latexmath:' in lengths[0]:
@@ -363,9 +363,9 @@ class ValidityOutputGenerator(OutputGenerator):
                 if (length) == 'null-terminated': # This should always be the last thing. If it ever isn't for some bizarre reason, then this will need some massaging.
                     asciidoc += 'null-terminated '
                 elif (length) == '1':
-                    asciidoc += 'pointers to '
+                    asciidoc += 'valid pointers to '
                 else:
-                    asciidoc += 'pointers to arrays of '
+                    asciidoc += 'valid pointers to arrays of '
                     # Handle equations, which are currently denoted with latex
                     if 'latexmath:' in length:
                         asciidoc += length
@@ -416,7 +416,7 @@ class ValidityOutputGenerator(OutputGenerator):
             # Could be multi-level pointers (e.g. ppData - pointer to a pointer). Handle that.
             asciidoc += 'a '
             for i in range(0, pointercount):
-                asciidoc += 'pointer to a '
+                asciidoc += 'valid pointer to a '
 
             # Handle void* and pointers to it
             if paramtype.text == 'void':
--- a/doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
+++ b/doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
@@ -299,7 +299,7 @@ ifdef::VK_KHR_shared_presentable_image[]
 endif::VK_KHR_shared_presentable_image[]
   * [[VUID-VkSwapchainCreateInfoKHR-imageSharingMode-01277]]
     If pname:imageSharingMode is ename:VK_SHARING_MODE_CONCURRENT,
-    pname:pQueueFamilyIndices must: be a pointer to an array of
+    pname:pQueueFamilyIndices must: be a valid pointer to an array of
     pname:queueFamilyIndexCount basetype:uint32_t values
   * [[VUID-VkSwapchainCreateInfoKHR-imageSharingMode-01278]]
     If pname:imageSharingMode is ename:VK_SHARING_MODE_CONCURRENT,
--- a/doc/specs/vulkan/chapters/cmdbuffers.txt
+++ b/doc/specs/vulkan/chapters/cmdbuffers.txt
@@ -504,7 +504,7 @@ recorded into it, becomes <<commandbuffers-lifecycle, invalid>>.
     All elements of pname:pCommandBuffers must: not be in the
     <<commandbuffers-lifecycle, pending state>>
   * [[VUID-vkFreeCommandBuffers-pCommandBuffers-00048]]
-    pname:pCommandBuffers must: be a pointer to an array of
+    pname:pCommandBuffers must: be a valid pointer to an array of
     pname:commandBufferCount sname:VkCommandBuffer handles, each element of
     which must: either be a valid handle or `NULL`
 ****
--- a/doc/specs/vulkan/chapters/descriptorsets.txt
+++ b/doc/specs/vulkan/chapters/descriptorsets.txt
@@ -747,7 +747,7 @@ avoid wasted memory.
     If pname:descriptorType is ename:VK_DESCRIPTOR_TYPE_SAMPLER or
     ename:VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, and
     pname:descriptorCount is not `0` and pname:pImmutableSamplers is not
-    `NULL`, pname:pImmutableSamplers must: be a pointer to an array of
+    `NULL`, pname:pImmutableSamplers must: be a valid pointer to an array of
     pname:descriptorCount valid sname:VkSampler handles
   * [[VUID-VkDescriptorSetLayoutBinding-descriptorCount-00283]]
     If pname:descriptorCount is not `0`, pname:stageFlags must: be a valid
@@ -1665,7 +1665,7 @@ in pname:pDescriptorSets are invalid.
     All submitted commands that refer to any element of
     pname:pDescriptorSets must: have completed execution
   * [[VUID-vkFreeDescriptorSets-pDescriptorSets-00310]]
-    pname:pDescriptorSets must: be a pointer to an array of
+    pname:pDescriptorSets must: be a valid pointer to an array of
     pname:descriptorSetCount sname:VkDescriptorSet handles, each element of
     which must: either be a valid handle or dlink:VK_NULL_HANDLE
   * [[VUID-vkFreeDescriptorSets-pDescriptorSets-00311]]
@@ -1838,19 +1838,19 @@ bindings as needed to update all pname:descriptorCount descriptors.
     ename:VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE,
     ename:VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, or
     ename:VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT, pname:pImageInfo must: be a
-    pointer to an array of pname:descriptorCount valid
+    valid pointer to an array of pname:descriptorCount valid
     sname:VkDescriptorImageInfo structures
   * [[VUID-VkWriteDescriptorSet-descriptorType-00323]]
     If pname:descriptorType is ename:VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER
     or ename:VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER, pname:pTexelBufferView
-    must: be a pointer to an array of pname:descriptorCount valid
+    must: be a valid pointer to an array of pname:descriptorCount valid
     sname:VkBufferView handles
   * [[VUID-VkWriteDescriptorSet-descriptorType-00324]]
     If pname:descriptorType is ename:VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER,
     ename:VK_DESCRIPTOR_TYPE_STORAGE_BUFFER,
     ename:VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC, or
     ename:VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC, pname:pBufferInfo must:
-    be a pointer to an array of pname:descriptorCount valid
+    be a valid pointer to an array of pname:descriptorCount valid
     sname:VkDescriptorBufferInfo structures
   * [[VUID-VkWriteDescriptorSet-descriptorType-00325]]
     If pname:descriptorType is ename:VK_DESCRIPTOR_TYPE_SAMPLER or
@@ -2434,8 +2434,8 @@ include::../api/protos/vkUpdateDescriptorSetWithTemplateKHR.txt[]
 .Valid Usage
 ****
   * [[VUID-vkUpdateDescriptorSetWithTemplateKHR-pData-01685]]
-    pname:pData must: be a pointer to a memory that contains one or more
-    valid instances of slink:VkDescriptorImageInfo,
+    pname:pData must: be a valid pointer to a memory that contains one or
+    more valid instances of slink:VkDescriptorImageInfo,
     slink:VkDescriptorBufferInfo, or slink:VkBufferView in a layout defined
     by pname:descriptorUpdateTemplate when it was created with
     flink:vkCreateDescriptorUpdateTemplateKHR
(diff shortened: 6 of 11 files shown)


doc/specs/vulkan/chapters/fundamentals.txt
```diff
@@ -286,8 +286,8 @@ objects, and not to destroy them while they are still in use.
 
 Application-owned memory is immediately consumed by any Vulkan command it is
 passed into.
-The application can: alter or free this memory as soon as the commands that
-consume it have returned.
+The application must: not write to nor free this memory before the command
+that consumed it have returned.
```
Tobski: What's the reasoning for this change? This is practically how *all* functions work in any API, doesn't feel like it needs explicitly stating. The original statement was to clarify that pointers are not held onto - your new statement sort of re-iterates that you can't do something you can't really do in any C API, removing the original intent. I'd be ok if found a way to express both notions though, I just don't want to lose the original intent.

src/spec/validitygenerator.py
```diff
@@ -348,9 +348,9 @@ def createValidationLineForParameterIntroChunk(self, blockname, param, params, t
             if (lengths[0]) == 'null-terminated':
                 asciidoc += 'a null-terminated '
             elif (lengths[0]) == '1':
-                asciidoc += 'a pointer to '
+                asciidoc += 'a valid pointer to '
```
Tobski: This actually was how it was previously written, but the options are limited to:
"it's a pointer to X", "it's NULL", or "it's a nonsensical value"

The first is already implicitly "valid" as it implicitly is not the other two things.
The second is covered by stating that it "must: be NULL or...".
The third is covered by having no VU.

I'm not against the change particularly, in fact I coded it this way when I first added the implicit VU scripts, but I don't really see that it adds anything, and other group members have stronger opinions about this AFAIK.
What's your reasoning for this change?

doc/specs/vulkan/chapters/memory.txt
```diff
@@ -1721,6 +1721,7 @@ ifdef::VK_KHX_device_group[]
   * [[VUID-vkMapMemory-memory-00683]]
     pname:memory must: not have been allocated with multiple instances.
 endif::VK_KHX_device_group[]
+  * pname:ppData must: be a valid pointer to a pointer
```
Tobski: Isn't this what would be generated automatically by your script change anyway?

TobiasHector:
Whoops. Accidentally reviewed with my personal account. @Tobski is me.

doc/specs/vulkan/chapters/memory.txt
krOoze: autogenerated original would be:
> pname:ppData must: be a pointer to a pointer

autogenerated new version would be:
> pname:ppData must: be a valid pointer to a valid pointer

Both would be lies.
As the manual words say, the outer pointer does have to be "valid" the inner does not.

Autogenerated words aside, it is a matter of semantics of the `vk.xml`. Autovalidity would imply both the outer and inner pointer have to be "valid" in some sense. You are silently relying on `void*` with implicit `len==1` to be valid by some too smartistic fashion (little comparable to SFINAE).
Which fails, because elsewhere you make a different semantic for `void*`, that it is "a pointer to bytes" and **not** "pointer to nothing", e.g. `VkPipelineCacheCreateInfo`:

> Valid Usage (Implicit)
>------------------------
> If `initialDataSize` is not `0`, `pInitialData` **must** be a pointer to an array of `initialDataSize` bytes

TobiasHector: > autogenerated new version would be:

> > pname:ppData must: be a valid pointer to a valid pointer

> Both would be lies.
> As the manual words say, the outer pointer does have to be "valid" the inner does not.

Huh. From looking at your script changes I'd have expected it to be what you wrote for the explicit VU. If you've tested this and are sure, fine - though it'd be nicer for it to not do this weirdness automatically!

src/spec/validitygenerator.py
krOoze: By current unspoken convention of the spec just the lack of "valid" prefix implies the object is exempt from validity. So this makes sense to me even from the consistency perspective.

krOoze: Besides, you are using "insider knowledge".

The reader does not necessarily caries in the head, what the autovalidity generator will spit out and not under what circumstances.

I could equally say for e.g.:

> `renderPass` **must** be a <s>valid</s> VkRenderPass handle

"Ok, there would be no VU statement if `renderPass` can be nonsense value, ergo I do not need to say \'valid\' in the VU if it does have to be valid handle"

doc/specs/vulkan/chapters/memory.txt
krOoze: There's a hack in the script. It reuses the `"pointer to "` string and erases the last 4 char. I added "valid", so it would end up "valid pointer". Somebody should clean that up eventually, but yeah...

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: Well, I will try to

> you can't really do in any C API

I wish. You **can** though. And if some concurrent situations it does even make sense.
It would made some sense if a `VkFence` was signalled, and there would be a doubt whether the pointer is still needed by the command even if it hasn't already returned.
For that matter you can even write `const` in C API.

doc/specs/vulkan/chapters/memory.txt
krOoze: PS Actually e.g. `vkUpdateDescriptorSetWithTemplateKHR` is a freak case too. Generates just:

> `pData` **must** be a pointer

That just sounds wrong.

Perhaps the `vk.xml` parsing script should throw error/exception when there is a `void*` without `len` && without `noautovalidity.

TobiasHector: > There's a hack in the script. It reuses the "pointer to " string and erases the last 4 char.

Oh yea I remember that. Knew it would bite me in the ass one day D:

> I could make the script do it implicitly, but considering the weirdness of void and that it is a one occurence of void* without len it is perhaps better of as explicit validity.

Yea that's fair.

> Perhaps the vk.xml parsing script should throw error/exception when there is a void* without len && without noautovalidity.

Hmm. Let me have a think on that one, you're probably right though.

src/spec/validitygenerator.py
TobiasHector: > By current unspoken convention of the spec just the lack of "valid" prefix implies the object is exempt from validity. So this makes sense to me even from the consistency perspective.

Reasonable. Consistency should probably trump redundancy in this case.

> Besides, you are using "insider knowledge".

Perhaps :)

Like I said - I'm fine with this change, it's the rest of the WG I'm asking on behalf of since we changed it away from this in the past.

doc/specs/vulkan/chapters/fundamentals.txt
TobiasHector: > I wish. You can though.

Yea ok that's actually fair. If we can find a way to say both things sanely, I'd be quite happy.

doc/specs/vulkan/chapters/memory.txt
krOoze: TBH `void*` with `len` should have perhaps been `uint8_t`, but I guess too late to change that (till Vulkan 2.0).

TobiasHector: Eh. I mean we at least differentiated in the parameter names (size vs. count) - there are times when you just don't know what the data is in C, and I think for the most part we've limited our use of void* to just those. Don't necessarily think that needs changing.

> Perhaps the vk.xml parsing script should throw error/exception when there is a void* without len && without noautovalidity.

Also I think perhaps a void* without a length is probably something we should catch. Could you raise an issue about it? (or a PR if you're up to it :P)

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: BTW, I corrected this one. I kept the original intention except that I felt the need to replace "consumed" (which usually means the exact opposite of what it was used to describe here) with "borrow" (borrowed from Rust-like terminology).

I moved most of the threading intentions into the Threading chapter.
I moved to a more detailed description. I have added a strong const guarantee (though some layers like `unique_handles` could benefit from lesser guarantee).

doc/specs/vulkan/chapters/memory.txt
krOoze: Well, the `len` on `void*` here implies bytes. `void` is not byte, neither really is `char` semantically (of course that's how people still coded it for decades anyway). It does not probably even have good alignment to be cast to anything else (so I know what data it is &mdash; bytes &mdash; unless otherwisely said by the spec).

> Could you raise an issue about it? (or a PR if you're up to it :P)

Sure. I could do it as part of this PR, or do you think it is getting bit overloaded?
Oh, and plese tell me if you would like something here unsquashed or perhaps as a separate PR.

src/spec/validitygenerator.py
krOoze: Well tell me how it goes with the board. There are way worse things than this being dropped.

doc/specs/vulkan/chapters/memory.txt
TobiasHector: I'd suggest a separate PR, otherwise I think this is going to get a bit difficult to review (plus higher chance of conflicts with later revisions if this drags, etc.).

doc/specs/vulkan/chapters/fundamentals.txt
```diff
@@ -284,10 +284,10 @@ frees.
 It is an application's responsibility to track the lifetime of Vulkan
 objects, and not to destroy them while they are still in use.
 
-Application-owned memory is immediately consumed by any Vulkan command it is
-passed into.
+Application-owned memory is borrowed by any Vulkan command it is passed into
```
TobiasHector: I think "borrowed" makes more sense, but I am concerned we've not used this term before. Would feel happier if you could put it as a glossary entry as well?

There's some notion of this concept being needed more widely in the API, since lifetimes are a bit tangled at the moment. For instance, ideally I'd want to be able to tag parameters for each call saying whether they are borrowed during the call, and whether they are released or not, and for later calls, when stored parameters are released. Similar to the externsync stuff, I guess. Potentially that would make the whole "command pools are externsync during command recording" easier to express as well.

I like "borrowed" since currently we don't have a term for it.

TobiasHector: (I don't know if "borrowed" is really the opposite to "released", just winging it here :P)

jeffbolznv: I dislike "borrowed", it implies a temporary change in ownership which doesn't exist.

krOoze: @TobiasHector I guess it does, if by "released" you mean something like "return borrowed item". It was intended for pass by reference (pointer) parameters, but I suppose it would work for other things too (handles). It does not necessarily implies `externsync`, but whether the parameter is borrowed and ignored, passed as const, or as mutable is a strong indication.

@jeffbolznv How so? That's exactly what it is. You no longer own the object (i.e. you have no moral right to free it while someone else owns it). Definition of "ownership" is not needed, because "borrowed" would be defined explicitly anyway. I stole it from [Rust here](https://doc.rust-lang.org/1.8.0/book/references-and-borrowing.html) if anyone wonders.
Anyway, I dislike "consumed" (it means `std::move()` in my dictionary) and I fail to find other precedent for this term on the internet.

doc/specs/vulkan/chapters/fundamentals.txt
```diff
@@ -707,13 +726,21 @@ commands, which will silently ignore these values.
...
 ==== Valid Usage for Pointers
 
-Any parameter that is a pointer must: either be a valid pointer, or if
-_explicitly called out in the specification_, can: be `NULL`.
 A pointer is valid if it points at memory containing values of the number
 and type(s) expected by the command, and all fundamental types accessed
 through the pointer (e.g. as elements of an array or as members of a
 structure) satisfy the alignment requirements of the host processor.
 
+Valid Usage section of a command or of a relevant parameter type does state
```
oddhack: Just a few grammatical comments:

"The Valid Usage section of commands or structure types states if a command parameter or structure member requires a valid pointer."

doc/specs/vulkan/chapters/fundamentals.txt
```diff
@@ -707,13 +726,21 @@ commands, which will silently ignore these values.
...
+if a command requires a valid pointer, or not.
+
+[NOTE]
+.Note
+====
+To some commands under some conditions a pointer can: be passed that is not
```
oddhack: "Under some conditions, a pointer can: be specified to command parameters or structure members that is NULL, or is not valid as defined above. In such cases the pointer value is ignored."

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: Yeah, that's probably better. It has a bit german sentence structure still. I might try something more direct and short like:

> If a command, or a structure type requires a valid pointer, the Valid Usage section will say so.

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: That would be a lie. `NULL` value is likely not ignored.
Then again, the application should not really care as long as it is not dereferenced.
I guess I will try something new.
What I wanted to do is to inform that some command descriptions say a parameter is "ignored". Then I wanted to say that lack of "valid" in the VU means it does not have to be valid (which was implied by above specifying sentence, but not said explicitly/humanely)

doc/specs/vulkan/chapters/fundamentals.txt
TobiasHector: I guess we're talking about ownership here, and actually the spec does call out "acquire" and "release" as verbs for this. So you could say that calling the command "acquires ownership of the pointer" and then when the command returns "ownership is released". 

There's possibly some question here with regards externsync and const pointers about whether they're fully owned (i.e. might be written, don't even read them during that call) or if there's shared ownership (i.e. only read, app can read at the same time).

doc/specs/vulkan/chapters/fundamentals.txt
TobiasHector: +1 to @oddhack's suggestion. I think it's probably to the point enough and reads quite naturally.

doc/specs/vulkan/chapters/fundamentals.txt
TobiasHector: How about:

"Pointer parameters passed to commands can additionally be `NULL` when specified by Valid Usage. If no valid usage is specified for a given pointer in a given circumstance, it can have any value as the parameter will be ignored by the calling command."

krOoze: @TobiasHector  I should not overthink it; it is only a Note, but:
- it cannot be "additionally"(to valid values) `NULL` in all cases. Sometimes it is "conditionally" `NULL`. Sometimes "only" `NULL` (mostly `pNext` with no known extensions).
- sometimes a valid usage is specified for a pointer parameter, but it still can be invalid (e.g. above in the `vkMapMemory`. Only the lack of word "valid" is significant

TobiasHector: ...can we live without the note :P?

krOoze: @TobiasHector We shall see. I guess first I should export the important stuff to other PRs, because on this there probably will be endless controversy and endless word-smithery despite relatively low added value :P

doc/specs/vulkan/chapters/memory.txt
krOoze: FYI: Sugested as PR #562.

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: Rewrote this trying "acquire ownership" term.

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: Reworded to something similar.

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: @TobiasHector Alright. Trying to live without it. 💔

doc/specs/vulkan/chapters/fundamentals.txt
```diff
@@ -268,10 +268,12 @@ frees.
 It is an application's responsibility to track the lifetime of Vulkan
 objects, and not to destroy them while they are still in use.
 
-Application-owned memory is immediately consumed by any Vulkan command it is
-passed into.
-The application can: alter or free this memory as soon as the commands that
-consume it have returned.
+[[fundamentals-objectmodel-lifetime-acquire]]
+The ownership of an application-owned memory is immediately acquired by any
```
tobine: "an application-owned memory" seems like it should be just "application-owned memory". The "an" is extraneous.

doc/specs/vulkan/chapters/fundamentals.txt
```diff
@@ -600,6 +602,16 @@ library) perform memory barriers as a part of mutual exclusion, so mutexing
 Vulkan objects via these primitives will have the desired effect.
 ====
 
+Similarly the application must: avoid any potential data hazard of
+application-owned memory that has its
+<<fundamentals-objectmodel-lifetime-acquire,ownership temporarily acquired>>
+by a Vulkan command.
+While the ownership of an application-owned memory remains acquired by a
```
tobine: Here again, is "an application-owned memory" intended/correct? Why the "an"?

krOoze: Because I have no idea how English works anymore. :) Care to write a specification about that? :p 
It is singular noun so I just assume there should be "an" or "the" or "this" or "any" or some such.

TobiasHector: Memory is a mass noun, rather than a singular, so it doesn't get an article (a/an/the). If you said "memory object", that would be singular. and thus get an article.

doc/specs/vulkan/chapters/fundamentals.txt
```diff
@@ -754,14 +766,14 @@ commands, which will silently ignore these values.
...
 
-Any parameter that is a pointer must: either be a valid pointer, or if
-_explicitly called out in the specification_, can: be `NULL`.
+A Valid Usage section always explicitly states if a command parameter or a
+structure member is required to be a valid pointer.
+
 A pointer is valid if it points at memory containing values of the number
 and type(s) expected by the command, and all fundamental types accessed
 through the pointer (e.g. as elements of an array or as members of a
 structure) satisfy the alignment requirements of the host processor.
```
krOoze: I should probably say something about the `void*` here; perhaps:

> Note: Pointer to a `void` type is either a byte array (if it has explicit expected size), or plain integer (i.e. unlike other pointers it has default expected size of 0 and expected type "nothing").

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: I guess...

doc/specs/vulkan/chapters/fundamentals.txt
```diff
@@ -754,14 +766,14 @@ commands, which will silently ignore these values.
 [[fundamentals-validusage-pointers]]
 ==== Valid Usage for Pointers
 
-Any parameter that is a pointer must: either be a valid pointer, or if
-_explicitly called out in the specification_, can: be `NULL`.
+A Valid Usage section always explicitly states if a command parameter or a
+structure member is required to be a valid pointer.
```
TobiasHector: I actually missed the subtlety when I reviewed this the first time, but this kind of implies that if we "miss" putting a "must be a valid pointer" in the VU, then it doesn't have to be valid even if the specification text says otherwise, which doesn't make any sense. I don't think we should take this paragraph change.

doc/specs/vulkan/chapters/fundamentals.txt
TobiasHector: I'm not sure what that would add - especially given that vkUpdateDescriptorSetWithTemplateKHR already violates this - it's not "just" a byte array. Also saying "it's an int" doesn't really make sense in the way I think you mean it to. I don't think we should add such a note in any capacity. The byte array stuff is kind of implicit by the nature of it being specified as a C99 API as well.

krOoze: Right. TBH I am half-tempted to delete most of this chapter, because it is mostly now explicit anyway since the autogenerator was made.

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: That is how it actually works though.
If the specification text says otherwise, such specification text is or should be moved to a VU section anyway.

TobiasHector: I don't disagree that it *should* be moved to the VU, and that any time we see something that isn't in a VU it should definitely move TO a VU, but we have not committed to saying that "If a VU isn't there it's fine" because we don't have a good process for ensuring that is the case.

IMO this is a massive over-commitment from what is true right now, though I would certainly love to be able to put that statement in a future spec. It just suggests something that is not currently guaranteed to be accurate, as much as we want it to be.

TobiasHector: NB: I have a proposal sitting around that makes it more tenable to add something like this but it's a ways off completion at the moment I'm afraid.

doc/specs/vulkan/chapters/fundamentals.txt
TobiasHector: Yea I mean this section was added as a sort of "guideline" to how to read the VU statements, rather than as a normative "spec works this way" section - because we knew full well it didn't (and still doesn't) entirely adhere to this.

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: >IMO this is a massive over-commitment from what is true right now

I don't really see it.

Firstly in this case it seems the script is fully handling it.
And if the author removes it from the script, he is fully responsible to add relevant VU.

By this statement I am not really commiting to something new, more like fixing the situation that sometimes pointers are allowed to be invalid (bit of an oxymoron really). I could say, they can be invalid, if explicitly stated in the spec, but that would be a lie.

So my only option here is either to add the statement as above, or pretty much remove the chapter. And removing the chapter would have exactly the same semantic meaning as the added statement, so not much a choice.

doc/specs/vulkan/chapters/fundamentals.txt
krOoze: Yeah, no. Disregard.
It adds nothing. It is fully covered by VU sections as defined below.
"Optional void*" has no VU (or lacks word "valid"),
array of bytes are explicitly in VUs too,
and `void*` without optional, len or noautovalidity should not happen (it would be either `char` instead or complete nonsense).

doc/specs/vulkan/chapters/fundamentals.txt
TobiasHector: > So my only option here is either to add the statement as above, or pretty much remove the chapter. And removing the chapter would have exactly the same semantic meaning as the added statement, so not much of a choice.

That's kind of extreme, but I see your point.

I think the problem I have is literally the strength of the statement, it's like asserting there will never be bugs. Also you've sort of changed the tone of the paragraph so it doesn't really fit with the rest of them. I think a reword would make me less uneasy, how about:

> Any parameter that is a pointer must: be a valid pointer only if it is explicitly called out by a Valid Usage statement.

krOoze: > [...], it's like asserting there will never be bugs.

Heh, that's bit of a circular. If there was a bug, that would make the statement also a bug. When the bug is fixed, it would make the statement true again :p 

Well, I think it is a healthy assertion throughout the specification that it is bug free, even if not literally true due to a human error. I would be more worried if the spec said: "There may: be unknown amount of bugs in this specification. Enjoy!" 😆 

>I think a reword would make me less uneasy, how about:
>
>>Any parameter that is a pointer must: be a valid pointer only if it is explicitly called out by a Valid Usage statement.

I am fine with that, as I see absolutely no difference in it for my purposes.

krOoze:
Updated with suggestions + added "valid" to explicit VUs to match the autogenerated ones.

Issue #364, opened 2016-09-16
vkCmdDraw*AMD not affected by pipeline


krOoze:
`vkCmdBindPipeline()` description says:

> The pipeline bound to `VK_PIPELINE_BIND_POINT_GRAPHICS` controls the behavior of `vkCmdDraw`, `vkCmdDrawIndexed`, `vkCmdDrawIndirect`, and `vkCmdDrawIndexedIndirect`. No other commands are affected by the pipeline state.

I think `vkCmdDrawIndirectCountAMD()` and `vkCmdDrawIndexedIndirectCountAMD()` commands should be in the list.

krOoze:
This should be fixed, due in the 1.0.63 spec update.

PR #549, opened 2017-08-23
Fix vkCmdDraw*AMD text bugs

### LINKED ISSUE #360: vkCmdDrawIndexedIndirectCountAMD indirect version of ...
> `vkCmdDrawIndexedIndirectCountAMD` behaves similar to `vkCmdDrawIndirectCountAMD` except that the draw count is read by the device from a buffer during execution.

That is not the difference between those two commands.

Also "similar to".

Similar error as was for the `vkCmdDrawIndexedIndirect` in #315.

krOoze:
Fixes #360 and #364.

### DIFF
--- a/doc/specs/vulkan/chapters/pipelines.txt
+++ b/doc/specs/vulkan/chapters/pipelines.txt
@@ -1565,8 +1565,7 @@ bind point.
 The pipeline bound to ename:VK_PIPELINE_BIND_POINT_COMPUTE controls the
 behavior of flink:vkCmdDispatch and flink:vkCmdDispatchIndirect.
 The pipeline bound to ename:VK_PIPELINE_BIND_POINT_GRAPHICS controls the
-behavior of flink:vkCmdDraw, flink:vkCmdDrawIndexed,
-flink:vkCmdDrawIndirect, and flink:vkCmdDrawIndexedIndirect.
+behavior of all <<drawing, drawing commands>>.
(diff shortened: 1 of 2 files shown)


oddhack:
@drakos-amd please review and sign off, since it's your extension.

doc/specs/vulkan/chapters/pipelines.txt
```diff
@@ -1502,8 +1502,7 @@ bind point.
 The pipeline bound to ename:VK_PIPELINE_BIND_POINT_COMPUTE controls the
 behavior of flink:vkCmdDispatch and flink:vkCmdDispatchIndirect.
 The pipeline bound to ename:VK_PIPELINE_BIND_POINT_GRAPHICS controls the
-behavior of flink:vkCmdDraw, flink:vkCmdDrawIndexed,
-flink:vkCmdDrawIndirect, and flink:vkCmdDrawIndexedIndirect.
+behavior of all ftext:vkCmdDraw* commands.
```
TobiasHector: Minor: Might be better to do <<draw, draw commands>>? We use that elsewhere.

drakos-amd: Agree with Tobias' suggestion, otherwise this PR looks good to me.

krOoze: So be it!

drakos-amd:
This looks good to me.

Issue #555, opened 2017-08-29
vk.xml: VkShaderModuleCreateInfo's pCode member has some garbage in its "len" attribute


yzsolt:
It looks like this in the latest vk.xml (line 846):
`<member len="latexmath:[codeSize \over 4]">`
Should be simply codeSize without the LaTeX garbage.

krOoze:
The `codeSize` is in (single) bytes, but `pCode` is in `uint32_t` (4-bytes elements). So as it currently is defined it must be `codeSize/4`.
What exactly are you suggesting here?

yzsolt:
I didn't know it was intended, my bad.
Still, it seems weird to see something like that in the XML. The other len values are just simply the names of the corresponding members. I use this info to generate nice span-parameter setters for struct members (something like Vulkan-Hpp does with ArrayProxies for array command parameters). However, in its current form this is unnecessarily hard to parse for a Vulkan wrapper code generator.
I think the relation between the array's length and the pointed type's size could be documented in an additional attribute instead, and this purely display-related latexmath code could be then generated based on that attribute by the documentation generator.

krOoze:
> I didn't know it was intended, my bad.

Yes, I suspected that is the case. Wasn't sure though... One could argue `codeSize` could have been size in `uint32_t`s, or `pCode` could have been `uint8_t` array. But that cannot be changed before Vulkan 2.0.

> Still, it seems weird to see something like that in the XML.

If you do not like this, you are gonna love `VkPipelineMultisampleStateCreateInfo::pSampleMask` 😈 

If you are using `vk.xml` you shoud be reading [vk.xml schema spec](https://www.khronos.org/registry/vulkan/specs/1.0/registry.html). It is documented there `len` can be a latex expression. And documents other gotchas.

> I think the relation between the array's length and the pointed type's size could be documented in an additional attribute instead

How? `len` is already an/the additional attribute.

One thing that could be done is to move them to `noautovalidity`. I don't see the benefit though. You could equally just ignore `len` as soon as you see `latexmath:`.

krOoze:
PS: It is not that hard to parse.
You could just hardcode those two latexmath cases. New ones should not appear frequently, and it would be a 5 min fix for your parser if they do.

oddhack:
An alternate representation could be provided, but the alternative representation would *also* have to be parsed. It might be easier to parse

> ceil(rasterizationSamples/32)

[...]

PR #556, opened 2017-09-04
Fix remaining len latexmath confusion

krOoze:
- fix `latex:` used in validity for remaining comma separated lengths in `len`
- correct comments in `registry.rnc` also refering to `latex:`

fixes rest of the emergent issue in #555


krOoze:
> I don't think there are circumstances it has ever been relevant, though.

Correct.  Though I tested it on one of the string arrays by replacing `null-terminated` with some latexmath. Seemed to work.

TobiasHector:
Looks good to me. Ship it!

krOoze:
@oddhack Nice!
Consider using the "Rebase and merge" button instead to avoid confusing commit history log.

PR #559, opened 2017-09-11
Remove wrong autovalidity for some void*

krOoze:
Remove wrong autovalidity for some void* parameters where it never should have been:

- WSI types hidden behind void* already have explicit VU
- void* in ppData (void**) does not have to be valid (potentially
confusing with char*, should ban void* autovalidity later)
- Template descriptors need specific structure -> must not be subject to autovalidity

exported from #547 (was bit offtopic there and now allows better piecemeal reviewing)


TobiasHector:
It looks like this is subsumed completely by #562, so... all my comments on the overlap in that PR apply here too.

krOoze:
I am burning this one. All the discussions seems to be in #562. Because of change in direction there is no longer need for two PRs.

PR #560, opened 2017-09-11
Fix autovalidity for output Flags pointers

krOoze:
Previously it required to be valid combination + non-0
(for `vkGetDeviceGroupSurfacePresentModesKHX::pModes` and`vkGetDeviceGroupPeerMemoryFeaturesKHX::pPeerMemoryFeatures`)

exported from #547 (was bit offtopic there and now allows better piecemeal reviewing)

### DIFF
--- a/src/spec/validitygenerator.py
+++ b/src/spec/validitygenerator.py
@@ -438,16 +438,17 @@ class ValidityOutputGenerator(OutputGenerator):
         if asciidoc != '':
             asciidoc += '\n'
 
-            # Add additional line for non-optional bitmasks
-            if self.getTypeCategory(paramtype.text) == 'bitmask':
-                if param.attrib.get('optional') is None:
-                    asciidoc += self.makeAnchor(blockname, paramname.text, 'requiredbitmask')
-                    if self.paramIsArray(param):
-                        asciidoc += 'Each element of '
-                    asciidoc += 'pname:'
-                    asciidoc += paramname.text
-                    asciidoc += ' must: not be `0`'
-                    asciidoc += '\n'
+        # Add additional line for non-optional bitmasks
+        if self.getTypeCategory(paramtype.text) == 'bitmask':
+            isMandatory = param.attrib.get('optional') is None #TODO does not really handle if someone tries something like optional="true,false"
+            if isMandatory:
+                asciidoc += self.makeAnchor(blockname, paramname.text, 'requiredbitmask')
+                if self.paramIsArray(param):
+                    asciidoc += 'Each element of '
+                asciidoc += 'pname:'
+                asciidoc += paramname.text
+                asciidoc += ' must: not be `0`'
+                asciidoc += '\n'
 
         return asciidoc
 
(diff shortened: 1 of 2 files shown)


src/spec/validitygenerator.py
```diff
@@ -508,7 +513,15 @@ def createValidationLineForParameter(self, blockname, param, params, typecategor
                 asciidoc += '\n'
             else:
                 if self.paramIsArray(param):
-                    asciidoc += self.makeAsciiDocLineForParameter(blockname, param, params, 'combinations of ' + self.makeEnumerationName(bitsname) + ' value')
+                    if param.text is not None and 'const' in param.text:
+                        asciidoc += self.makeAsciiDocLineForParameter(blockname, param, params, 'combinations of ' + self.makeEnumerationName(bitsname) + ' value')
```
tobine: Should this one be "values" (plural) instead of "value"?

krOoze: I think there is a hack somewhere in the script adding "s" to pluralize things.

TobiasHector:
Looks good to me - @oddhack please feel free to merge this one. WG agreed to just let me use my best judgement on this one - and it looks good to me.

+1

src/spec/validitygenerator.py
TobiasHector: Indeed there is. It's... a fun script.

TobiasHector:
Oh except I see @tobine has reviewed this - do you have any concerns?

src/spec/validitygenerator.py
```diff
@@ -438,12 +438,17 @@ def createValidationLineForParameterIntroChunk(self, blockname, param, params, t
...
+        if self.getTypeCategory(paramtype.text) == 'bitmask':
+            if param.attrib.get('optional') is None:
+                isConstArrayOrPointer = (self.paramIsArray(param) or self.paramIsPointer(param)) and param.text is not None and 'const' in param.text;
+                isValue = not self.paramIsArray(param) and not self.paramIsPointer(param)
+                if isValue or isConstArrayOrPointer:
                     asciidoc += self.makeAnchor(blockname, paramname.text, 'requiredbitmask')
                     if self.paramIsArray(param):
                         asciidoc += 'Each element of '
+                    elif self.paramIsPointer(param):
+                        asciidoc += 'The value pointed by '
```
TobiasHector: Shoot I missed this since it's not actually showing up in the spec yet. This should be changed to "The value pointed at by" before it gets merged. Or actually just remove it - there's no obvious reason we'd ever have a const pointer to a bitmask in the spec, it makes zero sense.

krOoze: Actually should probably be marked `optional` in the xml instead of doing this guesswork here in the script based on const.

krOoze: Updated with that in mind. Those two are now `optional="false,true"`.

TobiasHector:
I'm fine with this now, as long as @tobine doesn't have further comment, this can be merged.

PR #562, opened 2017-09-12
Tweak some void* (auto)validity

krOoze:
- Make implicit validity for optional `void*` mean "nothing" in the
script
- Make `optional` mean non-NULL for `void*` in the registry (unimplemented in script though; there is not much use for non-optional void*)
- Uses of `void*` that point to something should preferrably have `noautovalidity`
(adding it)
- Add missing explicit VUs of `void*` parameters
(`vkCmdPushDescriptorSetWithTemplateKHR` and
`vkUpdateDescriptorSetWithTemplateKHR`)
- Also indirectly fixes `vkMapMemory`, `VkAllocationCallbacks`,
`VkDebugReportCallbackCreateInfoEXT`, and Apple and NN surface creation
- Remove ugly text hack from script

### DIFF
--- a/doc/specs/vulkan/chapters/descriptorsets.txt
+++ b/doc/specs/vulkan/chapters/descriptorsets.txt
@@ -2373,6 +2373,15 @@ include::../api/protos/vkUpdateDescriptorSetWithTemplateKHR.txt[]
     of slink:VkDescriptorImageInfo, slink:VkDescriptorBufferInfo, or
     slink:VkBufferView used to write the descriptors.
 
+.Valid Usage
+****
+  * pname:pData must: be a pointer to a memory that contains one or more
+    valid instances of slink:VkDescriptorImageInfo,
+    slink:VkDescriptorBufferInfo, or slink:VkBufferView in a layout defined
+    by pname:descriptorUpdateTemplate when it was created with
+    flink:vkCreateDescriptorUpdateTemplateKHR
+****
+
 include::../validity/protos/vkUpdateDescriptorSetWithTemplateKHR.txt[]
 
 .API example
@@ -2699,6 +2708,11 @@ include::../api/protos/vkCmdPushDescriptorSetWithTemplateKHR.txt[]
     The pipelineBindPoint specified during the creation of the descriptor
     update template must: be supported by the pname:commandBuffer's parent
     sname:VkCommandPool's queue family
+  * pname:pData must: be a pointer to a memory that contains one or more
+    valid instances of slink:VkDescriptorImageInfo,
+    slink:VkDescriptorBufferInfo, or slink:VkBufferView in a layout defined
+    by pname:descriptorUpdateTemplate when it was created with
+    flink:vkCreateDescriptorUpdateTemplateKHR
 ****
 
 include::../validity/protos/vkCmdPushDescriptorSetWithTemplateKHR.txt[]
--- a/src/spec/validitygenerator.py
+++ b/src/spec/validitygenerator.py
@@ -406,22 +406,28 @@ class ValidityOutputGenerator(OutputGenerator):
 
         elif self.paramIsPointer(param):
             # Handle pointers - which are really special case arrays (i.e. they don't have a length)
+            #TODO  should do something here if someone ever uses some intricate comma-separated `optional`
             pointercount = paramtype.tail.count('*')
 
-            # Could be multi-level pointers (e.g. ppData - pointer to a pointer). Handle that.
-            for i in range(0, pointercount):
-                asciidoc += 'a pointer to '
-
+            # Treat void* as an int
             if paramtype.text == 'void':
-                # If there's only one pointer, it's optional, and it doesn't point at anything in particular - we don't need any language.
-                if pointercount == 1 and param.attrib.get('optional') is not None:
+                pointercount -= 1
+
+            # Could be multi-level pointers (e.g. ppData - pointer to a pointer). Handle that.
+            asciidoc += 'a '
+            for i in range(0, pointercount):
+                asciidoc += 'pointer to a '
+
+            # Handle void* and pointers to it
+            if paramtype.text == 'void':
+                # If there is only void*, it is just optional int - we don't need any language.
+                if pointercount == 0 and param.attrib.get('optional') is not None:
                     return '' # early return
                 else:
-                    # Pointer to nothing in particular - delete the " to " portion
-                    asciidoc = asciidoc[:-4]
-            else:
-                # Add an article for English semantic win
-                asciidoc += 'a '
+                    if param.attrib.get('optional').split(',')[pointercount] is not None:
+                        # The last void* is just optional int (e.g. to be filled by the impl.)
+                        typetext = 'pointer value'
+
 
             # If a value is "const" that means it won't get modified, so it must be valid going into the function.
             if param.text is not None and paramtype.text != 'void':
--- a/src/spec/vk.xml
+++ b/src/spec/vk.xml
@@ -1527,7 +1527,7 @@ private version is maintained in the 1.0 branch of the member gitlab server.
             <member values="VK_STRUCTURE_TYPE_VI_SURFACE_CREATE_INFO_NN"><type>VkStructureType</type> <name>sType</name></member>
             <member>const <type>void</type>*                      <name>pNext</name></member>
             <member optional="true"><type>VkViSurfaceCreateFlagsNN</type>   <name>flags</name></member>
-            <member><type>void</type>*                            <name>window</name></member>
+            <member noautovalidity="true"><type>void</type>*                            <name>window</name></member>
         </type>
         <type category="struct" name="VkWaylandSurfaceCreateInfoKHR">
             <member values="VK_STRUCTURE_TYPE_WAYLAND_SURFACE_CREATE_INFO_KHR"><type>VkStructureType</type> <name>sType</name></member>
@@ -2317,13 +2317,13 @@ private version is maintained in the 1.0 branch of the member gitlab server.
             <member values="VK_STRUCTURE_TYPE_IOS_SURFACE_CREATE_INFO_MVK"><type>VkStructureType</type> <name>sType</name></member>
             <member>const <type>void</type>*                                    <name>pNext</name></member>
             <member optional="true"><type>VkIOSSurfaceCreateFlagsMVK</type>     <name>flags</name></member>
-            <member>const <type>void</type>*                                    <name>pView</name></member>
+            <member noautovalidity="true">const <type>void</type>*                                    <name>pView</name></member>
         </type>
         <type category="struct" name="VkMacOSSurfaceCreateInfoMVK">
             <member values="VK_STRUCTURE_TYPE_MACOS_SURFACE_CREATE_INFO_MVK"><type>VkStructureType</type> <name>sType</name></member>
             <member>const <type>void</type>*                                    <name>pNext</name></member>
             <member optional="true"><type>VkMacOSSurfaceCreateFlagsMVK</type>   <name>flags</name></member>
-            <member>const <type>void</type>*                                    <name>pView</name></member>
+            <member noautovalidity="true">const <type>void</type>*                                    <name>pView</name></member>
         </type>
         <type category="struct" name="VkViewportWScalingNV">
             <member><type>float</type>          <name>xcoeff</name></member>
@@ -3796,7 +3796,7 @@ private version is maintained in the 1.0 branch of the member gitlab server.
             <param><type>VkDeviceSize</type> <name>offset</name></param>
             <param><type>VkDeviceSize</type> <name>size</name></param>
             <param optional="true"><type>VkMemoryMapFlags</type> <name>flags</name></param>
-            <param><type>void</type>** <name>ppData</name></param>
+            <param optional="false,true"><type>void</type>** <name>ppData</name></param>
         </command>
         <command>
             <proto><type>void</type> <name>vkUnmapMemory</name></proto>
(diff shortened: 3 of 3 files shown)


doc/specs/vulkan/chapters/descriptorsets.txt
```diff
@@ -2373,6 +2373,12 @@ include::../api/protos/vkUpdateDescriptorSetWithTemplateKHR.txt[]
     of slink:VkDescriptorImageInfo, slink:VkDescriptorBufferInfo, or
     slink:VkBufferView used to write the descriptors.
 
+.Valid Usage
+****
+  * pname:pData must: be a pointer to a raw memory structure as defined by
+    pname:descriptorUpdateTemplate when it was created
```
TobiasHector: Wouldn't it make more sense to say "must: be a pointer to an array of instances of VkDescriptorImageInfo,
VkDescriptorBufferInfo, and VkBufferView structure."?

Ideally we should be more precise and have a number of statements indicating at which specific structures go where but... we should probably save that for a future PR :)

doc/specs/vulkan/chapters/descriptorsets.txt
```diff
@@ -2699,6 +2705,8 @@ include::../api/protos/vkCmdPushDescriptorSetWithTemplateKHR.txt[]
     The pipelineBindPoint specified during the creation of the descriptor
     update template must: be supported by the pname:commandBuffer's parent
     sname:VkCommandPool's queue family
+  * pname:pData must: be a pointer to a raw memory structure as defined by
+    pname:descriptorUpdateTemplate when it was created
```
TobiasHector: Same comment as above.

src/spec/validitygenerator.py
```diff
@@ -413,7 +413,15 @@ def createValidationLineForParameterIntroChunk(self, blockname, param, params, t
...
 
             if paramtype.text == 'void':
-                # If there's only one pointer, it's optional, and it doesn't point at anything in particular - we don't need any language.
+                self.logMsg('error',
+                            ' Parameter ' + blockname + '::' + paramname.text + ' is a ' + paramtype.text + paramtype.tail.strip() + '.'
+                            ' void* without noautovalidity and without len attribute is a source of errors.'
+                            ' It has been prohibited.'
+                            ' Likely you forgot len or forgot noautovalidity attribute in vk.xml.'
+                            ' In the rare case this is true pointer to nothing or opaque memory,'
+                            ' you are now required to add noautovalidity too.')
```
TobiasHector: Do we really need the error? For other cases where it's ambiguous the autovalidity script simply doesn't attempt to do anything with it. If we ever add a function or struct, people will look at the built spec, see what's missing, and add it - works for other parameter types, think it's probably fine for this? I'd probably rather just not handle this type of parameter and silently do nothing. If we add an error for this then there might be an expectation for other such error checks which simply do not exist right now.

TobiasHector:
@krOoze does this include all the commits in #559, or should I review that separately? I couldn't tell..

krOoze:
@TobiasHector Yeah, this is an optimistic PR completely based on the #559 branch (and always synced up unless I mess up). First commit is the #559, second commit is the subject of this PR.

doc/specs/vulkan/chapters/descriptorsets.txt
krOoze: I haven't study the extension in that much detail.
Still my version is sort of complete, but too terse.
Your version does reference the needed types, but does say pretty much nothing about the validity. What does "being a pointer to array of three types" even mean? How would you test it?

src/spec/validitygenerator.py
krOoze: That's too psychologically and sociologically deep for me...
I guess I could make it into inline comment and in the error message say something cryptic like "Parsing error - invalid registry" to preserve the image of an unreliable script.

> people will look at the built spec, see what's missing

I think you have too much trust in people though. Half of the formating bugs wouldn't be there in the first place if people did so. E.g. care to explain how the broken blocks below `oldSwapchain` happened if people built the spec and evaluated what is missing? No offence meant to anybody...

TobiasHector: > I guess I could make it into inline comment and in the error message say something cryptic like "Parsing error - invalid registry" to preserve the image of an unreliable script.

I don't think that's a good idea. If you're going to throw an error, say something useful. I just don't see the value in this particular error - you're basically just forcing users to awkwardly acknowledge that there's no VU for void* parameters, by throwing an error and telling them to put "noautovalidity" on it.

If you drop the "noautovalidity" requirement and drop any sort of handling for void* parameters, I don't think anything really changes. The main benefit of this change IMO is not dumping out misleading/incorrect VU statements - which is what it currently does.

> I think you have too much trust in people though. Half of the formating bugs wouldn't be there in the first place if people did so. E.g. care to explain how the broken blocks below oldSwapchain happened if people built the spec and evaluated what is missing? No offense meant to anybody...

Oh no they definitely screw it up - it's just that in this case the worst possible outcome of not throwing an error is "oh I missed a VU", which is trivially fixable later. We definitely need more sanity checks for other things though, but that's a whole other issue..

doc/specs/vulkan/chapters/descriptorsets.txt
TobiasHector: > What does "being a pointer to array (of unspecified size) of three types" even mean? How would you test it?

The validation layers would be able to cast to an untyped Vulkan struct with "sType" in it, and thus could determine the size and type of the struct, then they have enough information to iterate to the next one and repeat until they run out. We could have a vague statement of "n structs according to the number of structs specified for the descriptor template" which is imprecise, but enough to go to the spec to figure out how many structs there should be in total.

So specifically that would be:

>  * pname:pData must: be a pointer to an array of [eq]#n# instances of VkDescriptorImageInfo, VkDescriptorBufferInfo, and VkBufferView structure, where [eq]#n# is the number of descriptors described by pname:descriptorUpdateTemplate.

It's not super precise, but it's enough to do some basic validation. We can iterate on more precise VU language after this PR is in...

krOoze: > The validation layers would be able to cast to an untyped Vulkan struct with "sType" in it, and thus could determine the size and type of the struct

`VkBufferView` is not a structure.

src/spec/validitygenerator.py
krOoze: > If you drop the "noautovalidity" requirement and drop any sort of handling for void* parameters, I don't think anything really changes.

Yeah, well, it is the raison d'etre of this PR...

Alternativelly implicit validity for `void*` could be redefined to, well, have no implicit validity. That would result in no VU statements generated and no `noautovalidity` needed (even in #559).
I simply liked the `noautovalidity` version better, because it implies there must be explicit VU. And there must be explicit VU. And unlike any other type, `void*` only on implicit validity is almost sure to be an error (except for pass-through pointers like `pUserData`).

> error is "oh I missed a VU", which is trivially fixable later.

Fixable and fixed are two different things.
PR #559 is evidence that those kinds of human errors do happen, and also that their authors never fix them before release(, or at least in timely manner anyway). (And it fixes those errors.)
This PR strives to make sure that the same error never happens in the future.

doc/specs/vulkan/chapters/descriptorsets.txt
TobiasHector: Hm good point - that would change the wording a bit. But given that it's just VkBufferView that isn't a type, they'd see that it's not one of the sTypes, so they could check that the 64-bit value is a valid VkBufferView that's been created by the app. Not a big deal.

As for what "n" is, someone will have to look at the spec and figure it out from the definition of the descriptor templates. This would require extra work to actually define in the VU properly. Best we can do for now is make a link to the vkCreateDescriptorUpdateTemplateKHR function.

So updated wording:

>  * pname:pData must: be a pointer to an array of [eq]#n# instances of slink:VkDescriptorImageInfo, slink:VkDescriptorBufferInfo structures or slink:VkBufferView handles, where [eq]#n# is the number of descriptors described by pname:descriptorUpdateTemplate as defined by flink:vkCreateDescriptorUpdateTemplateKHR .

src/spec/validitygenerator.py
TobiasHector: > Alternativelly implicit validity for void* could be redefined to, well, have no meaningful implicit validity. That would result in no VU statements generated and no noautovalidity needed (even in #559 cases). That what you proposing?

Yes, precisely that.

> Fixable and fixed are two different things.

Sure, but we don't add void* parameters to the spec very often anyway (it's largely discouraged). Adding additional VU constraints is more about clarifying the situation for the validation layers rather than changing any behaviour - it's always been invalid, but the VU will make it obvious that's the case. If we were at 100% coverage for existing VUs I'd be more worried, but as it stands there's probably a few hundred VU statements missing from the spec in various edge cases - this won't really help with that wider problem.

doc/specs/vulkan/chapters/descriptorsets.txt
krOoze: @TobiasHector Well, we are slowly iterating it now anyway. Might as well figure out what the best wording is now. We have a saying in my country about a lazyman that died from overwork.

TobiasHector: Hmm, if we want to do that then (regrettably) I'd say we should split that work off - I'd rather get the misleading VU statements out sooner rather than waiting on us to iterate precise wording - plus it'll require additional feedback from the group. I currently have group mandate to sign this PR off on my own, but if we get to that level of detail I'll need to go back to the working group for other people to review.

So by all means, yea, let's do that, but in another PR. For this PR then we should remove the explicit VU for these parameters, if that's the route you want to go.

src/spec/validitygenerator.py
krOoze: > Yes, precisely that.

Very well then, I will prepare such variant.

> If we were at 100% coverage for existing VUs I'd be more worried, but as it stands there's probably a few hundred VU statements missing from the spec in various edge cases - this won't really help with that wider problem.

Ew, hundreds? Oh my!

doc/specs/vulkan/chapters/descriptorsets.txt
krOoze: @TobiasHector Nah, seems solid to me. I was just being sarcastic at the energy expended to figure out a quick placeholder, until your pretty much came with a version that looks more final than both initially proposed versions.

TobiasHector: Haha fair enough - sorry the subtlty was lost on me :D

src/spec/validitygenerator.py
TobiasHector: Yea I could probably pick a random function, stare at it for an hour and find a chunk of things that need to be there. I just... don't want to do that :D

krOoze: I wish I had your discipline. You are describing precisely what tends to happen to me. :D 

Hmm, if I get rid of `noautovalidity`, should `optional` still have (non-NULL) meaning? In that case the VU section would be spammed with redundant non-NULL statements. Which would force authors to use `noautovalidity` there anyway.
Damit, I am doing it right now. 😱 Ok I am gonna need to stare at it for an hour 😅

doc/specs/vulkan/chapters/descriptorsets.txt
krOoze: Actually, reading the extension there are also `offsets` and `strides` allowing for holes and whatnot.
I would drop the "n" and "array", which is not that meaningful concept in that context. So proposing another placeholder (until authors figure out what exactly they mean here):

> `pData` is a pointer to a memory which contains one or more valid `VkDescriptorImageInfo`, `VkDescriptorBufferInfo`, or `VkBufferView` in a layout as definded by `descriptorUpdateTemplate` when it was created with [vkCreateDescriptorUpdateTemplateKHR]().

TobiasHector: Perfect - let's go with that.

src/spec/validitygenerator.py
TobiasHector: > I wish I had your discipline. You are describing precisely what tends to happen to me. :D

Haha yea I can't say I hadn't noticed ;)

> Hmm, if I get rid of noautovalidity, should optional still have (non-NULL) meaning? In that case the VU section would be spammed with redundant non-NULL statements. Which would force authors to use noautovalidity there anyway.
Damit, I am doing it right now. 😱 Ok I am gonna need to stare at it for an hour 😅

Ewww... Ugh damnit yea now I'm going to have to go stare at them too :P!

krOoze: @TobiasHector I am gonna go with true implicit validity for `pUserData`, and `noautovalidity` elsewhere. If future author does not add `noautovalidity`, there would be no harm except VU saying "non-NULL" redundant to his own explicit VU statement.

krOoze: @TobiasHector Yeah, I am gonna go with true implicit for `pUserData`, indirect optional for `void**` in mapMemory and `noautovalidity` elsewhere. There's not much harm if future author screws it up -- it would only result in benign non-NULL implicit VU redundant to his own explicit VU.

krOoze:
Ok, the changes are up.

TobiasHector:
@oddhack please feel free to merge this one as you see fit.

src/spec/validitygenerator.py
```diff
@@ -406,22 +406,27 @@ def createValidationLineForParameterIntroChunk(self, blockname, param, params, t
...
             pointercount = paramtype.tail.count('*')
 
+            # Treat void* as an int
+            if paramtype.text == 'void':
+                pointercount -= 1
+
             # Could be multi-level pointers (e.g. ppData - pointer to a pointer). Handle that.
             for i in range(0, pointercount):
-                asciidoc += 'a pointer to '
+                asciidoc += 'a pointer to a '
```
krOoze: Forgot to move the first "a " outside the loop. Wil fix momentarily.
Would break only non-existent `***`+ pointers; not a big deal.

PR #585, opened 2017-10-13
Update VkSwapchainCreateInfoKHR descriptions

krOoze:
- fix `pretransform` and `compositeAlpha`; they are values not bitmasks
- update `imageFormat`, `imageColorSpace`, `imageUsage`  and `oldSwapchain` to be specific to behavior of this command
- fix broken unintended code block
- move `VK_TRUE` clipping text to correct bullet point
- rewrite Note to (hopefully) help clarity and brevity
- random tweaks...

### DIFF
--- a/doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
+++ b/doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
@@ -110,106 +110,110 @@ include::../../api/structs/VkSwapchainCreateInfoKHR.txt[]
   * pname:sType is the type of this structure.
   * pname:pNext is `NULL` or a pointer to an extension-specific structure.
   * pname:flags is a bitmask of elink:VkSwapchainCreateFlagBitsKHR
-    indicating parameters of swapchain creation.
-  * pname:surface is the surface to which the swapchain will present images.
-    The swapchain is associated with pname:surface.
+    indicating parameters of the swapchain creation.
+  * pname:surface is the surface onto which the swapchain will present
+    images.
+    If the creation succeeds, the swapchain becomes associated with
+    pname:surface.
   * pname:minImageCount is the minimum number of presentable images that the
     application needs.
-    The platform will either create the swapchain with at least that many
-    images, or will fail to create the swapchain.
-  * pname:imageFormat is a slink:VkFormat that is valid for swapchains on
-    the specified surface.
-  * pname:imageColorSpace is a slink:VkColorSpaceKHR that is valid for
-    swapchains on the specified surface.
-  * pname:imageExtent is the non-zero size (in pixels) of the swapchain.
-    Behavior is platform-dependent when the image extent does not match the
-    surface's pname:currentExtent as returned by
+    The implementation will either create the swapchain with at least that
+    many images, or it will fail to create the swapchain.
+  * pname:imageFormat is a elink:VkFormat value specifying the format
+    the swapchain image(s) will be created with.
+  * pname:imageColorSpace is a slink:VkColorSpaceKHR value specifying the
+    way the swapchain interprets image data.
+  * pname:imageExtent is the size (in pixels) of the swapchain image(s).
+    The behavior is platform-dependent if the image extent does not match
+    the surface's pname:currentExtent as returned by
     fname:vkGetPhysicalDeviceSurfaceCapabilitiesKHR.
 
 [NOTE]
 .Note
 ====
-On some platforms, flink:vkGetPhysicalDeviceSurfaceCapabilitiesKHR can
-return in slink:VkSurfaceCapabilitiesKHR a pname:currentExtent that has a
-pname:width and/or pname:height equal to zero.
-This can happen, e.g., while a native window object is minimized.
-In this situation, the native window object may have to be restored from the
-minimized state before a sname:VkSwapchainKHR can be created, because:
+During some situations, sname:VkSwapchainKHR cannot: be created because of
+the Valid Usage of the pname:imageExtent parameter.
 
-  * The pname:minImageExtent and pname:maxImageExtent values returned by
-    flink:vkGetPhysicalDeviceSurfaceCapabilitiesKHR may match the
-    pname:currentExtent.
-  * On some platforms, the pname:imageExtent member of
-    sname:VkSwapchainCreateInfoKHR must match the pname:currentExtent
-    returned by flink:vkGetPhysicalDeviceSurfaceCapabilitiesKHR.
+The flink:vkGetPhysicalDeviceSurfaceCapabilitiesKHR may: return in
+slink:VkSurfaceCapabilitiesKHR a pname:currentExtent that has the
+pname:width and/or pname:height equal to 0, or that is less than its
+counterpart in pname:minImageExtent.
+Similarly, pname:maxImageExtent may: also have 0 values for pname:width
+and/or pname:height.
+For example, on some platforms that may: happen while an associated window
+is minimized.
 ====
 
   * pname:imageArrayLayers is the number of views in a multiview/stereo
     surface.
     For non-stereoscopic-3D applications, this value is 1.
-  * pname:imageUsage is a bitmask of elink:VkImageUsageFlagBits, indicating
-    how the application will use the swapchain's presentable images.
-  * pname:imageSharingMode is the sharing mode used for the images of the
+  * pname:imageUsage is a bitmask of elink:VkImageUsageFlagBits describing
+    the intended usage of the (acquired) swapchain images.
+  * pname:imageSharingMode is the sharing mode used for the image(s) of the
     swapchain.
   * pname:queueFamilyIndexCount is the number of queue families having
-    access to the images of the swapchain in case pname:imageSharingMode is
+    access to the image(s) of the swapchain when pname:imageSharingMode is
     ename:VK_SHARING_MODE_CONCURRENT.
   * pname:pQueueFamilyIndices is an array of queue family indices having
-    access to the images of the swapchain in case pname:imageSharingMode is
+    access to the images(s) of the swapchain when pname:imageSharingMode is
     ename:VK_SHARING_MODE_CONCURRENT.
-  * pname:preTransform is a bitmask of elink:VkSurfaceTransformFlagBitsKHR,
+  * pname:preTransform is a elink:VkSurfaceTransformFlagBitsKHR value
     describing the transform, relative to the presentation engine's natural
     orientation, applied to the image content prior to presentation.
     If it does not match the pname:currentTransform value returned by
     fname:vkGetPhysicalDeviceSurfaceCapabilitiesKHR, the presentation engine
     will transform the image content as part of the presentation operation.
-  * pname:compositeAlpha is a bitmask of elink:VkCompositeAlphaFlagBitsKHR
+  * pname:compositeAlpha is a elink:VkCompositeAlphaFlagBitsKHR value
     indicating the alpha compositing mode to use when this surface is
     composited together with other surfaces on certain window systems.
   * pname:presentMode is the presentation mode the swapchain will use.
     A swapchain's present mode determines how incoming present requests will
     be processed and queued internally.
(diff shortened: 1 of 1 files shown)


doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -115,13 +115,14 @@ include::../../api/structs/VkSwapchainCreateInfoKHR.txt[]
...
-    the specified surface.
-  * pname:imageColorSpace is a slink:VkColorSpaceKHR that is valid for
-    swapchains on the specified surface.
-  * pname:imageExtent is the non-zero size (in pixels) of the swapchain.
+    The implementation will either create the swapchain with at least that
+    many images, or it will fail to create the swapchain.
+  * pname:imageFormat is a slink:VkFormat that the swapchain image(s) will
+    be created with.
+  * pname:imageColorSpace is a slink:VkColorSpaceKHR that identifies how
+    will the swapchain interpret the swapchain image data.
```
TobiasHector: nit: This reads weirdly, suggest instead:

> pname:imageColorSpace is the elink:VkColorSpaceKHR which the presentation engine will interpret the color values as.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -115,13 +115,14 @@ include::../../api/structs/VkSwapchainCreateInfoKHR.txt[]
...
-    The platform will either create the swapchain with at least that many
-    images, or will fail to create the swapchain.
-  * pname:imageFormat is a slink:VkFormat that is valid for swapchains on
-    the specified surface.
-  * pname:imageColorSpace is a slink:VkColorSpaceKHR that is valid for
-    swapchains on the specified surface.
-  * pname:imageExtent is the non-zero size (in pixels) of the swapchain.
+    The implementation will either create the swapchain with at least that
+    many images, or it will fail to create the swapchain.
+  * pname:imageFormat is a slink:VkFormat that the swapchain image(s) will
```
TobiasHector: should be "elink:" (existing problem, but since you're here...)

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -115,13 +115,14 @@ include::../../api/structs/VkSwapchainCreateInfoKHR.txt[]
...
+    be created with.
+  * pname:imageColorSpace is a slink:VkColorSpaceKHR that identifies how
+    will the swapchain interpret the swapchain image data.
+  * pname:imageExtent is the non-zero size (in pixels) of the swapchain
```
TobiasHector: nit: Existing problem, but given that the non-zero-ness is in the valid usage, this part should be removed. It should just say:

> pname:imageExtent is the size (in pixels) of the swapchain image(s).

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -147,23 +148,23 @@ minimized state before a sname:VkSwapchainKHR can be created, because:
...
-    how the application will use the swapchain's presentable images.
-  * pname:imageSharingMode is the sharing mode used for the images of the
+  * pname:imageUsage is a bitmask of elink:VkImageUsageFlagBits describing
+    the intended usage of the (acquired) swapchain images.
+  * pname:imageSharingMode is the sharing mode used for the image(s) of the
     swapchain.
   * pname:queueFamilyIndexCount is the number of queue families having
-    access to the images of the swapchain in case pname:imageSharingMode is
-    ename:VK_SHARING_MODE_CONCURRENT.
+    access to the image(s) of the swapchain in case pname:imageSharingMode
```
TobiasHector: nit: existing but probably should be simply "when" instead of "in case". Just weird phrasing...

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
ianelliottus: Given that "non-zero" was just added with this week's release, I'd rather not remove it in this pull request.  Probably best to discuss internally.

TobiasHector: Um, from what I could see after I wrote this there's an internal change in flight to *remove* it... happy to double check though.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -175,41 +176,44 @@ minimized state before a sname:VkSwapchainKHR can be created, because:
...
-after reacquiring them and if their pixel shaders do not have any side
+after reacquiring them, and if their pixel shaders do not have any side
 effects that require them to run for all pixels in the presentable image.
 ====
 
   * pname:oldSwapchain, if not dlink:VK_NULL_HANDLE, specifies an existing
-    non-retired swapchain that is associated with surface.
+    non-retired swapchain that is associated with pname:surface to indicate
+    that the application wishes to still allow use of pname:oldSwapchain's
+    images that are currently acquired.
```
ianelliottus: This is not the full and complete reason.  The application can: use it that way.  The original intention was so that the PE can reuse the memory of pname:oldSwapchain in creating the new swapchain.

ianelliottus:
I mostly like this pull request.  I agree with all but one of Tobias's proposed changes.  I dislike one of the proposed additions to one sentence.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -115,13 +115,14 @@ include::../../api/structs/VkSwapchainCreateInfoKHR.txt[]
...
-  * pname:imageFormat is a slink:VkFormat that is valid for swapchains on
-    the specified surface.
-  * pname:imageColorSpace is a slink:VkColorSpaceKHR that is valid for
-    swapchains on the specified surface.
-  * pname:imageExtent is the non-zero size (in pixels) of the swapchain.
+    The implementation will either create the swapchain with at least that
+    many images, or it will fail to create the swapchain.
+  * pname:imageFormat is a slink:VkFormat that the swapchain image(s) will
+    be created with.
+  * pname:imageColorSpace is a slink:VkColorSpaceKHR that identifies how
```
oddhack: Typo "how will"

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: It is tad excessive though. It is mentioned here, in the VU, and in the Note also.
Then again, it is just one small word. What's the harm in keeping it?

TobiasHector: If this is contentious, happy to not do it in this PR.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: @ianelliottus True. I actually suspected that might be the case and use it that way.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: Probably will change whole for better estheticity as @TobiasHector suggests below.

But out of curiosity, what is wrong with "how will"?

oddhack: ... identifies *how* (the swapchain) *will* interpret. I don't know the right grammar term for this but the current usage isn't good.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
ianelliottus: I don't care personally, but wanted the author of those words to have a chance to weigh in.  I just sent him/Tobias a private email, pointing him here.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
ianelliottus: I don't have suggested wording changes.  I'd either remove the additional/new text, or let's figure out what to change it to.  Do you have a preference?

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
ianelliottus: I agree with @oddhack.  I like changing it to "... identifies how the swapchain will ..."

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: It is always good to have a spare Tobias...

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: Seems like a very useful information. I would prefer to add it. If not here, then below in the freeform descriptions.

krOoze: My beef with the current parameter description is that it describes validity, not semantics/behavior.

ianelliottus: Sounds good.  I look forward to your new wording.  Thanks!

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
jjulianoatnv: I'm author of those words. It looks like the non-zeroness is replicated in the three places pointed out: VU, NOTE, and parameter definition.

The VU section is the one place that must state the values cannot be zero.

The .NOTE is meant to be helpful, as the interaction is non-obvious, and multiple shipping applications were found to blow up when the window size became zero. The .NOTE is not essential, although it seems helpful.

If the normative language in the parameter definition does not need to mention zero values, then I agree with removing the non-zero clause from this section, and relying instead on the normative language in the VU section.  I defer to others to explain the requirement on normative language in the parameter section.

TobiasHector: Generally the parameter descriptions are short and to the point. Anything more verbose goes to the "description" text, and we only usually mention VU stuff in the actual description text if we feel it aids understanding - otherwise it stays exclusively in the VU.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
jjulianoatnv: I dislike the proposed additional text because it is not a complete list, but as written it suggests it might be. That creates ambiguity.

A key reason for an app to use the oldSwapchain parameter is that a new swapchain cannot be created that uses the same window object. The app must use the oldSwapchain parameter, or it must destroy the old swapchain before creating a new swapchain over the same window object.  Some APIs have swapchain resize, whereas Vulkan has oldSwapchain.

Another key reason is that the memory for non-acquired images from oldSwapchain can immediately be reused by the new swapchain. This can be very important when the swapchain images consume a sizable fraction of total graphics memory.

Continued use of previously-acquired images from oldSwapchain can lead to a more glitch-free experience. It can be important too.

This passage is not the right place to enumerating all of the reasons described above, and additional reasons if there are additional reasons.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
jjulianoatnv: Thanks for that clarification. I was overly verbose, and the proposal make sense to remove this "zeroness" language from the parameter description.

krOoze:
Updated w/ suggestions
+ some additional tweaks
+ rewritten the newly added note

ianelliottus:
This version LGTM.  Thanks for the edits!

Issue #331, opened 2016-08-13
List of name decoration/markup issues


krOoze:
This is a consolidated list of minor formating issues not worth making separate Issue for each one.

I am gonna **add** to this list as I see something reading the spec and I am gonna **remove** fixed items (from time to time).

- Probably missing decoration for "height" and "depth" several times in `VkImageCreateInfo` VU section.
- All  auto-generated composite names in VU or in Synchronization section (e.g. `commandBufferCount` and `commandPool` in `vkAllocateCommandBuffers` VU and Synchronization section. or e.g. `pWaitSemaphores` and `pSignalSemaphores` in Synchronization section of `vkQueueSubmit`)
- `sizeof()` operators should perhaps be decorated too (e.g. in `vkCmdDrawIndirect` vkCmdDrawIndirect VU)
- In VU for `vkCmdSetDepthBias` there is some problem with the `o.0`
- lot of the extension appendixes (or how do you call em?) have various undecorated names (e.g. `gl_Position.y` in `VK_KHR_maintenance1`, or `sType` and `pName` in `VK_KHR_get_physical_device_properties2` extension appendix
- Words like "required" and variants are not purple (e.g. last item in `VkShaderModuleCreateInfo` VU)
- The extension names should perhaps have some decoration &mdash; they look weird in plain text. (e.g. "VK_KHR_surface" in 30.2. WSI Surface)
- Undecorated `may`s in 31. Extended Functionality chapter (counted 3x)
- In Features, Limits, and Formats ch. there's "Note on extensibility". I think the styleguide does not allow such type of Note.
- "NOTE" (in capitals) in 30.2.1. Android Platform
- There is twice "z coordinate" in plain text in vkCmdBlitImage description. ("z" should perhaps be italicized or something)
- Undecorated NULL in `vkEnumerateInstanceExtensionProperties` and `vkEnumerateDeviceExtensionProperties`
- Undecorated "may" in `object` desc. of `vkDebugReportMessageEXT`
- Undecorated `VK_ERROR_MEMORY_MAP_FAILED` in Note in memory chapter
- Undecorated `VK_ERROR_OUT_OF_DATE_KHR` in `vkGetSwapchainCounterEXT` descriptions.
- Undecorated and misnamed `DebugReportCallback` in `pUserData` parameter desc. of `PFN_vkDebugReportCallbackEXT`
- Undecorated `VkDisplayKHR` in `vkGetRandROutputDisplayEXT` descriptions
[...]

PR #600, opened 2017-10-26
WIP: Random editing

krOoze:
Fix random markup, formatting and typos.

Edit @oddhack: marked the title WIP as this will not be accepted in its monolithic form.

### DIFF
--- a/doc/specs/vulkan/appendices/VK_EXT_debug_report.txt
+++ b/doc/specs/vulkan/appendices/VK_EXT_debug_report.txt
@@ -12,7 +12,7 @@ include::meta/VK_EXT_debug_report.txt[]
 
 Due to the nature of the Vulkan interface, there is very little error
 information available to the developer and application.
-By enabling optional validation layers and using the +VK_EXT_debug_report+
+By enabling optional: validation layers and using the `VK_EXT_debug_report`
 extension, developers can: obtain much more detailed feedback on the
 application's use of Vulkan.
 This extension defines a way for layers and the implementation to call back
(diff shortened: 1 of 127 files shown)


oddhack:
While I appreciate the intent, this is too much unrelated material for me to digest in a single PR. I'd be reviewing it for the next couple of hours, and almost any change we commit from the Khronos side is likely to cause merge conflicts. Maybe @tobias is able to put the time into it but I can't.

I wouldn't want it broken down at the per-file level either, but something like grouping it in PRs by category of markup you're addressing would be more accessible and more likely to actually get reviewed sometime in the relatively near future. In particular I don't want things that are just markup fixes or minor wording changes mixed in with changes to normative spec language, which IP-owning companies need to look at very carefully.

TobiasHector:
Holy moly - no this is way too much even for me. I don't even care too much how the split is done, but I am not going to be able to sit down and review all of this in one sitting 😛 

Split it any way you see fit @krOoze - but please split it or I don't think anyone's ever going to review this 😅

krOoze:
Heh, it is bit of a logistical conundrum, ain't it?
Splitting and categorizing this stuff will take some time too.
All of it should be non-semantic. The worst I did is change some **can**s and **may**s if I was confident enough.

krOoze:
Anyway then, consider this a semi-private list (akin to #331) where I stage the small issues that I find randomly from time to time (anyone adding here is welcome too).

I will find a way to portion this, so we can burn this list down week by week....

doc/specs/vulkan/appendices/VK_EXT_debug_report.txt
```diff
@@ -51,9 +51,9 @@ to the application for events of interest to the application.
 
 === Examples
 
-+VK_EXT_debug_report+ allows an application to register multiple callbacks
+`VK_EXT_debug_report` allows an application to register multiple callbacks
 with the validation layers.
-Some callbacks may log the information to a file, others may cause a debug
+Some callbacks can: log the information to a file, others can: cause a debug
```
courtney-g: Why can:? I deliberately chose 'may' because this section describes examples and isn't specifying extension behavior.

krOoze: Because it is describing "something that the Application can achieve", **not** "something that is optional to the Implementation" as per 1.3. Terminology chapter.

krOoze:
OK, most of this trash pile should be exported to other PRs. There are two or three things left. But this is DNM, so closing.

PR #602, opened 2017-10-26
Fix random math markup

krOoze:
incl. two unparsable latexmath blocks

exported from #600 

PS: It would be nice to note in styleguide how to do a block equation using the [eq]. (Make it just separate paragraph? Or envelop it in something?)

PS2: Variables in `[eq]` should probably be italicized to match latexmath. Also worth noting in styleguide...

### DIFF
--- a/doc/specs/vulkan/appendices/VK_NV_clip_space_w_scaling.txt
+++ b/doc/specs/vulkan/appendices/VK_NV_clip_space_w_scaling.txt
@@ -18,27 +18,28 @@ to the final post-processed image.
 This extension provides a mechanism to render VR scenes at a non-uniform
 resolution, in particular a resolution that falls linearly from the center
 towards the edges.
-This is achieved by scaling the "w" coordinate of the vertices in the clip
-space before perspective divide.
-The clip space "w" coordinate of the vertices can: be offset as of a
-function of "x" and "y" coordinates as follows:
+This is achieved by scaling the [eq]#w# coordinate of the vertices in the
+clip space before perspective divide.
+The clip space [eq]#w# coordinate of the vertices can: be offset as of a
+function of [eq]#x# and [eq]#y# coordinates as follows:
 
-            w' = w + Ax + By
+[eq]#w' = w + Ax + By#
 
 In the intended use case for viewport position scaling, an application
 should use a set of 4 viewports, one for each of the 4 quadrants of a
 Cartesian coordinate system.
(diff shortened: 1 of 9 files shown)


doc/specs/vulkan/appendices/VK_NV_clip_space_w_scaling.txt
```diff
@@ -18,27 +18,28 @@ to the final post-processed image.
 This extension provides a mechanism to render VR scenes at a non-uniform
 resolution, in particular a resolution that falls linearly from the center
 towards the edges.
-This is achieved by scaling the "w" coordinate of the vertices in the clip
-space before perspective divide.
-The clip space "w" coordinate of the vertices can: be offset as of a
-function of "x" and "y" coordinates as follows:
+This is achieved by scaling the [eq]#w# coordinate of the vertices in the
```
oddhack: I'm not sure why you're converting plaintext "w" into [eq]#w# here, while converting [eq]#s# into *s* below. For single variables it makes little difference, but consistency is nice.

I like having the [eq] style tag on equations in order to keep track of them, but haven't yet customized a stylesheet to do anything special. Eventually they should use the same font and italicization conventions as latexmath markup does.

oddhack:
Generally for block asciidoc math equations I offset them by e.g.

> :: [eq]#foo = bar#

This isn't ideal but asciidoctor really hates the idea of explicit layout control. It could probably be done in the stylesheet once we have one - I'm experimenting with the asciidoctor stylesheet factory now for other purposes and may get to this soon.

doc/specs/vulkan/appendices/VK_NV_clip_space_w_scaling.txt
krOoze: If it appears in an equation somewhere I make it `[eq]` too.
If it doesn't I make it bold with `*`, which is more noticable.
I think elsewhere @TobiasHector advised me to sometimes use `*var*`.

krOoze: PS: not sure if a style can help things. Mostly we would probably want only the names in italic, while the rest (e.g. operators + -  , or function names) straight.

krOoze:
That would be relevant to [L26](https://github.com/KhronosGroup/Vulkan-Docs/pull/602/files#diff-abd2061cb52b4f6edc3a6f69be6d7893R26). But I guess it is fine as separate paragraph.

PR #605, opened 2017-10-27
Fix missing "valid" in some obscure cases

krOoze:
- update script to detect better structs that do not need "valid"
- update `vk.xml` and schema to say `noautovalidity` as a hint to the script
- fix missing `const` in `VkValidationFlagsEXT`
- update some cases that were probably avoiding the bug

complements changes in PR #589

### DIFF
--- a/chapters/VK_KHR_display/display.txt
+++ b/chapters/VK_KHR_display/display.txt
@@ -411,6 +411,13 @@ include::../../api/structs/VkDisplayModeParametersKHR.txt[]
 For example, a 60Hz display mode would report a pname:refreshRate of 60,000.
 ====
 
+.Valid Usage
+****
+  * The pname:width member of pname:visibleRegion must: be greater than `0`
+  * The pname:height member of pname:visibleRegion must: be greater than `0`
+  * pname:refreshRate must: be greater than `0`
+****
+
 include::../../validity/structs/VkDisplayModeParametersKHR.txt[]
 --
 
@@ -448,16 +455,6 @@ include::../../api/structs/VkDisplayModeCreateInfoKHR.txt[]
     describing the display parameters to use in creating the new mode.
     If the parameters are not compatible with the specified display, the
     implementation must: return ename:VK_ERROR_INITIALIZATION_FAILED.
-
-.Valid Usage
-****
-  * [[VUID-VkDisplayModeCreateInfoKHR-width-01250]]
-    The pname:width and pname:height members of the pname:visibleRegion
-    member of pname:parameters must: be greater than `0`
-  * [[VUID-VkDisplayModeCreateInfoKHR-refreshRate-01251]]
-    The pname:refreshRate member of pname:parameters must: be greater than
-    `0`
-****
 
 include::../../validity/structs/VkDisplayModeCreateInfoKHR.txt[]
 --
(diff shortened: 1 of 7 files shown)


doc/specs/vulkan/chapters/pipelines.txt
```diff
@@ -617,7 +617,8 @@ endif::VK_KHR_maintenance2[]
     If no element of the pname:pDynamicStates member of pname:pDynamicState
     is ename:VK_DYNAMIC_STATE_VIEWPORT, the pname:pViewports member of
     pname:pViewportState must: be a valid pointer to an array of
-    pname:pViewportState::pname:viewportCount sname:VkViewport structures
+    pname:pViewportState::pname:viewportCount valid sname:VkViewport
```
tobine: VkViewport doesn't have an sType. Is is appropriate to to refer to "valid structures" without an sType? What makes it a valid structure?
There's a separate VU for "vkCmdSetViewport()" that refers to "valid VkViewport structures" and I agree that these two should be consistent. Just not sure "valid * structure" makes sense for structs w/o sTypes and reviewing "Valid Usage for Structure*" sections of [spec](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#fundamentals-validusage-sType) does not shed any light on how a non-sType struct could/would be "valid."

krOoze: Yes. You are looking at wrong chapter of implicit valid usage of `sType`. Look here:

> Valid usage conditions are described in a block labelled “Valid Usage” following **each** command or **structure** they apply to.

`VkViewport` has Valid Usage section and therefore needs to be said it has to be "valid".
Lack of "valid" means that structure either is returned only in that role, or it is guaranteed to be always valid (i.e. has no Valid Usage section; be it implicit or explicit).

Tobski: > PS BTW: I would be for writing "valid" everywhere applicable, as it makes the spec less brittle. E.g. someone may decide later on to add some VUs to VkRect2D too.

We could alternatively just state that "must: be an X" where the parameter is readable implies it must be a valid X? Only non-const pointers are written, so should be easy enough to write something sane up. I'd be ok with that.

oddhack:
@krOoze we're ready to accept this, once conflicts are resolved.

krOoze:
@oddhack  rebased.

src/spec/validitygenerator.py
```diff
@@ -465,11 +465,14 @@ def makeAsciiDocLineForParameter(self, blockname, param, params, typetext):
 
         return asciidoc
 
-    # Try to do check if a structure is always considered valid (i.e. there's no rules to its acceptance)
+    # Check if a structure is always considered valid (i.e. there's no rules to its acceptance)
     def isStructAlwaysValid(self, blockname, structname):
```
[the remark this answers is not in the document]
krOoze: Well, for the purpose of unknown presence of explicit VUs I documented the [`noautovalidity` hint](https://github.com/KhronosGroup/Vulkan-Docs/pull/605/files#diff-1e507a11436f18f2583412c8ebc1a70c).

There is certainly a lot of space for further refactoring.

Issue #608, opened 2017-11-05
Accompanying destruction functions are not easily accessible in the spec


avoroshilov:
Hello!

In online registry pages like this one: https://www.khronos.org/registry/vulkan/specs/1.0/man/html/vkCreateImage.html
or in the spec pages, the accompanying destruction functions are not referenced near to the creation functions. They could only be accessed if to go to the upper level (through the **See also** section):
https://www.khronos.org/registry/vulkan/specs/1.0/man/html/VkImage.html
and then reference to `vkDestroyImage` will be somewhere in the middle of **See also** block, not immediately obvious too.

So the suggestion would be to:
1. add accompanying destruction functions to the **See also** section of the creation function and vice versa for the destruction functions (i.e. `vkCreateImage` will reference `vkDestroyImage`, and `vkDestroyImage` will reference `vkCreateImage`);
2. put the lifetime management functions at the beginning of the **See also** block of the object itself (e. g. `VkImage` **See also** block will have `vkCreateImage` as a first reference and `vkDestroyImage` as a second) - or maybe even have a separate subsection specifically for that

Thank you!

oddhack:
Thanks, that's a good suggestion. Parts 1 and 2 can be done by adding the proper commands in the xrefs= attribute of the open block surrounding each relevant command, which is easy but tedious. It would require some changes to the ref page extraction scripts to do part 3. I'll take a look at this as time allows, or you can propose a PR to the repository making these changes.

avoroshilov:
I forked the Vulkan-Docs repository today and played around with it, but initially as I didn't know that references could be implicit, I failed to follow your instructions at first.
Instead, I went the other way, and modified **doc\specs\vulkan\genRef.py**, function `seeAlsoList` (see [this commit](https://github.com/avoroshilov/Vulkan-Docs/commit/8e4661ea611dead664f1d22609a9df4601d1eecd) for the exact code), which allowed to create automatic matching of the functions.
There are 2 main reasons why I don't like this approach:

1. the pairs and priority prefixes are defined inside the script, while script should be agnostic to the actual vulkan values;
2. there are functions that do not have strictly matching pairs, like `vkCreateMirSurfaceKHR` which will have matching `vkDestroySurfaceKHR`; the pair search that I implemented has a verification via presence in `mapDict.keys`, so "invalid" functions should never get to the **See also** list, i.e. `vkDestroyMirSurfaceKHR` will not be a thing listed/displayed;

[...]

PR #615, opened 2017-11-11
PR for automatic paired functions cross-referencing

avoroshilov:
Originally defined in issue #608
**Short description**:
1. Two config files added: `doc/specs/vulkan/config/vulkan-api-tuples.txt` and `doc/specs/vulkan/config/vulkan-api-sorting.txt`.
2. `vulkan-api-tuples.txt` defines paired functions, pairs delimiter is empty string.
3. `vulkan-api-sorting.txt` defines the order how prefixes appear in the cross-references list, `*` (meaning "any other function) denotes place where functions that didn't match any prefix go; function keep case-sensitive alphabetical order within the prefix, as it was prior to this change.
4.  `doc/specs/vulkan/genRef.py` is modified to read, process and apply data supplied by the two config files mentioned.
5. `doc/specs/vulkan/Makefile` modified to have the config files as dependencies, to facilitate rebuild when config files changed
6. `doc/specs/vulkan/chapters/pipelines.txt` modified to have explicit cross references, due to the way paired functions are named: (vkCreateComputePipelines, vkCreateGraphicsPipelines) vs vkDestroyPipeline.

The idea behind this way of implementing the cross-references is [in this comment](https://github.com/KhronosGroup/Vulkan-Docs/issues/608#issuecomment-342226603).

There is a problem with pairing functions, provided by extensions (e.g. `vkCreateSwapchainKHR` vs `vkDestroySwapchainKHR`) - they are NOT automatically generated. Manual cross-referencing won't work too, because `doc/specs/vulkan/vkapi.py` simply doesn't have anything but the core functions, hence auto-cross-reference validation will fail and functions won't appear on the reference list. Manual referencing via adding `xref` - won't work either, because `macroPrefix` function in `doc/specs/vulkan/genRef.py` will not recognize the reference properly and return **"UNKNOWN"** prefix. Hence, in this revision paired extension functions do not get cross-references.

### DIFF
--- a/doc/specs/vulkan/Makefile
+++ b/doc/specs/vulkan/Makefile
@@ -285,7 +285,7 @@ MANCOPYRIGHT = $(MANDIR)/copyright-ccby.txt $(MANDIR)/footer.txt
 # also generated, though they are not useable at present.
 
 LOGFILE = man/logfile
-man/apispec.txt: $(SPECFILES) genRef.py reflib.py vkapi.py
+man/apispec.txt: $(SPECFILES) genRef.py reflib.py vkapi.py config/vulkan-api-sorting.txt config/vulkan-api-tuples.txt
 	$(PYTHON) genRef.py -log $(LOGFILE) $(SPECFILES)
 
 # These dependencies don't take into account include directives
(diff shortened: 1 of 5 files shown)


Khronoswebmaster:
We changed the CLA to acknowledge that changes to the specification source documents fall under the Khronos IP policy, unlike changes to the Apache-licensed scripts. This change requires you, the contributor, to re-sign the CLA before your contribution can be accepted.

Issue #568, opened 2017-09-24
vulkan.h collision with future C++ keyword


krolli:
Hi,
I was just experimenting with C++ modules under Visual Studio in an existing codebase that uses vulkan, and I ran into problem with VkPipelineShaderStageCreateInfo as it has member named 'module'. According to current version of the TS for modules, 'module' is likely to become a keyword in the language and so this produces a bunch of errors.
I'm unsure whether to bring this up with C++ people or here, but since I don't know of any place to report this for C++, I thought I'd start here. C++ folks probably know that adding new keywords will break stuff, so I guess this wouldn't be news for them anyway.

NicolBolas:
I don't think the C++ committee is going to choose or not choose to keyword `module` based solely on Vulkan using it as a member name. They're aware that keywords can break code, and they're aware that "module" is an identifier that has significant use in the world.

The question is really how can Vulkan get out of C++'s way on this. Would it be possible to create a `VkPipelineShaderStageCreateInfo2`, but one which has the same layout (and `stype` value) as the original?

ratchetfreak:
the names of the fields in pod structs don't really matter, only the order and sizes. So a rename would be the simple option

TobiasHector:
We noted this discrepancy last week in fact, and are looking into it. I'm not sure what the agreed upon resolution was because for some reason it didn't get written into the internal bug, but I know there is a resolution here which wasn't just "tell C++ to go away".

Note that @ratchetfreak is correct though - the binary interface is preserved if we renamed the struct member, so existing projects using the current header would continue to work. Only developers updating the header would have any issues, and the expectation is that this member is likely not referenced in too many places throughout anyone's codebase (so impact should be small).

I'll try to keep you posted when we've agreed a resolution on this one.

fraggamuffin:
We have submitted this paper to the C++ Std as to object to the naming, as this is really the last chance when they are triaging Modules Ballot comment from the National Body and each nation/company has a lot of power to change it at this moment. Please consider asking your company rep to support this paper at the next C++ Std meeting in Albuquerque on Nov 6-11. 
[P0795R0_ From Vulkan with love_ a plea to reconsider the Module Keyword to be contextual.pdf](https://github.com/KhronosGroup/Vulkan-Docs/files/1388020/P0795R0_.From.Vulkan.with.love_.a.plea.to.reconsider.the.Module.Keyword.to.be.contextual.pdf)

krOoze:
[...]

PR #616, opened 2017-11-12
Add compatibility with C++ `module` keyword

krOoze:
Add compatibility with C++ `module` keyword as required by issue #568 .

User wishing to use experimental modules support in compiler can use alternative (wrapper) header `vulkan.cppcompat.h` and access the member as `shaderModule`.

Alternativelly he can provide `VK_PIPELINE_SHADER_STAGE_CREATE_INFO_MEMBER_MODULE_NAME` directly himself before first including `vulkan.h`. But I am not sure such use should be encouraged...

### DIFF
--- a/doc/specs/vulkan/registry.txt
+++ b/doc/specs/vulkan/registry.txt
@@ -466,6 +466,9 @@ contain several semantic tags:
     succeed. Builtin C types should not be wrapped in tag:type tags.
   * The tag:name tag is required, and contains the struct/union member
     name being described.
+  ** attr:cppcompatname is optional tag:name attribute. It contains
+     alternative name to be used if there is a collision with C++ standard
+     keywords.
   * The tag:enum tag is optional. It contains text which is a valid
     enumerant name found in another tag:type tag, and indicates that this
     enumerant must be previously defined for the definition of the command
--- a/src/spec/Makefile
+++ b/src/spec/Makefile
@@ -39,7 +39,7 @@ OUTDIR	= ../../out/1.0
 # clean - remove installed and intermediate files.
 
 VULKAN	    = ../vulkan
-HEADERS     = $(VULKAN)/vulkan.h
+HEADERS     = $(VULKAN)/vulkan.h $(VULKAN)/vulkan.cppcompat.h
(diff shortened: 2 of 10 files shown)


NicolBolas:
All generators for languages will have to look for name collisions and decide what to do about them should they come up. Alternate language-specific names of this sort, shouldn't be a formal part of the Vulkan specification. Either Khronos takes it on themselves to avoid collisions with various languages by setting the name, or the loader generators have to.

Personally, I would prefer the former, to avoid having different loaders using different member names arbitrarily. But in that case, it's Vulkan that needs to get out of the way. And not with alternate language names; by changing the actual name.

krOoze:
>  [...] And not with alternate language names; by changing the actual name.

That's preferable to me as well.
But some transitional period may be desirable.

> Either Khronos takes it on themselves to avoid collisions with various languages by setting the name, or the loader generators have to.

It is also not obvious how long will it take Khronos to decide this is the best course of action. If they even decide to go this (the former) way.
This PR goes the generator (latter) way in the meantime. It does not mean the former way cannot be chosen later.

> Alternate language-specific names of this sort, shouldn't be a formal part of the Vulkan specification.

Having the alternative name in `vk.xml` just seemed simpler to me. I suppose it can be in the generator script instead.

TobiasHector:
@krOoze We're deferring fixing this until we get some sort of final answer from the C++ committee, so we're not going to accept this right now.

Additionally, we've discussed this internally and had kind of settled on the idea of not doing anything complex - we would just flat out rename the variable and it'd be up to the header version to determine when the change happened should anyone wish to detect this - we wouldn't carry the old name permanently in any form.

krOoze:
@TobiasHector You might wait for 3 years though...
If you intend to change it, now might be as good time as any. Many compilers already use the keyword **now** independently of C++ commitee might decide.

TobiasHector:
As per issue #568, we got a response (well before 3 years @krOoze :P) and they've accepted the proposal to make it contextual, so we have no plans to make any changes to the specification/header to accomodate for this keyword.

Accordingly, I'm going to close this PR.

PR #621, opened 2017-11-20
Added some constants for working with vkCmdUpdateBuffer to prevent magic constants in user code

Jasper-Bekkers:
It appears that there are 3 different constants that come in to play when calling vkCmdUpdateBuffer that are mentioned in the spec. It would be really nice if there were defines in vulkan.h corresponding to these constants.

I may have missed updating some part of the spec / the naming might not be what we'll agree upon but it's a bit odd that these values are hard-coded as-is in the spec so I created this PR to clean it up a bit.

### DIFF
--- a/doc/specs/vulkan/chapters/clears.txt
+++ b/doc/specs/vulkan/chapters/clears.txt
@@ -495,13 +495,13 @@ include::../api/protos/vkCmdUpdateBuffer.txt[]
     recorded.
   * pname:dstBuffer is a handle to the buffer to be updated.
   * pname:dstOffset is the byte offset into the buffer to start updating,
-    and must: be a multiple of 4.
+    and must: be a multiple of ename:VK_UPDATE_BUFFER_OFFSET_ALIGNMENT.
   * pname:dataSize is the number of bytes to update, and must: be a multiple
-    of 4.
+    of ename:VK_UPDATE_BUFFER_SIZE_MULTIPLE.
   * pname:pData is a pointer to the source data for the buffer update, and
     must: be at least pname:dataSize bytes in size.
 
-pname:dataSize must: be less than or equal to 65536 bytes.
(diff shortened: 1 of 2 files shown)


krOoze:
This is not the place this appears. E.g. also in `VkBufferImageCopy` -- it probably maps to the same HW restriction...

Jasper-Bekkers:
If that's the case I'll defer to whoever is responsible within Khronos. But it seems like a decent idea to have a macro for that case as well (eg. VK_BUFFER_OFFSET_ALIGNMENT) to prevent user code having to have magic constants all over the place.

oddhack:
Unfortunately, this would have to be done as an extension to avoid breaking the API for compatibility with older 1.0 headers. I'll ask if the WG is interested in doing this in the future but we can't accept it in this form (also, additional changes to vk.xml would be required to get the symbols into the output headers).

Jasper-Bekkers:
How would that be a breaking change?

TobiasHector:
The Vulkan WG discussed this, and we came to the conclusion that we shouldn't be hard coding constants like this in the header in case we do decide to relax or change them in future - it would arbitrarily tie us to those values forever.

We're keeping our internal issue on this open as we believe there is an issue here - if we really think they might change then they should probably be queryable in some way. We mostly just don't think a `#define` in the header is the right way to go - but we're not sure what the right approach is yet.

As such we are not accepting this PR. If you want to carry on this discussion publicly, please feel free to raise an issue and we can continue to have a back and forth about this there.

Jasper-Bekkers:
@TobiasHector I disagree. Hardcoding these values in the spec is worse then hardcoding them in header. If you feel these values are likely to change make them query-able through the existing mechanisms instead.

TobiasHector:
No we agree with that - "just hardcoding them in the spec" is certainly not ideal, but hard coding them in the header isn't really a viable solution as some of these values are liable to change, hence not accepting this PR.

Again, we're keeping an internal issue open to consider how we might deal with this moving forwards, we just don't have an answer yet.

Issue #533, opened 2017-07-24
Specification should not assume extension is enabled


krOoze:
Specification should not assume an extension is enabled. That requires reader to read 2<sup>N</sup> cumulative combinations of the specification to use Vulkan safely.

For example the newest offender `VK_EXT_depth_range_unrestricted` removes important core Vulkan text this way:
```
.Valid Usage
 ****
+ifndef::VK_EXT_depth_range_unrestricted[]
   * [[VUID-VkClearDepthStencilValue-depth-00022]]
     pname:depth must: be between `0.0` and `1.0`, inclusive
+endif::VK_EXT_depth_range_unrestricted[]
 ****
```

(It also leaves behind ugly empty VU section.)
It should be preferred to make extension changes in such a way that the text is still semantically a valid Vulkan core specification if a particular extension is **not** enabled. E.g. for this case something like:

```
+ifdef::VK_EXT_depth_range_unrestricted[]
   * [[VUID-Whatever-is-not-useful-in-this-case-anyway-00042]]
     If `VK_EXT_depth_range_unrestricted` extension is not enabled pname:depth must: be between `0.0` and `1.0`, inclusive
+else
   * [[VUID-VkClearDepthStencilValue-depth-00022]]
     pname:depth must: be between `0.0` and `1.0`, inclusive
+endif::VK_EXT_depth_range_unrestricted[]
```

TobiasHector:
So we don't want to put "if n is enabled" all over the spec text, not least because it is a duplication of the semantic meaning of the ifdefs, but it'd also lead to some fairly awkward language hoops. We have plans afoot to do something where those ifdefs turn into semantic meaning in the output, but it's slow going because as per usual there's about a thousand other things to do. I'll treat this as a ping on that particular issue though :)

krOoze:
@TobiasHector I am gonna reping this one. It is super annoying that VUs dissapear just because you are reading the "wrong" version of the specification. (IMO this Issue is not Enhancement, but a Bug, because the rendered specification text is simply wrong in those cases)

I would caution against some automated solution. I am not sure it can work for all situations and sounds little bit too smartistic and apparently too time-intensive to implement.

oddhack:
The particular offender you identified (depth_range_unrestricted) will be addressed in the next spec update, although the general issue remains.

krOoze:
@oddhack Great!
[...]

PR #623, opened 2017-11-23
Clean up VkViewport VU section

### LINKED ISSUE #584: VK_AMD_negative_viewport_height placeholder VU
In `VkViewport` VU section it says:

- If the `VK_AMD_negative_viewport_height` extension is enabled, `height` **can** also be negative.

That's not really an orthogonal statement to the other VUs. "can" statements does not belong in this section.

### LINKED ISSUE #622: VK_KHR_maintenance1 missing sum VU
The VkViewport VUs say:

> - height **must** be greater than or equal to `-VkPhysicalDeviceLimits::maxViewportDimensions[1]` and less than or equal to `VkPhysicalDeviceLimits::maxViewportDimensions[1]`
> - `x` and `y` **must** each be between `viewportBoundsRange[0]` and `viewportBoundsRange[1]`, inclusive
> - `(y + height)` **must** be less than or equal to `viewportBoundsRange[1]`

This would allow `y=viewportBoundsRange[0]` and `height=-1.0`. That sum would be less than `viewportBoundsRange[0]`.

I think there is a missing VU. Something like:

> - `(y + height)` **must** be greater than or equal to `viewportBoundsRange[0]`

krOoze:
- fixes #622 
- fixes #584 

Should really say something like:

> - If `VK_KHR_maintenance1` or `VK_AMD_negative_viewport_height` is not enabled `height` **must** be greater than `0.0`
> - If `VK_EXT_depth_range_unrestricted` is not enabled `minDepth` **must** be between `0.0` and `1.0`, inclusive

but that's more of a systemic spec problem (#533)...

### DIFF
--- a/doc/specs/vulkan/chapters/vertexpostproc.txt
+++ b/doc/specs/vulkan/chapters/vertexpostproc.txt
@@ -680,32 +680,24 @@ The floating-point viewport bounds are represented with an
 
 .Valid Usage
 ****
-  * [[VUID-VkViewport-width-01227]]
-    pname:width must: be greater than `0.0` and less than or equal to
+  * pname:width must: be greater than `0.0`
+  * pname:width must: be less than or equal to
     sname:VkPhysicalDeviceLimits::pname:maxViewportDimensions[0]
-ifndef::VK_KHR_maintenance1[]
-  * [[VUID-VkViewport-height-01228]]
-    pname:height must: be greater than `0.0` and less than or equal to
+ifndef::VK_KHR_maintenance1,VK_AMD_negative_viewport_height[]
+  * pname:height must: be greater than `0.0`
+endif::VK_KHR_maintenance1,VK_AMD_negative_viewport_height[]
+  * The absolute value of pname:height must: be less than or equal to
     sname:VkPhysicalDeviceLimits::pname:maxViewportDimensions[1]
-endif::VK_KHR_maintenance1[]
-ifdef::VK_KHR_maintenance1[]
-  * [[VUID-VkViewport-height-01229]]
-    pname:height must: be greater than or equal to
-    -sname:VkPhysicalDeviceLimits::pname:maxViewportDimensions[1] and less
-    than or equal to
-    sname:VkPhysicalDeviceLimits::pname:maxViewportDimensions[1]
-endif::VK_KHR_maintenance1[]
-ifdef::VK_AMD_negative_viewport_height[]
-  * [[VUID-VkViewport-height-01230]]
-    If the +VK_AMD_negative_viewport_height+ extension is enabled,
-    pname:height can: also be negative.
-endif::VK_AMD_negative_viewport_height[]
-  * [[VUID-VkViewport-x-01231]]
-    pname:x and pname:y must: each be between pname:viewportBoundsRange[0]
-    and pname:viewportBoundsRange[1], inclusive
+  * pname:x must: be greater than or equal to pname:viewportBoundsRange[0]
   * [[VUID-VkViewport-x-01232]]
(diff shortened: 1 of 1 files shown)


oddhack:
@TobiasHector can you make a call on this, or should we take it back to the WG for approval?

doc/specs/vulkan/chapters/vertexpostproc.txt
```diff
@@ -680,32 +680,28 @@ The floating-point viewport bounds are represented with an
...
-    pname:height can: also be negative.
-endif::VK_AMD_negative_viewport_height[]
-  * [[VUID-VkViewport-x-01231]]
-    pname:x and pname:y must: each be between pname:viewportBoundsRange[0]
-    and pname:viewportBoundsRange[1], inclusive
+  * pname:x must: be greater than or equal to pname:viewportBoundsRange[0]
   * [[VUID-VkViewport-x-01232]]
     [eq]#(pname:x {plus} pname:width)# must: be less than or equal to
     pname:viewportBoundsRange[1]
+  * pname:y must: be greater than or equal to pname:viewportBoundsRange[0]
```
dgkoch: I think this should be viewportBoundsRange[1]

critsec:
LGTM

doc/specs/vulkan/chapters/vertexpostproc.txt
```diff
@@ -680,32 +680,28 @@ The floating-point viewport bounds are represented with an
...
     sname:VkPhysicalDeviceLimits::pname:maxViewportDimensions[0]
-ifndef::VK_KHR_maintenance1[]
-  * [[VUID-VkViewport-height-01228]]
-    pname:height must: be greater than `0.0` and less than or equal to
+ifndef::VK_KHR_maintenance1,VK_AMD_negative_viewport_height[]
+  * pname:height must: be greater than `0.0`
+endif::VK_KHR_maintenance1,VK_AMD_negative_viewport_height[]
+ifdef::VK_KHR_maintenance1,VK_AMD_negative_viewport_height[]
+  * pname:height must: be greater than or equal to
+    -sname:VkPhysicalDeviceLimits::pname:maxViewportDimensions[1]
```
TobiasHector: Hmm, it's kind of unclear to put the `-` in front of this given that it doesn't even remotely look like a mathematical statement otherwise - I think it'd make more sense to say explicitly "the negation of ..maxViewportDimensions[1]" or similar language.

Granted that's sort of an existing problem, but would be good to fix whilst you're here - the output just looks broken.

krOoze: That's probably not a bad idea. Would just wording out "minus" be better?

TobiasHector: "equal to minus blah" is weird phrasing IMO, I think that only works with numerical constants (e.g. "minus five").

"negative" sounds ok in speech but I think might be grammatically incorrect in this particular case...? Not sure. "Negative value" maybe?

I suggested "the negation of" because it's accurate, and mirrors where we've put "the sum/product/difference/quotient of" elsewhere in valid usage. I think in VU it pays to be formal, though I'd use something less formal if I were to express it elsewhere.

krOoze: Hmm. Alternatively "an absolute value of height must be less than or equal to maxViewportDimensions[1]".

TobiasHector: That would work, and wouldn't even need to be conditional based on extension.

krOoze: Alright, updated that way.

PR #624, opened 2017-11-27
Update Ubuntu subsystem build instructions

krOoze:
Update/simplify Ubuntu subsystem build instructions.

Resetting my environment, I did not encounter any gem compatibility problems currently, and did not need `rbenv` nor `rvm` at all.

`MATHEMATICAL_SKIP_STRDUP` does not seem to be needed anymore in new versions.

Did mot find what `gtk-doc-tools` is used for. Things seem to work without it.

### DIFF
--- a/BUILD.adoc
+++ b/BUILD.adoc
@@ -2,74 +2,82 @@
 :toc2:
 :toclevels: 2
 
+ifdef::env-github[]
+:note-caption: :information_source:
+endif::[]
 
 [[intro]]
 == Introduction
 
-This README describes important stuff for getting the Vulkan API
-specification and reference pages building properly.
+This README describes how to build the Vulkan API specification,
+reference pages, and\or other related targets.
 
+It documents how to set up your build environment, build steps and
+targets, and contains some troubleshooting advice.
 
 [[building]]
 == Building The Spec
 
 First, clone the Khronos Github repository containing the Vulkan
-specification to your local Linux, Windows, or Mac PC. The repository is
-located at https://github.com/KhronosGroup/Vulkan-Docs/ .
+specification to your local Linux, Windows, or Mac PC.
+The repository is located at https://github.com/KhronosGroup/Vulkan-Docs/.
 
 Next, install all the necessary build tools (see <<depends,Software
 Dependencies>> below).
 
-Finally, go to the root directory of your local repository clone, and
+Finally, go to the root directory of your local repository clone, and do
 
     $ make html
 
-builds an HTML5 specification output.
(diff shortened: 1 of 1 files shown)


TobiasHector:
@oddhack most of the bleeding edge stuff is in the ruby gems themselves, which are updated separately from the core Ruby, and most maintain decent compatibility across several revisions of Ruby core. The issue previously was simply that the version of Ruby available by default on Ubuntu was hideously out of date - ridiculously so. If that's no longer the case, then it's probably a good idea to have simpler instructions for people coming to this with fresh eyes. Any step we can take to make getting the toolchain setup easier is a win IMO, and this is one such change.

However not everyone is on the latest version of Ubuntu for whatever reason, so we probably should keep the more complex instructions around for a while - but they should shuffle down a peg to a section specifically for people dealing with older/other distros.

oddhack:
@tobias AFAICT - my recollection is a bit vague as it was last year and traumatic enough I wanted to forget it - trying to update a packaged gem installed with the OS with a later version from the gem maintainer is nigh-impossible (for starters, you have to do it with chroot privileges, and second, it screws up use of the package database to be changing system files out from underneath it). That's what led me to rbenv, and to completely wiping the system ruby infrastructure, which was interacting destructively and mysteriously at times with my installation. I'll grant that was on Debian, but Ubuntu is pretty close to Debian where packaging issues are concerned.

I don't object to having more instructions for a specific Ubuntu variant if things happen to just work OOTB in that case now - but I have zero expectation that situation will persist longterm.

TobiasHector:
@oddhack Oh right I know you run into issues if you install a gem package via `apt install` or equivalent instead of `gem install`, so that makes sense. I've managed that before, and that's still definitely an issue. I think identifying that as "troubleshooting" would suffice, and we should make a strong recommendation to not install ruby packages via apt install (which tbh I think is the recommendation from most Ruby users too).

krOoze:
@TobiasHector Yea, that's my take on this. Despite the impression given by the readme, the build is actually quite simple.

@oddhack I should have mentioned: this is tested on Ubuntu on Windows, which is a LTS (16.04). Guys using anything older probably know what they are getting themselves into.

oddhack:
So, where things stand with this is that if you want to restructure the change to give more prominence to things working OOTB on Ubuntu for Windows at present, that's cool. But I don't want the other information to disappear, for the potential benefit of those on different versions or OSes.

krOoze:
This PR is revived.

krOoze:
NTS: might also require `libssl-dev`

PR #630, opened 2017-11-30
Cute up contact list

krOoze:
- <s>make it one contact person per line</s>
- make github handles links

ex #600.

BTW some contacts seem to be wrong. 
Some of those @ are possibly meant to be Twitter, rest I dunno.
Some don't have "sufficient contact" only name. Some don't have names...

- `@tobias` probably links to different person, I have seen him around as `TobiasHector`
- Jeff Bolz lists `jbolz`, but I seen him around as `jeffbolznv`
- `pdaniell` probably links to different person, I have seen him around as `pdaniell-nv`
- `johnk` probably links to different person, senn him as `johnkslang`
- Alexander lists @debater, but seems to be around as alegal-arm
- Courtney Goeltzenleuchter lists @courtney and @courtneygo, but seems to be around as courtney-lunarg
- Carsten Rohde has no contact
- dominik.witczak@amd.com does not have name
- baldurk@baldurk.org does not have name
- Dominik Witczak lists dead @dominikwitczak_amd, probably should be DominikWitczakAMD
- quentin.lin@amd.com does not have a name
- Matthaeus G. Chajdas seem to be misspelled
- @jaakko wrong person, probably jaakkowork?

### DIFF
--- a/xml/extensionmetadocgenerator.py
+++ b/xml/extensionmetadocgenerator.py
@@ -248,6 +248,12 @@ class Extension:
         write('*Contact*::', file=fp)
         contacts = self.contact.split(',')
         for c in contacts:
+            # convert GitHub usernames to profile links
+            words = c.split();
+            for word in words:
+                if word[0] == '@':
+                    c = c.replace(word, 'link:https://github.com/' + word[1:] + '[' + word + ']')
+
             write('  * ' + c, file=fp)
 
         fp.close()


src/spec/extensionmetadocgenerator.py
```diff
@@ -133,7 +133,12 @@ def beginFeature(self, interface, emit):
                 write('  - Requires <<' + dep + ',`' + dep + '`>>', file=fp)
 
         write('*Contact*::', file=fp)
-        write('  - ' + contact, file=fp)
+        for contact in contacts:
```
oddhack: It's a good longterm goal, but many of these contacts are internal gitlab handles, not github, so it would make things worse to treat them all as valid links. And a bunch of Khronos participants don't have github handles, or at least not work-associated ones. So deferred for now.

krOoze: Ah, that makes sense.
Though the XML schema says they have to be reachable, and mentions GitHub specifically...

oddhack:
The 1.1.75 update either uses @handle for github handles, or gitlab:@handle for internal Khronos gitlab handles in the event they don't have a github account. N.b. this does not mean the contact will necc. be paying attention to the handle, as some of them were created just because I nagged at people to provide a public contact. Filing issues here is best.

I'm not sure, but the script changes here probably still function as intended with this handle syntax, since they appear to assume everything beginning with '@' is a github handle. So we could probably progress this now.

krOoze:
@oddhack I suppose so.

oddhack:
@krooze if you want to de-WIP and verify still working, we can accept this. It looks like the output logic won't try to turn an internal gitlab:@handle contact into a github link so it should be fine with the current contact format. Thanks.

krOoze:
I piggybacked the linebreaks on some other PR, which was actually the main thing.

I am not sure how much links to the GitHub profile are actually useful. I would be inclined to close this PR.

I think something more could be done here:

1) It is not obvious what the handles refer to, and "gitlab:" seems ugly to me
-> could add GitHub\GitLab logos there after username...

2) I could link it to new GitHub Issue page instead, and even prefill textbox with the Mention (assuming it is GitHub username), and perhaps prefill the extension name into the Issue title.

What do you think?

krOoze:
It is updated though, and seems to work. There seems to be no outright dead GitHub links.

oddhack:
It's up to you to close it or not. If you want to tag the github handles with icons that would be OK. I'm not sure it does much good to tag the gitlab handles from the point of view of external users, since they can't access the underlying gitlab server (maybe it would have some utility for internal discussion, but we pretty much all know each other anyway).

krOoze:
#770

PR #631, opened 2017-11-30
Cute up extension typography

### LINKED ISSUE #629: Description for VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE references extension by wrong name
The description for `VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE` (part of `VkSamplerAddressMode`) says:

> `VK_SAMPLER_ADDRESS_MODE_MIRROR_CLAMP_TO_EDGE` specifies that the mirror clamp to edge wrap mode will be used. This is only valid if the `VK_KHR_mirror_clamp_to_edge` extension is enabled.

The extension's name is `VK_KHR_sampler_mirror_clamp_to_edge`.

krOoze:
- make extension names links where appropriate
- use tick markup elsewhere
- fix #629

ex #600.

### DIFF
--- a/doc/specs/vulkan/appendices/VK_EXT_direct_mode_display.txt
+++ b/doc/specs/vulkan/appendices/VK_EXT_direct_mode_display.txt
@@ -37,16 +37,16 @@ None.
 === Issues
 
 1) Should this extension and its related platform-specific extensions
-leverage +VK_KHR_display+, or provide separate equivalent interfaces.
+leverage <<VK_KHR_display>>, or provide separate equivalent interfaces.
 
-*RESOLVED*: Use +VK_KHR_display+ concepts and objects.
-+VK_KHR_display+ can be used to enumerate all displays on the system,
+*RESOLVED*: Use <<VK_KHR_display>> concepts and objects.
+<<VK_KHR_display>> can be used to enumerate all displays on the system,
 including those attached to/in use by a window system or native platform,
-but +VK_KHR_display_swapchain+ will fail to create a swapchain on in-use
+but <<VK_KHR_display_swapchain>> will fail to create a swapchain on in-use
 displays.
 This extension and its platform-specific children will allow applications to
 grab in-use displays away from window systems and/or native platforms,
-allowing them to be used with +VK_KHR_display_swapchain+.
+allowing them to be used with <<VK_KHR_display_swapchain>>.
 
 2) Are separate calls needed to acquire displays and enable direct mode?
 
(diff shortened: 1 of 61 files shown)


krOoze:
1. I suppose. They should be benign though.
2. I think they are suppose to mean literal string. All the "_" were probably considered a hazard. But it should not trigger formatting, because asciidoc requires double "__" inside words.

krOoze:
PS: It is probably not the best solution, but it is the one I got. It would ideally be better to have the URLs elsewhere, but I am not quite sure how to do that.

TobiasHector:
> I suppose. They should be benign though.

Ah true, since the link formatting currently overrides code formatting. Would be useful to have consistency in case this changes though.

> I think they are supposed to mean "literal string"

Yea that sounds familiar. Don't know why I can't find this in the Asciidoctor user manual, although I wonder if this was more relevant for AsciiDoc...?

TobiasHector:
> PS: It is probably not the best solution, but it is the one I got. It would ideally be better to have the URLs elsewhere, but I am not quite sure how to do that.

Hum. If you want to centralise some of these links, you can set an attrib to the value of each registry (for instance), then use that attrib in the link. Think there's an example of this in the asciidoctor user manual somewhere...

doc/specs/vulkan/appendices/VK_EXT_direct_mode_display.txt
```diff
@@ -37,16 +37,18 @@ None.
...
 
 1) Should this extension and its related platform-specific extensions
-leverage +VK_KHR_display+, or provide separate equivalent interfaces.
-
-*RESOLVED*: Use +VK_KHR_display+ concepts and objects.
-+VK_KHR_display+ can be used to enumerate all displays on the system,
-including those attached to/in use by a window system or native platform,
-but +VK_KHR_display_swapchain+ will fail to create a swapchain on in-use
-displays.
+leverage <<VK_KHR_display,`VK_KHR_display`>>, or provide separate equivalent
```
oddhack: AFAICT it's fine to use the more compact markup - seems to do the right thing in both HTML and PDF outputs.  If you want to update the MR soon along these lines that would be good, otherwise when I accept it I'll make corresponding edits myself.

> `<<VK_KHR_display>>`

oddhack:
@TobiasHector re

> Yea that sounds familiar. Don't know why I can't find this in the Asciidoctor user manual, although I wonder if this was more relevant for AsciiDoc...?

That's my recollection. At this point all it does is call out the extension names, but they can be searched for with regexps anyway in the source. I think the \+VK_extname\+ markup can be done away with, without harm to anything, now.

doc/specs/vulkan/appendices/VK_EXT_direct_mode_display.txt
krOoze: on it

krOoze:
@oddhack Updated to the simpler link markup. Removed the external extension links -- TBD later.

PR #632, opened 2017-12-04
Fix random grammar

krOoze:
- fix random typos
- correct markup for quotation marks, apostrofes and em-dashes
- make some SPIR-V and OpenGL ES unbreakable

ex #600

### DIFF
--- a/doc/specs/vulkan/appendices/VK_AMD_rasterization_order.txt
+++ b/doc/specs/vulkan/appendices/VK_AMD_rasterization_order.txt
@@ -85,7 +85,7 @@ to use the extension.
 using the new relaxed mode?
 
 *RESOLVED*: No.
-The rasterization order is completely implementation dependent in this case
+In this case the rasterization order is completely implementation dependent,
 but in practice it is expected to partially still follow the order of
 incoming primitives.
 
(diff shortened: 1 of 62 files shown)


TobiasHector:
I started reviewing this but it's still too enormous to practically review in one sitting. It would be a big help I think to at least separate out the hyphenation change from everything else - that alone seems to have resulted in well over half the changes, and mixed in with everything else makes it rather a slog.

That said, I'm not entirely sure what the motivation behind changing all the dashes is, it seems somewhat pointless if it's done as a blanket change. It might be the right answer, but I'd like some notional explanation of the motivation for the change.

The same goes for the quotation marks - if we're blanket changing every instance of "x" to "`x`", then we should maybe consider other approaches.

krOoze:
@TobiasHector Sure, easy, I can do that.
The motivation is to further the goals of social grammar nationalism.

@oddhack The style guide is getting bit unwieldy. At some point it could be automatically checked. E.g. hyphen surrounded by spaces is almost guaranteed to be an error.

It is not really hard. People should know this from school. Em-dash is largest (literally the size of letter M) and you use it in places you could use comma, semicolon or parentheses.

En-dash is also usual typography. I dunno why asciidoctor does not support it. It is for time spans, and sports score separation.

Hyphen only belong inside (compound) words.

krOoze:
This should all be exported to other PRs now. Closing.

PR #634, opened 2017-12-04
Make all queue names starting with capital

krOoze:
ex #600


oddhack:
This is just inside the "Command Properties" tables, right @TobiasHector ? Otherwise we'd want to look at non-capitalized use of "graphics queue" / "compute queue" / etc. elsewhere in the spec, of which there are some examples.

krOoze:
@oddhack That should be so. This appears only in the table of the include files. It is already in asciidoc format, so I would assume it is not further processed.

PR #636, opened 2017-12-04
Fix Y'CbCr and R'G'B' markup

krOoze:
ex #600

### DIFF
--- a/doc/specs/vulkan/appendices/glossary.txt
+++ b/doc/specs/vulkan/appendices/glossary.txt
@@ -388,7 +388,7 @@ Execution Dependency Chain::
 
 ifdef::VK_KHR_sampler_ycbcr_conversion[]
 Explicit chroma reconstruction::
-    An implementation of sampler Y'C~B~C~R~ conversion which reconstructs
+    An implementation of sampler Y\'C~B~C~R~ conversion which reconstructs
     reduced-resolution chroma samples to luma resolution and then separately
     performs texture sample interpolation.
     This is distinct from an implicit implementation, which incorporates
@@ -575,7 +575,7 @@ Immutable Sampler::
 
 ifdef::VK_KHR_sampler_ycbcr_conversion[]
 Implicit chroma reconstruction::
-    An implementation of sampler Y'C~B~C~R~ conversion which reconstructs
+    An implementation of sampler Y\'C~B~C~R~ conversion which reconstructs
     the reduced-resolution chroma samples directly at the sample point, as
     part of the normal texture sampling operation.
     This is distinct from an _explicit chroma reconstruction_
--- a/doc/specs/vulkan/chapters/descriptorsets.txt
+++ b/doc/specs/vulkan/chapters/descriptorsets.txt
@@ -301,7 +301,7 @@ images that can: come from the same descriptor or from other combined or
 separate descriptor types.
 ifdef::VK_KHR_sampler_ycbcr_conversion[]
 A ename:VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER descriptor set entry that
-uses <<samplers-YCbCr-conversion,sampler Y'C~B~C~R~ conversion>> must: use
+uses <<samplers-YCbCr-conversion,sampler Y\'C~B~C~R~ conversion>> must: use
 the sampler with the sampled image from the same descriptor.
 endif::VK_KHR_sampler_ycbcr_conversion[]
 There are no additional restrictions on how a separate sampler or sampled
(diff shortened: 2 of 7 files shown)


fluppeteer:
Hmm. Good point; I'd tried to be more consistent about this in the Data Format Spec, but I lost the will to live with the corresponding Vulkan spec wording!

Technically I think we should be writing ```&#8242;``` rather than ```\'``` though (```&prime;``` doesn't seem to work).

krOoze:
Huh, that's what the "tics" are called...
Yeah, technically. Single quote mark seems to be the established ASCII replacement for the character though.

fluppeteer:
It doesn't always seem to be rendered the same as single quotes in all fonts. The PDF version looks identical to me - and different from the latexmath font version. However, at least as I view it (I didn't build this PR, but I did make a local edit to test options), the HTML version does show a prime (non-curling, diagonally sloped dash) that looks like the corresponding maths formulae.

I've not checked recently, but I believe the style guide used to frown on unicode, and we tended to use aliases - if only for consistency of editors. I admit that contrib/attribs.txt has inlined unicode characters, though. I guess it would be preferred to define ":prime:", but I don't know whether that's any clearer than inlining the HTML. Or just biting the bullet and using the unicode locally.

I'll poke Tobias and Jon for their preferred solution. ```\'``` is better than ```'```, but if we're going to fix it, I guess we should do it properly.

krOoze:
BTW If you ever decide to drop the non-unicode requirement, maybe get rid of the 76 line-width non-sense too &mdash; it serves only to confuse diffs.

fluppeteer:
Yes; attribs.txt includes definitions for unicode entities so they can be referred to with plain text in the rest of the spec, I believe. If we defined an alias for it, that's where ```:prime:``` would live. But I've asked the spec editors for their opinion - they periodically mess with things like this, so I'd not like to second-guess them.

The 76-width rule does have a big effect on legibility in a number of tools (including the web tools). It shouldn't be making the diffing appreciably _worse_ - we already have a "start each sentence on a new line" guideline to minimise the mess. It's true that occasionally an edit will invoke line wraps to the end of the sentence, and some diff tools would show you the subset of the sentence that was changed - but others would just tell you which line had changed. I think I'd be pushing back on changing that one without a lot of reassurance about all our tools. That was a long discussion last time, but I'm sure it'll get revisited at some point. :-)

I'm not sure how convincingly asciidoctor "sanitises" stuff, but we could probably tell it to if the output isn't to our liking!

krOoze:
> The 76-width rule does have a big effect on legibility in a number of tools (including the web tools). 

I wouldn't know. My editor already has word wrap; and so does GitHub.
Only one I can think of this affects is those that hack in console like it is the eighties.

On the other hand, you cannot see here what words I changed if I had to reflow the sentence to the 76 chars. Not to mention the extra work of reflowing it.

fluppeteer:
No problem, perfectly reasonable to ask.

"Those that hack in consoles" are most of the working group, myself included. My editor has word wrap (if I want it to); not everything else I work with does. I can certainly reflow a paragraph with a couple of keystrokes, though, and tools which do diffs ignoring line reflow shouldn't be hard to find (or write). Unfortunately, while it's been a little while since I checked, I believe the Khronos internal gitlab does _not_ do a sensible job of line wrap.

I'm sympathetic that a fixed line width restricts the use of screen area somewhat (yes, you can have more windows, and there's a reasonable amount of research suggesting a maximum comfortable line width, but still...) and it does complicate the default diff behaviour. But there are advantages and disadvantages to the approach, and I'm painfully aware of how long the discussion dragged out when we tried it before. That's not a reason we can't change, but it's a reason to have a very good reason!

Just please don't ask about tabs. :-)

krOoze:
> Just please don't ask about tabs. :-)

Death to the spaces!!! 😇

fluppeteer:
That noise is me grinding my teeth. :-)

But at least we can agree on prime symbols. (I'll save my campaign for the correct types of dash for later.)

fluppeteer:
Jon got back to me, and suggests defining ```{YCbCr}``` and ```{RGB}``` with the prime symbols (presumably in config/attribs.txt). I can try to take that on internally if you're happy for me to do so, rather than having you respin the PR here?

krOoze:
@fluppeteer Sure, if you want to, don't let me stop you.

fluppeteer:
Thanks. Might be a couple of weeks because I'm quite busy, but I'll get it done.

oddhack:
Closing - we'll track this internally.

oddhack:
Improvements were made in the 1.2.131 spec update.

PR #637, opened 2017-12-06
Refine swapchain association with surface

### LINKED ISSUE #438: Surface–Swapchain association restriction not consistent
This restriction:

> The native window referred to by `surface` must not already be associated with a swapchain other than `oldSwapchain`, or with a non-Vulkan graphics API surface.

seems too restrictive to me.

I think it should also be allowed that the surface can also be associated with the oldSwapchain of the oldSwapchain, and also with the oldSwapchain of the oldSwapchain of the oldSwapchain and so on.

(BTW, should perhaps be "nor" for clarity in the sentence)

Also:

> A native window cannot be associated with more than one swapchain at a time. Further, swapchains cannot be created for native windows that have a non-Vulkan graphics API surface associated with them.

That is in contradiction with the above. Apparently **can** be associated with two swapchains too.
Also this statement says "cannot", while the first quote says "must not" about the same thing, so which is it?

krOoze:
- make **must** statement a proper VU
- fix sentence saying only one swapchain can be associated
- fix languge saying **must** not instead of **cannot**
- clarify all the cases when swapcahin cannot be created
- fixes #438

### DIFF
--- a/chapters/VK_KHR_swapchain/wsi.txt
+++ b/chapters/VK_KHR_swapchain/wsi.txt
@@ -23,8 +23,8 @@ presentation.
 An application renders to the image, and then queues the image for
 presentation to the surface.
 
-A native window cannot: be associated with more than one swapchain at a
-time.
+A native window cannot: be associated with more than one non-retired
+swapchain at a time.
 Further, swapchains cannot: be created for native windows that have a
 non-Vulkan graphics API surface associated with them.
 
@@ -338,8 +338,9 @@ ifdef::VK_VERSION_1_1,VK_KHR_device_group[]
     1, pname:flags must: not contain
     ename:VK_SWAPCHAIN_CREATE_SPLIT_INSTANCE_BIND_REGIONS_BIT_KHR
 endif::VK_VERSION_1_1,VK_KHR_device_group[]
-  * [[VUID-VkSwapchainCreateInfoKHR-oldSwapchain-01674]]
-    pname:oldSwapchain must: not be in the retired state
+  * If pname:oldSwapchain is not dlink:VK_NULL_HANDLE, pname:oldSwapchain
+    must: be a non-retired swapchain associated with native window referred
+    to by pname:surface
   * [[VUID-VkSwapchainCreateInfoKHR-imageFormat-01778]]
     pname:imageFormat, pname:imageUsage, pname:imageExtent, and
     pname:imageArrayLayers must: be supported for ename:VK_IMAGE_TYPE_2D
@@ -454,14 +455,17 @@ endif::VK_VERSION_1_1,VK_KHR_device_group[]
 The pname:surface must: not be destroyed until after the swapchain is
 destroyed.
 
-If pname:oldSwapchain is not dlink:VK_NULL_HANDLE then pname:surface must:
-be associated with pname:oldSwapchain.
-Otherwise, the native window referred to by pname:surface must: not already
-be associated with another swapchain, and must: not already be associated
-with a non-Vulkan graphics API surface.
+If pname:oldSwapchain is dlink:VK_NULL_HANDLE, and the native window
+referred to by pname:surface is already associated with a Vulkan swapchain,
+ename:VK_ERROR_NATIVE_WINDOW_IN_USE_KHR must: be returned.
+
+If the native window referred to by pname:surface is already associated with
+a non-Vulkan graphics API surface, ename:VK_ERROR_NATIVE_WINDOW_IN_USE_KHR
+must: be returned.
+
 The native window referred to by pname:surface must: not become associated
-with a non-Vulkan graphics API surface before the swapchain has been
-destroyed.
+with a non-Vulkan graphics API surface before all associated Vulkan
+swapchains have been destroyed.
 
 Like core functions, several WSI functions, including
 fname:vkCreateSwapchainKHR return ename:VK_ERROR_DEVICE_LOST if the logical


krOoze:
Hmm, actually the automatic validity already generates awkwardly something similar:

> If `oldSwapchain` is a valid handle, it **must** have been created, allocated, or retrieved from `surface`

the `VkSurfaceKHR` is listed as parent for `VkSwapchainKHR`, but the logic by which this VU is created seems bit strained to me.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -347,6 +347,8 @@ ifdef::VK_KHX_device_group[]
 endif::VK_KHX_device_group[]
   * [[VUID-VkSwapchainCreateInfoKHR-oldSwapchain-01674]]
     pname:oldSwapchain must: not be in the retired state
+  * If pname:oldSwapchain is not dlink:VK_NULL_HANDLE, pname:surface must:
```
critsec: This should probably be more complete, echoing the language describing this field: "pname:oldSwapchain must: be dlink:VK_NULL_HANDLE or the non-retired swapchain associated with pname:surface."

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -436,11 +438,15 @@ endif::VK_KHX_device_group[]
...
 
-If pname:oldSwapchain is not dlink:VK_NULL_HANDLE then pname:surface must:
-be associated with pname:oldSwapchain.
-Otherwise, the native window referred to by pname:surface must: not already
-be associated with another swapchain, and must: not be already be associated
-with a non-Vulkan graphics API surface.
+If pname:oldSwapchain is dlink:VK_NULL_HANDLE, the swapchain cannot: be
+created if the native window referred to by pname:surface is already
+associated with a swapchain.
+If pname:oldSwapchain is a valid swapchain, the swapchain cannot: be
```
critsec: I think this condition is redundant with the VU statement you've added (with my suggested modifications). Right?

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: Well, yes. That is what the VU above this one says.
It would be a matter of preference, whether they should be two separate VUs or one.

From validation layers standpoint, these seemed like two different things probably checked in two different ways.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: Wrong.
This describes when `VK_ERROR_NATIVE_WINDOW_IN_USE_KHR` should be returned. It gets bit complicated to catch all corner cases:

1. `oldSwapchain == VK_NULL_HANDLE` and `surface` is not associated with any swapchain -> `VK_SUCCESS`
2. `oldSwapchain == VK_NULL_HANDLE` and `surface` associated with any (retired or non-retired) swapchain -> `VK_ERROR_NATIVE_WINDOW_IN_USE_KHR`
3. `oldSwapchain == valid` and `surface` is associated only with `oldSwapchain` -> `VK_SUCCESS`
4. `oldSwapchain == valid` and `surface` is associated only with `oldSwapchain` and other retired swapchains -> `VK_SUCCESS`
5. `oldSwapchain == valid` and `surface` is associated with non-retired swapchain that is not `oldSwapchain` -> `VK_ERROR_NATIVE_WINDOW_IN_USE_KHR`

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -442,11 +444,15 @@ endif::VK_KHX_device_group[]
...
 destroyed.
 
-If pname:oldSwapchain is not dlink:VK_NULL_HANDLE then pname:surface must:
-be associated with pname:oldSwapchain.
-Otherwise, the native window referred to by pname:surface must: not already
-be associated with another swapchain, and must: not already be associated
-with a non-Vulkan graphics API surface.
+If pname:oldSwapchain is dlink:VK_NULL_HANDLE, the swapchain cannot: be
+created if the native window referred to by pname:surface is already
+associated with a swapchain.
```
ianelliottus: Given the spirit of this pull request, I think you should end this sentence with "... associated with a non-retired swapchain."

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: @ianelliottus I guess we are on a level playing field regarding to what I meant few months ago. 😄 Let's see...

So, we have
> `oldSwapchain` **must** not be in the retired state

and

> If `oldSwapchain` is not `VK_NULL_HANDLE`, `surface` **must** be associated with `oldSwapchain`

those are the undefined behaviors. Those two should be kinda equivalent to @critsec propositon, which says the same thing as one VU statement:

> `oldSwapchain` **must** be `VK_NULL_HANDLE` or a non-retired swapchain associated with `surface`.

(I like his proposition, because it does not silently assume `oldSwapchain` is valid unlike the first VU statement. I probably just forgot to update the PR.)

<hr/>

Now we need to cover the cases that are supposed to return `VK_ERROR_NATIVE_WINDOW_IN_USE_KHR`.  That is where @critsec is wrong. This pinned paragraph should describe only such cases (and VUs must not interfere and let those cases through, otherwise they would be undefined behavior). So it cannot be redundant to the VU (original, nor proposed).

>If `oldSwapchain` is `VK_NULL_HANDLE`, the swapchain **cannot** be created if the native window referred to by `surface` is already associated with a swapchain.

is supposed to describe the case nr. 2 in previous post.

>If `oldSwapchain` is a valid swapchain, the swapchain **cannot** be created if the native window referred to by `surface` is already associated with a non-retired swapchain not specified in `oldSwapchain`.

is supposed to describe the case nr. 5 in previous post.

<hr/>

Yes, the wording is bit confusing, but it should cover **all** the possible inputs (and their states).

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: No, both retired, and non-retired swapchains should result in `VK_ERROR_NATIVE_WINDOW_IN_USE_KHR`.

E.g.
```
vkCreateSwapchain(surface, oldSwapchain=null, &swapchain1); // get a non-retired swapchain1; NW behind surface gets associated with swapchain1
vkCreateSwapchain(surface, oldSwapchain=swapchain1, &swapchain2); // swapchain2 becomes retired, get a non-retired swapchain2, NW behind surface gets associated with swapchain2
vkDestroySwapchain(swapchain2); // swapchain2 becomes invalid
vkCreateSwapchain(surface, oldSwapchain=null, &swapchain3); // surface still associated with retired swapchain1 -> should return VK_ERROR_NATIVE_WINDOW_IN_USE_KHR
```

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
krOoze: Oh yeah, I guess it actually conflicts with the VU.

We need to decide if nr. 5 should be undefined behavior or a `VK_ERROR_NATIVE_WINDOW_IN_USE_KHR`.

Technically it was undefided behavior (`must not`) before, then again `must not` was improperly used for the other cases too, so I cannot be 100 % sure this is what was initially meant.

krOoze:
Updated with suggestions.

Should match this list now:
1. `oldSwapchain == VK_NULL_HANDLE` and `surface.NW` **not** associated with anything -> `VK_SUCCESS`
2. `oldSwapchain == VK_NULL_HANDLE` and `surface.NW` associated with anything (retired, non-retired, or external swapchain) -> `VK_ERROR_NATIVE_WINDOW_IN_USE_KHR`
3. `oldSwapchain == valid` and  `oldSwapchain` is retired -> undefined behavior
4. `oldSwapchain == valid`, `oldSwapchain` is non-retired, and `surface.NW` is **not** associated with `oldSwapchain` -> undefined behavior
5. `oldSwapchain == valid`, `oldSwapchain` is non-retired, and `surface.NW` is associated with `oldSwapchain` (and potentially with some retired swapchain ascendants of `oldSwapchain`)  -> `VK_SUCCESS`
6. `oldSwapchain == valid`, `oldSwapchain` is non-retired, and `surface.NW` is associated with `oldSwapchain`, and with non-Vulkan swapchain  -> undefined behavior at the non-Vulkan swapchain creation time (not allowed during `VkSwapchain` lifetime).
<hr/>

7. `oldSwapchain == valid`, `oldSwapchain` is non-retired, `surface.NW` is associated with `oldSwapchain` , and `surface` is associated with another non-retired swapchain -> cannot happen because of nr. 2 (UB) and  nr. 4 (will fail) and nr. 5 (will make `oldSwapchain` retired); there never are two non-retired swapchains
8. `oldSwapchain == valid`, `oldSwapchain` is non-retired, `surface.NW` is associated with `oldSwapchain` , and `surface` is associated with a retired swapchain that is not ascendant of `oldSwapchain` -> cannot happen because of nr. 2 (will fail to create any non-retired swapchain, when orphaned retired swapchains exist) and nr. 3 (UB when no non-retired swapchain available). Retired swapchains always belong to the same retirement chain.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
ianelliottus: Oh, I see your point.  I'm not sure if we ever thought about that case.  Seems like we've thought about the case where a window is continually resized (e.g. dragging the corner of the window), and the application has several retired swapchains.  Each swapchain only becomes an oldSwapchain once, and will eventually be destroyed.  In the case you raise, since swapchain2 was fully destroyed before swapchain1 was destroyed (but is already retired), is it an error for oldSwapchain to be NULL.  I don't think that was our intent (but does match the old language).  I'll raise it internally.

krOoze: @ianelliottus Maybe something was lost in the previous updates. The old-old version was badly worded, but I vaguely remember it could have been interpreted that all retired swapchains need to be destroyed before the current swapchain. Which would simplify things.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -442,14 +443,16 @@ endif::VK_KHX_device_group[]
...
+
+The swapchain cannot: be created if the native window referred to by
+pname:surface is already associated with a non-Vulkan graphics API surface.
+
 The native window referred to by pname:surface must: not become associated
-with a non-Vulkan graphics API surface before the swapchain has been
-destroyed.
+with a non-Vulkan graphics API surface before all associated Vulkan
+swapchains has been destroyed.
```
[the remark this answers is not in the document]
krOoze: done

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
ianelliottus: Following up from an internal discussion.  We agree that in the scenario you outline, VK_ERROR_NATIVE_WINDOW_IN_USE_KHR is the appropriate result.

doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
```diff
@@ -454,14 +455,16 @@ endif::VK_VERSION_1_1,VK_KHR_device_group[]
 The pname:surface must: not be destroyed until after the swapchain is
 destroyed.
 
-If pname:oldSwapchain is not dlink:VK_NULL_HANDLE then pname:surface must:
-be associated with pname:oldSwapchain.
-Otherwise, the native window referred to by pname:surface must: not already
-be associated with another swapchain, and must: not already be associated
-with a non-Vulkan graphics API surface.
+If pname:oldSwapchain is dlink:VK_NULL_HANDLE, the swapchain cannot: be
```
ianelliottus: I just noticed your use of "cannot:" here and elsewhere.  It's consistent with the old text at the top of the file, but I think it wrong (or at least isn't the best) for what the intention of "cannot:" is (i.e. describing what an application cannot do).  The old language that uses "must:" (i.e. "... must: not already
-be associated with another swapchain ...") seems better.  Please fix that.

krOoze: Seems to fit in this case. The creation will fail if the app attempts to do so -> "cannot".
In contrast to "must", which would mean doing so is undefined behavior.

ianelliottus: I think the problem I'm having is with how you are placing the nouns and verbs (I'm not an English major, so I don't know the correct terms to describe what I'm trying to say, but there's some concepts around the subjects of verbs).  Your wording makes it sound like the swapchain is the subject of the "cannot:" (i.e. a swapchain is part of the implementation; for which "cannot:" cannot be used) instead of the application.  If you say the application cannot: do something I wouldn't object.

At this point, I prefer to keep the old wording.

cubanismo: I don't think either is correct.  "cannot:" (normative form) is supposed to be reserved for statements like "The application cannot: destroy physical devices," because there is actually no API to destroy a physical device.  It never refers to something affecting the implementation, which is arguably the case here.  Hence, it's used pretty sparsely in the spec, because there are few things Vulkan prevents applications from attempting at all ;-)

"must: not" is wrong for the reason @krOoze notes.

What we're really supposed to do, based on examples elsewhere in the spec, is refer to the error code generated.  I.e., "If pname:oldSwapchain is dlink:VK_NULL_HANDLE and the native window referred to by pname:surface is already associated with a swapchain, VK_ERROR_NATIVE_SURFACE_IN_USE_KHR must: be returned."

krOoze: That's a shame. **cannot** seem to be born for this situation;
**cannot** definition:

> This word means that the API and/or the execution environment provide no mechanism through which the application can express or accomplish the action described.

The app can "express" it but won't "accomplish" creating swapchain with surface already in use.

@cubanismo I somewhat dislike that style, because it is incareful with the RFC keyword, which is to be always absolute; TBH it lies. E.g. to me `VK_ERROR_OUT_OF_MEMORY` seems as much valid in the situation as `VK_ERROR_NATIVE_SURFACE_IN_USE_KHR`, but that is out of the question if one of them says "this one **must** be returned".

I am gonna change it that way, because as you say it seems to be the established style, but it is perhaps worth an Issue...

krOoze: updated...

ianelliottus:
Thanks @krOoze and @cubanismo, the new version LGTM.  I'm happy to recommend that we approve it for merging.

ianelliottus:
@oddhack the SI TSG approved this today to be merged.

PR #641, opened 2017-12-11
Fix CUBIC feature flag documentation

krOoze:
- correct ul below `VkFormatFeatureFlagBits`
- document in extension appendix


TobiasHector:
Gah wrong account again - Tobski is me.

PR #642, opened 2017-12-11
Fix some dash typography

krOoze:
ex #632

### DIFF
--- a/doc/specs/vulkan/appendices/VK_EXT_debug_report.txt
+++ b/doc/specs/vulkan/appendices/VK_EXT_debug_report.txt
@@ -174,7 +174,7 @@ slink:VkDebugReportCallbackCreateInfoEXT::pname:flags to be non-zero?
 requiring the sname:VkDebugReportCallbackCreateInfoEXT::pname:msgFlags at
 create-time be non-zero.
 One can imagine that apps may prefer it as it allows them to set the mask as
-desired - including nothing - at runtime without having to check.
+desired -- including nothing -- at runtime without having to check.
 
 3) What is the difference between ename:VK_DEBUG_REPORT_DEBUG_BIT_EXT and
 ename:VK_DEBUG_REPORT_INFORMATION_BIT_EXT?
--- a/doc/specs/vulkan/appendices/VK_EXT_swapchain_colorspace.txt
+++ b/doc/specs/vulkan/appendices/VK_EXT_swapchain_colorspace.txt
@@ -10,33 +10,33 @@ include::meta/VK_EXT_swapchain_colorspace.txt[]
 === New Enum Constants
 
   * Extending elink:VkColorSpaceKHR:
-  ** ename:VK_COLOR_SPACE_DISPLAY_P3_NONLINEAR_EXT - supports the Display-P3
-     color space and applies an sRGB-like transfer function.
-  ** ename:VK_COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT - supports the extended
+  ** ename:VK_COLOR_SPACE_DISPLAY_P3_NONLINEAR_EXT -- supports the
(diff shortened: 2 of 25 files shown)


fluppeteer:
Ohno. You know I was kidding when I mentioned fixing the dashes in #636, right?

I think you're technically correct - the best kind of correct. But I'm still expecting to get the blame for this.

I'd slightly prefer quite a lot of these to be colons rather than em-dashes, if we're fixing them, since they're an explanation of a term rather than an interruption of a sentence. I tend to be quite excessive with my use of colons, however, so YMMV.

krOoze:
@fluppeteer What is this "kidding" you speak of?

Nah, I already had them fixed long ago. I am just exporting it for easy review at the request of @TobiasHector .

fluppeteer:
Phew. I like it when I'm not to blame for things!

In that case, you're right that these dashes should be em-dashes. Whether they should be dashes at all is possibly more open to debate, but "not definitely wrong" seems like an improvement.

The colon war has started! And I'm bringing... oh.

krOoze:
> Phew. I like it when I'm not to blame for things!

Although, I should not look gift scapegoat in the mouth.

TobiasHector:
The issue I have with this is simply that it's super invasive and I don't imagine people will keep on with it - it's highly likely that it'll become inconsistent again shortly after this change. I wonder if there is some sane default behaviour here that could be applied to Asciidoctor instead in order to facilitate this, rather than manually changing things through the spec.

krOoze:
@TobiasHector In defense of asciidoc, it is relatively conventional markup though &mdash; Em-dash is commonly written as two hyphens on typewriters in forever.

oddhack:
I agree with @TobiasHector. This is something we would be dealing with forever, because the other folks who propose MRs are rarely going to remember it.

krOoze:
@oddhack Isn't it better to have it wrong some of the time, than all of a time.
The rationale to me sounds bit like "we will not fix a typo, because people will keep introducing typos".

TobiasHector:
@krOoze IMO this would be solved better with a "smart dash" in an asciidoctor extension, which we could add with some relatively trivial rules - e.g. something like "if there's numbers on both sides, en-dash, if there's spaces on both sides then it's an em-dash, and if there's no spaces it's a hyphen". That we everyone can just use hyphens and not worry too much about it - with the ability to put -- to force an em-dash if necessary.

krOoze:
@TobiasHector OK, but it seems unnecessarily complex solution to people not knowing proper grammar.
Might as well be printing correct your vs you're, and let people write whatever in the source.

TobiasHector:
@krOoze I think the difference in your heinous example there is that one of them is obvious in the source language - the other is not. :P

krOoze:
@TobiasHector Which is which? Both are obvious in the source.

TobiasHector:
No I mean obvious that you're doing the "right thing". Anyway, this is going nowhere very quickly :P

krOoze:
@TobiasHector As all grammar discussions do... :wink:

If you mean em-dash it is the correct way to write it as "--" for the last 30 years. It is not some obscure syntax Asciidoc invented to annoy you.

TobiasHector:
Sure, the issue is that most of us aren't technical writers by trade, we are engineers who happen to be able to write. Big distinction there - em-dash vs. en-dash vs. hyphen is lost on basically all of us, thus it's easy to trip up on. In this particular case I think an automatic hammer is really what's needed.

krOoze:
@TobiasHector You don't have to be a technical writer to get it -- every kiddo with Twitter or blog can do it. Actually us trained formal language experts (or as we folks say "programmers") can do even better &mdash; see.

But I am banging my head against a wall here. So automatic hammer will have to be used after all. (🤣 OK, I will see myself out. )

PR #645, opened 2017-12-11
Make SPIR-V and GL ES unbreakable

krOoze:
ex #632 
\+ some more of them while at it


TobiasHector:
Functionally seems fine to me, but I'd like @oddhack to weigh in with his opinion on this one from an aesthetic perspective.

oddhack:
@TobiasHector I think this is another case where people are just not going to remember to do it - and why should they have to? - so consistency would require someone (or some TBD tool) to go over the spec periodically. Also, I poked at web searches of asciidoctor and hyphenation and saw enough complaints and issues raised to make it fairly clear that hyphenation support isn't all that robust in the toolchain at present, and can interact badly with Asciidoctor-pdf in particular.

If it were essential to do something about this I'd probably rather we write the terms using attributes; but I don't think we actually need to do anything about this. If it were up to me alone I'd WONTFIX this with some notion to come back to it in the future when the toolchain has a more sophisticated notion of this stuff.

I don't think asciidoctor is structured around controlling typography at very fine levels, at least relative to say LaTeX, so don't favor putting a lot of effort into trying to do it anyway (esp. when there are so many functional spec issues to be worked).

krOoze:
@oddhack Well, it is not a WYSIWYG editor; while frienlier than LaTex, it is still a typesetting source file and requires the same kind of discipline.

Anyway, there happen to be like three times this is broken on two lines, which looks weird, 's all.

It is more of a probabilistic matter. It does not neccessarily be marked up correctly every time to reap the benefits most of the time. 
Though if it is marked up right, there is more chance it is marked up right in the future. Monkey see, monkey do.

krOoze:
Sigh, asciidoctor-pdf indeed does not work with this... closing for now.

PR #649, opened 2017-12-23
Cute up external extension links

krOoze:
cont. #631 

- make external extensions links
- centralize external links as a references in appropriate extension appendices 
  

PS: I am not sure what the deal is with Asciidoc adding some "[]" around the links, so I am providing the explicit link names into the anchors(which seems to work).


krOoze:
Closing this as I am not willing to keep it rebased.

Though _some_ style guide should be put in place manage these links. Links are so convenient than just plaintext. At the same time it is ugly and brittle to copy-paste URL on each occurence of the external extension name.

PR #664, opened 2018-02-21
Make VK_* constants into appropriate type

### LINKED ISSUE #662: VK_TRUE / VK_FALSE are signed, while VkBool32 is unsigned
As the title says, is this an oversight? Currently when using wrappers for the API functions one needs to explicitly cast to VkBool32 because of this.

### LINKED ISSUE #717: XML Type Issues
Why are these marked as a define? 
```
        <type category="define">struct <name>ANativeWindow</name>;</type>
        <type category="define">struct <name>AHardwareBuffer</name>;</type>
```
Although, reading over the schema doc I don't see how you would describe a forward declared struct? Does the struct/union type have any thing to mark them as a forward declaration? Otherwise a separate category. 

Additionally, why are these constants specified as enums if they are not an enum type? Moreover, some of these values are compile time dependent is that intended? For instance (~0U) may have a different bit width on 32 bit system vs a 64 bit system.
```
<enums name="API Constants" comment="Vulkan hardcoded constants - not an enumerated type, part of the header boilerplate">
        <enum value="256"   name="VK_MAX_PHYSICAL_DEVICE_NAME_SIZE"/>
        <enum value="16"    name="VK_UUID_SIZE"/>
        <enum value="8"     name="VK_LUID_SIZE"/>
        <enum               name="VK_LUID_SIZE_KHR" alias="VK_LUID_SIZE"/>
        <enum value="256"   name="VK_MAX_EXTENSION_NAME_SIZE"/>
        <enum value="256"   name="VK_MAX_DESCRIPTION_SIZE"/>
        <enum value="32"    name="VK_MAX_MEMORY_TYPES"/>
        <enum value="16"    name="VK_MAX_MEMORY_HEAPS" comment="The maximum number of unique memory heaps, each of which supporting 1 or more memory types"/>
        <enum value="1000.0f" name="VK_LOD_CLAMP_NONE"/>
        <enum value="(~0U)" name="VK_REMAINING_MIP_LEVELS"/>
        <enum value="(~0U)" name="VK_REMAINING_ARRAY_LAYERS"/>
        <enum value="(~0ULL)" name="VK_WHOLE_SIZE"/>
        <enum value="(~0U)" name="VK_ATTACHMENT_UNUSED"/>
        <enum value="1"     name="VK_TRUE"/>
        <enum value="0"     name="VK_FALSE"/>
        <enum value="(~0U)" name="VK_QUEUE_FAMILY_IGNORED"/>
        <enum value="(~0U-1)" name="VK_QUEUE_FAMILY_EXTERNAL"/>
        <enum               name="VK_QUEUE_FAMILY_EXTERNAL_KHR" alias="VK_QUEUE_FAMILY_EXTERNAL"/>
        <enum value="(~0U-2)" name="VK_QUEUE_FAMILY_FOREIGN_EXT"/>
        <enum value="(~0U)" name="VK_SUBPASS_EXTERNAL"/>
        <enum value="32"    name="VK_MAX_DEVICE_GROUP_SIZE"/>
        <enum               name="VK_MAX_DEVICE_GROUP_SIZE_KHR" alias="VK_MAX_DEVICE_GROUP_SIZE"/>
    </enums>

```

krOoze:
Constants should have a type appropriate to their use.
It potentially helps in C++ with template instantiation, function overloading, `auto`, and `decltype` ...

fixes #662 
fixes #717

### DIFF
--- a/appendices/boilerplate.txt
+++ b/appendices/boilerplate.txt
@@ -292,7 +292,7 @@ handle.
 It may be passed to and returned from Vulkan commands only when
 <<fundamentals-validusage-handles, specifically allowed>>.
 
-include::../api/defines/VK_NULL_HANDLE.txt[]
+include::../api/consts/VK_NULL_HANDLE.txt[]
 
 --
 
--- a/include/vulkan/vulkan_core.h
+++ b/include/vulkan/vulkan_core.h
@@ -46,9 +46,6 @@ extern "C" {
 #define VK_HEADER_VERSION 91
 
 
-#define VK_NULL_HANDLE 0
-
-
 #define VK_DEFINE_HANDLE(object) typedef struct object##_T* object;
 
 
@@ -92,15 +89,16 @@ VK_DEFINE_NON_DISPATCHABLE_HANDLE(VkDescriptorSet)
 VK_DEFINE_NON_DISPATCHABLE_HANDLE(VkFramebuffer)
 VK_DEFINE_NON_DISPATCHABLE_HANDLE(VkCommandPool)
 
+#define VK_NULL_HANDLE                    0
 #define VK_LOD_CLAMP_NONE                 1000.0f
-#define VK_REMAINING_MIP_LEVELS           (~0U)
-#define VK_REMAINING_ARRAY_LAYERS         (~0U)
-#define VK_WHOLE_SIZE                     (~0ULL)
-#define VK_ATTACHMENT_UNUSED              (~0U)
-#define VK_TRUE                           1
-#define VK_FALSE                          0
-#define VK_QUEUE_FAMILY_IGNORED           (~0U)
-#define VK_SUBPASS_EXTERNAL               (~0U)
+#define VK_REMAINING_MIP_LEVELS           UINT32_MAX
+#define VK_REMAINING_ARRAY_LAYERS         UINT32_MAX
+#define VK_WHOLE_SIZE                     UINT64_MAX
+#define VK_ATTACHMENT_UNUSED              UINT32_MAX
+#define VK_TRUE                           UINT32_C(1)
+#define VK_FALSE                          UINT32_C(0)
+#define VK_QUEUE_FAMILY_IGNORED           UINT32_MAX
+#define VK_SUBPASS_EXTERNAL               UINT32_MAX
 #define VK_MAX_PHYSICAL_DEVICE_NAME_SIZE  256
 #define VK_UUID_SIZE                      16
 #define VK_MAX_MEMORY_TYPES               32
@@ -3794,7 +3792,7 @@ VK_DEFINE_NON_DISPATCHABLE_HANDLE(VkDescriptorUpdateTemplate)
 
 #define VK_MAX_DEVICE_GROUP_SIZE          32
 #define VK_LUID_SIZE                      8
-#define VK_QUEUE_FAMILY_EXTERNAL          (~0U-1)
+#define VK_QUEUE_FAMILY_EXTERNAL          UINT32_C(0xFFFFFFFE)
 
 
 typedef enum VkPointClippingBehavior {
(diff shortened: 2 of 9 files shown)


NicolBolas:
This just adds to the amount of C that non-C/C++ header generators have to be able to parse and work with.

I know that `vk.xml` is meant for C, but do we really need to make it _this_ difficult for non-C/C++ implementations?

krOoze:
> I know that `vk.xml` is meant for C, but do we really need to make it this difficult for non-C/C++ implementations?

Yes. 😈 

Unless...
the xml could contain enough info to resolve this in the generator (and the `value` could be simple number).
There's also some issue that it is `enums` for reasons...

@NicolBolas I believe you were involved with making `alt-vk.xml`. By chance, do you have any progress with this particular issue?

NicolBolas:
> By chance, do you have any progress with this particular issue?

I never intended to solve this problem. This problem can only be solved at the level of the source vk.xml data. The only thing my format did was explicitly call out when a value is a number or a C-expression, so that you can tell when you need to parse a value and when you don't.

krOoze:
Alternatively more ambitious [krOoze/Vulkan-Docs/fix_constant_types2](https://github.com/krOoze/Vulkan-Docs/tree/fix_constant_types2) (need to also fix extension ordering though -- dependent `VK_KHX_device_group` appears before dependency `VK_KHX_device_group_creation`)

Sairony:
> I know that vk.xml is meant for C, but do we really need to make it this difficult for non-C/C++ implementations?
Is it really though? If it's a new xml attribute it would probably just be easier, since it would be possible to infer that VK_TRUE / VK_FALSE is related to VkBool32. Otherwise they can just ignore the attribute in which case nothings changed compared to as it is now.

NicolBolas:
@Sairony You should take a look at what's *actually* in the patch. It's not "a new xml attribute" describing the type. It's turning `value=1` into `value=(VkBool)1`.

Remember: this isn't a comment chain for your bug; it's a comment change for this particular pull request.

FacticiusVir:
On behalf of the "non-C/C++ implementations" folks, this is a horrible change. It still has all the complexity of parsing arithmetic expressions, but now we have to rewrite to include C consts like `UINT32_MAX`, functions like `UINT32_C(2)` (or is that a macro?) and arithmetic functions on arbitrary types `(~(VkDeviceSize)0)`, just so we can continue to infer the intended type.

Why not just add a type attribute and include that as a typecast in the C `#define`s? So:

`<enum type="VkDeviceSize" value="~0ULL" name="VK_WHOLE_SIZE"/>`

maps to

`#define VK_WHOLE_SIZE                     ((VkDeviceSize)~0ULL)`

It not's a breaking change for those of us with working code, but adds useful or usable information on both the XML & C header sides.

krOoze:
@FacticiusVir Well, I expected as much and started working on something like that in the new branch... though I would like to get rid of `<enum>` which should be reserved only for integral constants inside C `enum`, which makes it a breaking change.

It is a bit shame, because generally speaking it makes sense for constants to contain arbitrary expressions, not just raw numbers. Though the existing constants do not need it (except perhaps the `VK_API_VERSION_1_0`).

FacticiusVir:
@krOoze Ah, sorry, hadn't looked at the other branch properly - that's massively better, especially simplifying down to actual numeric literals. The shift from `<enum>` elements is just a minor schema change, so while it is breaking it's not as much work as parsing C expressions.

Sairony:
>@Sairony You should take a look at what's actually in the patch. It's not "a new xml attribute" describing the type. It's turning value=1 into value=(VkBool)1.
Remember: this isn't a comment chain for your bug; it's a comment change for this particular pull request.

Sorry, new to how this stuff works on github.

krOoze:
For popular demand swapped with the more invasive branch. And rebased to master.

WIP: For some reason the scripts make a parasitic change to `VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GROUP_PROPERTIES_KHR` and `VK_STRUCTURE_TYPE_DEVICE_GROUP_DEVICE_CREATE_INFO_KHR`.
There must be some hack based on the removed enums somewhere doing something...

PR #683, opened 2018-03-17
Clarify the interaction between multi-planar images and VK_IMAGE_USAGE_STORAGE_BIT

atomnuker:
Whilst the specifications are very clear that users may create single-plane VkImageView from multi-plane images, there's a concern from one implementor that flagging VK_IMAGE_USAGE_STORAGE_BIT as supported for such VkFormat will mean that creating a VkImageView with a format set to the original multi-plane VkFormat would be supported.
Hence, clarify that for multi-planar images, the VK_IMAGE_USAGE_STORAGE_BIT only applies to any single-plane VkImageViews created with a compatible VkFormat and not to VkImageViews created with the original multi-planar format.

### DIFF
--- a/doc/specs/vulkan/chapters/resources.txt
+++ b/doc/specs/vulkan/chapters/resources.txt
@@ -1226,6 +1226,11 @@ include::../api/enums/VkImageUsageFlagBits.txt[]
     to create a sname:VkImageView suitable for occupying a
     sname:VkDescriptorSet slot of type
     ename:VK_DESCRIPTOR_TYPE_STORAGE_IMAGE.
+    If this bit is set for any elink:VkFormat which is multi-planar, then
+    any created sname:VkImageView of a single-plane of the image may be
+    used to create a sname:VkDescriptorSet slot of type
+    ename:VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, however the multi-planar
+    itself elink:VkFormat may not be used to create sname:VkImageView.
   * ename:VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT specifies that the image can:
     be used to create a sname:VkImageView suitable for use as a color or
     resolve attachment in a sname:VkFramebuffer.


drakos-amd:
The proposed change doesn't seem to be necessary.

As part of the effort done for `VK_KHR_maintenance2` (part of Vulkan 1.1) we clarified the relation of the image usage flags and format specified at image creation time and the image usage flags and format specified for individual image views.

For images created without the `VK_IMAGE_CREATE_EXTENDED_USAGE_BIT` all the image usage flags specified at image creation time must be supported by the format specified for the image.

For images created with the `VK_IMAGE_CREATE_EXTENDED_USAGE_BIT` the image usage flags specified at image creation time only specify which usages may be used by any view (which may have a different format than that of the image's base format), and individual views can be created with a subset of those image usage flags (specified in `VkImageViewUsageCreateInfo`). Thus validating whether an image usage is supported for a particular format happens at image view creation time (see the valid usage section of `VkImageViewCreateInfo`).

What this means in particular for multi-planar images is that implementations are expected to report no support for `VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT` for the multi-planar format, but can report support for it for the individual per-plane view formats.

Thus if applications intend to use individual plane views of a multi-planar image as a storage image then they must create the image with `VK_IMAGE_CREATE_EXTENDED_USAGE_BIT` and `VK_IMAGE_USAGE_STORAGE_BIT`. At image view creation time the usage specified in `VkImageViewUsageCreateInfo` can thus contain `VK_IMAGE_USAGE_STORAGE_BIT` for any per-plane view whose format does support `VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT`, but should exclude `VK_IMAGE_USAGE_STORAGE_BIT` for a multi-plane view.

Please note though that this rule applies in general to all images that can be used to create image views with different formats that potentially support a larger (or simply different) set of format features than the parent image's format, and applies to any usage combination, thus is not limited to multi-planar images and storage image use.

For more details please look at the appropriate places of the spec discussing `VK_IMAGE_CREATE_EXTENDED_USAGE_BIT` and the related valid usage clauses of `VkImageViewCreateInfo`.

I'd like to thank you for raising this PR nevertheless, as it did reveal an actual issue in the current version of the spec, as (most probably due to a merging issue) the valid usage section of `VkImageCreateInfo` (not `VkImageViewCreateInfo`) still contains the old valid usage clauses on the format feature requirements in relation to image usage flags, even though those no longer apply. We will fix that in a future spec update.

atomnuker:
Thanks for replying, though I already figured out to use the EXTENDED usage creation flag through some discussion with the implementors. I just forgot to close this PR, so sorry about that.

PR #684, opened 2018-03-17
Correct mem VU ifdef scope in VkBindImageMemoryInfo 

krOoze:
- previously VU exluded if `1.1` or `VK_KHR_device_group`, but if swapchain ext. not defined memory param should be valid too
=> now excluded iff `VK_KHR_swapchain` is defined too.

`(1.1 or VK_KHR_device_group) and VK_KHR_swapchain` should show swapchain specific VU (which provides memory through `pNext` instead). Negative of that should show the core VU ("`memory` must be valid").

### DIFF
--- a/chapters/resources.txt
+++ b/chapters/resources.txt
@@ -3918,10 +3918,12 @@ ifndef::VK_VERSION_1_1,VK_KHR_dedicated_allocation[]
     dedicated for a specific buffer or image
 endif::VK_VERSION_1_1,VK_KHR_dedicated_allocation[]
 endif::VK_NV_dedicated_allocation[]
-ifndef::VK_VERSION_1_1,VK_KHR_device_group[]
+ifndef::VK_VERSION_1_1+VK_KHR_swapchain[]
+ifndef::VK_KHR_device_group+VK_KHR_swapchain[]
   * [[VUID-VkBindImageMemoryInfo-memory-01625]]
     pname:memory must: be a valid dname:VkDeviceMemory handle
-endif::VK_VERSION_1_1,VK_KHR_device_group[]
+endif::VK_KHR_device_group+VK_KHR_swapchain[]
+endif::VK_VERSION_1_1+VK_KHR_swapchain[]
 ifdef::VK_VERSION_1_1,VK_KHR_device_group[]
   * [[VUID-VkBindImageMemoryInfo-pNext-01626]]
     If the pname:pNext chain includes


doc/specs/vulkan/chapters/resources.txt
```diff
@@ -3958,6 +3958,12 @@ ifdef::VK_KHR_swapchain[]
     slink:VkBindImageMemorySwapchainInfoKHR, pname:memory must: be a valid
     dname:VkDeviceMemory handle
 endif::VK_KHR_swapchain[]
+ifndef::VK_KHR_swapchain[]
+  * pname:memory must: be a valid dname:VkDeviceMemory handle
+endif::VK_KHR_swapchain[]
+endif::VK_VERSION_1_1,VK_KHR_device_group[]
+ifndef::VK_VERSION_1_1,VK_KHR_device_group[]
```
oddhack: I believe the fallthrough VU at line 3966 replicates VUID-VkBindImageMemoryInfo-memory-01625 above.

krOoze: Yes, the VU aplies to two conditions:
- `VK_KHR_device_group && !VK_KHR_swapchain`
- `!VK_KHR_device_group`

Such condition cannot be implemented using current (documented) `ifdef` as a single VU.

krOoze: Alright, removed all duplicates.

drakos-amd:
Can you please clarify the motivation for this PR?

krOoze:
@drakos-amd  Certainly.

The 

> `memory` **must** be a valid `VkDeviceMemory` handle

VU was missing for the case of `(VK_VERSION_1_1 || VK_KHR_device_group) && !VK_KHR_swapchain`.

I apologize for the macro-fu, but I am not sure there is a better way in asciidocotor to represent that predicate as a single VU statement.

doc/specs/vulkan/chapters/resources.txt
```diff
@@ -3914,10 +3914,11 @@ ifndef::VK_VERSION_1_1,VK_KHR_dedicated_allocation[]
     dedicated for a specific buffer or image
 endif::VK_VERSION_1_1,VK_KHR_dedicated_allocation[]
 endif::VK_NV_dedicated_allocation[]
-ifndef::VK_VERSION_1_1,VK_KHR_device_group[]
-  * [[VUID-VkBindImageMemoryInfo-memory-01625]]
-    pname:memory must: be a valid dname:VkDeviceMemory handle
-endif::VK_VERSION_1_1,VK_KHR_device_group[]
+ifndef::VK_VERSION_1_1+VK_KHR_swapchain[]
+ifndef::VK_KHR_device_group+VK_KHR_swapchain[]
+  * pname:memory must: be a valid dname:VkDeviceMemory handle
```
cubanismo: I'm not entirely sure dropping the VUID is appropriate here.  We're not creating a new VU statement or even modifying an existing VU statement per-say.  We're just fixing the cases in which an existing VU is exposed.

krOoze: I wouldn't know.

The styleguide is bit vague on this case.
And I am not sure how the scripts are handling it.
Change would break layers, but at the same time alert that a change in code is necessary to catch this.

Your call; change, or keep?

mikes-lunarg: You can keep the VUID, we will be able to detect it shortly in the layers

cubanismo:
Thanks, that looks better.

ianelliottus:
@oddhack the SI TSG approved this pull request during its last meeting.

PR #685, opened 2018-03-19
Remove duplicate present layout note

krOoze:
Remove duplicate present layout note.

The identical thing is already below `VkPresentInfoKHR` where it belongs.

### DIFF
--- a/doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
+++ b/doc/specs/vulkan/chapters/VK_KHR_swapchain/wsi.txt
@@ -810,29 +810,6 @@ react appropriately.
 If presentation fails because of a mismatch in the surface and presented
 image sizes, a ename:VK_ERROR_OUT_OF_DATE_KHR error will be returned.
 
-Before an application can: present an image, the image's layout must: be
-transitioned to the ename:VK_IMAGE_LAYOUT_PRESENT_SRC_KHR
-ifdef::VK_KHR_shared_presentable_image[]
-layout, or for a shared presentable image the
-ename:VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR
-endif::VK_KHR_shared_presentable_image[]
-layout.
-
-.Note
-[NOTE]
-====
-When transitioning the image to
-ifdef::VK_KHR_shared_presentable_image[]
-ename:VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR or
-endif::VK_KHR_shared_presentable_image[]
-ename:VK_IMAGE_LAYOUT_PRESENT_SRC_KHR, there is no need to delay subsequent
-processing, or perform any visibility operations (as flink:vkQueuePresentKHR
-performs automatic visibility operations).
-To achieve this, the pname:dstAccessMask member of the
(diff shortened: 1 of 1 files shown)


krOoze:
@ianelliottus The note seems to be relevant for **present** op, not **acquire**.

ianelliottus:
Correct.

In looking at this again, it seems that ANI needs a sentence that talks about transitioning the layout that's different than what's there now.  Would you mind writing that (i.e. rather than dropping this whole duplicate, just transform it into what's needed here)?

krOoze:
@ianelliottus I started that way, but ended up just deleting it. Such note just reiterated what was actually already written there above it in another note.

What remains is the "must be transitioned from PRESENT layout". Which seemed better left to pre-existing VUs which will enforce that.
Wording it out here also proved more confusing than helpful. Saying simply "it must be transitioned from" is a lie (especially for `SHARED_PRESENT`). And saying the accurate truth is gets very convoluted.

ianelliottus:
Sigh.  This whole thing has grown complex.  I'm not sure I follow what pre-existing text you are referring to.  I started to include it all here, to see if we are referencing the same things.  However, I have a suggestion--to modify the following paragraph (copied from the latest 1.1 spec):

```
When successful, vkAcquireNextImageKHR acquires a presentable image from swapchain that an application can use, and sets pImageIndex to the index of that image within the swapchain. The presentation engine may not have finished reading from the image at the time it is acquired, so the application must use semaphore and/or fence to ensure that the image layout and contents are not modified until the presentation engine reads have completed. The order in which images are acquired is implementation-dependent, and may be different than the order the images were presented.
```

You are correct that it can be convuluted.  If the image is already in VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR, it doesn't need to be transitioned.  However, if it's not, it will need to be transitioned.  How about adding a sentence like the following?:

> The image’s layout must: be appropriate for the intended use, which won't require a transition if the layout is already VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR, but otherwise will.

I would add the sentence after the 1st sentence in the paragraph I quoted.  Thoughts?

krOoze:
I refer to the:

> An application must wait until either the semaphore or fence is signaled before accessing the image’s data.
>[Note]
When the presentable image will be accessed by some stage S, the recommended idiom for ensuring correct synchronization is:
blah blah

> The image’s layout must: be appropriate for the intended use, which won't require a transition if the layout is already VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR, but otherwise will.

That is a lie, because even `VK_IMAGE_LAYOUT_PRESENT_KHR` can get destroyed without any transition, or even re-presented (not useful, but valid).

On the other hand "The image’s layout must be appropriate for the intended use" sounds like a tautology. And I am not quite sure it is useful to say it there.

ianelliottus:
BTW, I'll recommend that the SI group accept this pull request, at an upcoming meeting.

krOoze:
> Backing up to my goals ... my target audience is somebody who is new to Vulkan

That is not traditionally the target audience of a specification. Though it is true that the Vulkan specification is quite readable and useful to novice as far as specifications go.

ianelliottus:
@oddhack the SI TSG approved this to be merged on Wednesday.

Issue #512, opened 2017-06-07
Document VK_IMG_format_pvrtc


lexaknyazev:
- Link from [Khronos Vulkan Registry](https://www.khronos.org/registry/vulkan/#repo-docs) points to nowhere: `https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VK_IMG_format_pvrtc`

- `VkFormat` enum lists 8 PVRTC entries:
  ```
  VK_FORMAT_PVRTC1_2BPP_UNORM_BLOCK_IMG
  VK_FORMAT_PVRTC1_4BPP_UNORM_BLOCK_IMG
  VK_FORMAT_PVRTC2_2BPP_UNORM_BLOCK_IMG
  VK_FORMAT_PVRTC2_4BPP_UNORM_BLOCK_IMG
  VK_FORMAT_PVRTC1_2BPP_SRGB_BLOCK_IMG
  VK_FORMAT_PVRTC1_4BPP_SRGB_BLOCK_IMG
  VK_FORMAT_PVRTC2_2BPP_SRGB_BLOCK_IMG
  VK_FORMAT_PVRTC2_4BPP_SRGB_BLOCK_IMG
  ```
  But there're 10 values for OpenGL ES:
  - **[IMG_texture_cmpression_pvrtc](https://www.khronos.org/registry/OpenGL/extensions/IMG/IMG_texture_compression_pvrtc.txt)**
    ```
    COMPRESSED_RGB_PVRTC_4BPPV1_IMG
    COMPRESSED_RGB_PVRTC_2BPPV1_IMG
    COMPRESSED_RGBA_PVRTC_4BPPV1_IMG
    COMPRESSED_RGBA_PVRTC_2BPPV1_IMG
    ```
  - **[EXT_pvrtc_sRGB](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_pvrtc_sRGB.txt)**
    ```
    COMPRESSED_SRGB_PVRTC_2BPPV1_EXT
    COMPRESSED_SRGB_PVRTC_4BPPV1_EXT
    COMPRESSED_SRGB_ALPHA_PVRTC_2BPPV1_EXT
    COMPRESSED_SRGB_ALPHA_PVRTC_4BPPV1_EXT
    ```
[...]

PR #690, opened 2018-03-24
Organize extension appendices

krOoze:
- hide from toc
- split into "current" extensions, and extensions that have since been 
deprecated by a better alternative
- make core promotion list generated
- add information to extension appendix about possible better
alternatives
- split contacts to separate lines
- add placeholder for missing `VK_IMG_format_pvrtc` appendix
- some general-purpose quality improvements to the script


xml/registry.rnc
```diff
@@ -419,6 +419,9 @@ Extensions = element extensions {
 #       extension
 #   supported - profile name(s) supporting this extension, e.g. 'vulkan'
 #       or 'disabled' to never generate output.
+#   precedes - Vulkan version or name of extension that succeeded this
```
Tobski: I think it'd make more sense to tag the "promotion level" with the actual tag name, rather than relying on string manipulation outside of the xml structure. So something like "promotedby=" for things that are fully promoted, and then other terms can be introduceds, something like "succeededby" (or some less generic word) for things that are actively not working any more and have an alternative (e.g. AMD_negative_viewport_height) .

Tobski: Also I'm not sure if "full promotion" is meant to be "this completely overtakes all of the functionality of this extension" or just "it's in a major version now"?

I thought the former, but now re-reading it might be the latter... otherwise it seems 'promoted' should be on a lot more things, and I'm not sure why you even have it... is it just semantic tagging to say "this is a version"?

krOoze: Right, I didn't want to explode the attribute count. Basically for no text manipulation I would need four, something: `promotedtoext`, `promotedtovk`, `obsoletedbyext`, and `obsoletedbyvk`. Which also complicates the rules (you can use only one of those attributes on a tag).

BTW, how should you prefer I write multi-word attribute names? Just all lowercase as above?

<hr />

So "promotion" I define:

If extension **A** is promoted to extension or Vulkan version **B**, that means **B** has all the interfaces (commands, structures, enums, etc) **A** has with only name change (e.g. removing KHR, or EXT->KHR suffix change). Changes to behavior of **B**  must be backwards-compatible with **A** (i.e. App that switches from **A to **B**, and only S&R the names must work). Changes to enabling mechanism is tolerated (e.g. from `ppEnabledExtensions` to `pEnabledFeatures`).

So basically what happened with some extensions and Vulkan 1.1 +- the last part, and that it can happen between extensions (e.g. a promotion from EXT to KHR extension).

The generic "succession" would be:

If extension **A** is suceeded by extension or Vulkan version **B**, that means an App can achieve the same results or use case enabled by **A** by using **B** instead.
E.g. `debug_report` and `debug_utils` serve the same purpose, but do not share interfaces. `debug_utils` is just supposed to be newer (and better) than `debug_report`, so apps should prefer to use`debug_utils` (if they have a choice).

krOoze: Hopefully correctly rebased.

Ok, I see what the confusion is. I marked all the extensions as "succession" only. I was probably lazy to check which are full-fledged promotion, and never got back to it. The script should be ready for that case though...

Tobski: > Right, I didn't want to explode the attribute count. Basically for no text manipulation I would need four, something like: promotedtoext, promotedtovk, obsoletedbyext, and obsoletedbyvk. Which also complicates the rules (you can use only one of those attributes on a tag).

I think just "promotedby" and "obsoletedby" would be sufficient - then just detect whether the value is a version or an extension name. That seems like it'd be more sane IMO?

> BTW, how should you prefer I write multi-word attribute names? Just all lowercase as above?

Yea I think that's consistent.

> So, "promotion" I define:

> If extension A is promoted to extension or Vulkan version B, that means B has all the interfaces (commands, structures, enums, etc) A has with only name change (e.g. removing KHR, or EXT->KHR suffix change). Changes to behavior of B must be backwards-compatible with A (i.e. App that switches from A to B, and only S&R the names must work). Changes to enabling mechanism is tolerated (e.g. from ppEnabledExtensions to pEnabledFeatures).

That's mostly consistent with what we've defined internally, though we're also including renames of APIs and other compatible tweaks (e.g. adding additional parameters).

> The more generic "succession" would be:

> If extension A is suceeded by extension or Vulkan version B, that means an App can achieve the same results or use case enabled by A by using B instead.

I had to double check, but we've used the term _deprecated_ for this in the spec (see the AMD_negative_viewport_height appendix). So "deprecatedby" would be appropriate.

> E.g. debug_report and debug_utils serve the same purpose, but do not share interfaces. debug_utils is just supposed to be newer (and better) than debug_report, so apps should prefer to usedebug_utils (if they have a choice).

As per my comment above, I definitely think we'd consider this promotion. We should definitely clarify this in the spec; post 1.1 we have an aliasing mechanism for "identical" APIs with different names/suffixes, but haven't used this for debug utils since that came before the mechanism - we should probably retroactively do that actually... I'll look at raising a separate issue for that though.

> Grey area things like AMD_negative_viewport_height could also belong here. The spec claims "Due to some slight behavioral differences", so it cannot be a "promotion".

It's deprecated, as per the spec's current language.

krOoze: OK, I shall update the language.

ad `debug_report` and `debug_utils`:
AIS above, I did not check the extensions thoroughly, if they fit the "promotion" category, but I think it is outside it because of `VkDebugReportObjectTypeEXT` and `PFN_vkDebugUtilsMessengerCallbackEXT` being different and such.

As long as I am adding things to the XML, I would like at least one of the new attributes to not be too fuzzy, and so potentially usable by various scripts. Having two fuzzy attributes with several undocumented exceptions would defeat the purpose, and we could just have one. The "promoted" should be reserved for cases of almost verbatim duplicates (as was done with extensions assimilated by 1.1).

oddhack:
One reason I'm afraid of wiring in this sort of knowledge is that although the term "deprecated" is used in the spec in a handful of places, the actual meaning of "deprecated" is not well-established, and we have something like half a dozen related internal issues about what it means to be "deprecated" and have a "deprecation model" and how that will affect other ecosystem components like layers. In the absence of very specific definitions of such terms, I don't want to try and force a definition in the XML that might not even be true in the future.

Tobski:
@oddhack I'm fairly sure we committed to a definition of deprecated, which is why it's in the spec. It just... didn't get recorded :|

Tobski:
I'll raise an comment on the "what is deprecation" issue and see if we can get some consensus on what I believe we agreed the definition should be.

Tobski:
Ok so something DID get recorded, but not actually actioned, nor any record made of consensus:

>  "Deprecated: something that has been subsumed by a better mechanism, but it's still usable for now (we don't really have good examples for this in Vulkan)" - but it's not been recorded whether we agreed to that  yet.

AMD_negative_viewport_height is actually considered "obsoleted" by that same bug, because it's actively unusable moving forward, in addition to being no longer the "modern" way to do specify it.

I believe we agreed to these terms in theory, but never recorded it - I'll ping on the issue, and we should bring it up at next Monday's meeting. Ideally we can then move forward with an actual definition, perhaps added to the style guide, and we can then update this issue too.

@krOoze For now, what I've suggested in my earlier response seems like a good place to start, but change AMD_negative_viewport_height to "obsoletedby". I suspect this will stick.

krOoze:
I don't need formal definition for this particular attribute. The only actionable result I can imagine of a `succededby`\\`deprecatedby` attribute is to give an user a nice info, that use of another alternative extension\vkversion is encouraged. So any informal definition, or the usual semantics of this word is just fine -- probably even good, as the attribute could be used for various reasons and cases.

@Tobski Alright, so `obsoletedby` for now...

Tobski:
Promotion already has a somewhat rigid definition, but it only covers promotion to core. As for the other two, I think the concern is using a misleading word and then having to change it to match the spec in future so things aren't completely confusing - so we need to get our ducks in a row before we commit to anything. Regardless, I've raised a change internally to sort out the definitions.

krOoze:
updated with the sugestions

Tobski:
I want to get sign off on the definitions of obsoleted/promoted/deprecated before we accept this, but assuming that doesn't get enormously derailed, I think this looks great.

MarkY-LunarG:
From an OpenXR standpoint, this functionality looks nice.  Please let me know when this goes in and I'll try to incorporate it into our tree.

Tobski:
Hi @krOoze, we've decided on the following definitions in the specification, and it's now just pending a new spec release:

> Deprecated::
    A feature is deprecated if it is no longer recommended as the correct or
    best way to achieve its intended purpose.
    Generally a newer feature will have been created that solves the same
    problem - in cases where no newer alternative feature exists,
    justification should be provided.
>
> Obsoleted::
    A feature is obsolete if it can no longer be used.
    For core features, making one obsolete would be in violation of the
    <<fundamentals-versionnum, compatibility rules>>, so must not be done.
    However extensions do not have these guarantees, and can be made
    obsolete by a newer core version or extension.
>
> Promoted::
    A feature is promoted if it is taken from an older extension and made
    available as part of a new <<versions,core version of the API>>, or a
    newer extension that is considered to be either as widely supported or
    more so.
    A promoted feature may have minor differences from the original such as:
      * It may be renamed
      * A small number of non-intrusive parameters may have been added
      * The feature may be advertised differently by
        <<features-features,device features>>
      * The author ID suffixes will be changed or removed as appropriate

I'll go through and review your tagging accordingly so we know where we stand.

One thing we want in addition to this is the option to obsolete or deprecate a feature *without* a new feature taking over the responsibility - there's at least one extension we're considering making obsolete, and we'd like to use this framework to achieve that. Maybe something as simple as having the string for "obsoletedby" allowed to be empty would suffice, but this will need handling properly in the scripts. Would you be willing to make that change?

Tobski:
> it needs some trivial changes: s\obsoleted\deprecated, and define obsoletedby (of which VK_AMD_negative_viewport is AFAIK the only instance).

@krOoze I got distracted halfway through reviewing, but yea that sounds right :)

> It would be nice to link to the impending Glossary terms. Do you plan to merge this now, or after the next spec release?

It'll be next spec release - we don't generally do out of band updates. There's no dedicated links for these though, so you could just link to the glossary.

> But assuming such feature was not broken by design it does not particularly make sense to offer no replacement at all. Any chance sharing which extension that is, to see what you mean?

It's not really my place to share I'm afraid - but I can assure you there are good reasons!

> However extensions do not have these guarantees, and can be made
obsolete by a newer core version or extension.

Yes I had noticed that. It doesn't say it *needs* a new core version or extension though 😛 (Yes this is only a semi-serious observation!) We'll probably either change it, or maybe consider adding a "revoked" or something instead, when we actually move to do it - we're sort of waiting on this PR to land before committing to a specific path forward; this PR adds useful framework we can re-use however we end up doing it.

> Newer version\extension is practically always less widely supported at release.

This is the least awkward way I could say it, but this means support in the "I support this extension being made" rather than "actually support it in the driver". So vendor->EXT, or an EXT to another EXT with more buy-in, or EXT->KHR or vendor->core... that sort of thing.

> I am also not sure what a "non-intrusive parameter" is.
Can it at least be guaranteed that zero-initializing that parameter leads to the original behavior of the extension being promoted?

The glossary is non-normative, so expect informal language to a degree. The reason no such guarantee has been included is because this is kind of intended to be a case-by-case decision. There are already a few cases where this isn't entirely true, and for good reasons - but we always document the "default" value for the legacy version of the functionality. We wouldn't expect to call something promoted if there's no possible default in the new functionality.

appendices/VK_IMG_format_pvrtc.txt
```diff
@@ -0,0 +1,13 @@
+include::meta/VK_IMG_format_pvrtc.txt[]
+
+=== New Enum Constants
+
+  * Extending elink:VkResult:
```
Tobski: This should be elink:VkFormat

krOoze: indeed.

PS: this stub appendix relates to #512. Should I add "IP Status: unresolved" or such?

Tobski: Ehhh... not for me to say. Is this change critical to any of the script changes you've made? If not, might be better to take it out.

xml/vk.xml
```diff
@@ -6828,7 +6828,7 @@ server.
                 <command name="vkQueueSignalReleaseImageANDROID"/>
             </require>
         </extension>
-        <extension name="VK_EXT_debug_report" number="12" type="instance" author="GOOGLE" contact="Courtney Goeltzenleuchter @courtney-g" supported="vulkan">
+        <extension name="VK_EXT_debug_report" number="12" type="instance" author="GOOGLE" contact="Courtney Goeltzenleuchter @courtney-g" supported="vulkan" obsoletedby="VK_EXT_debug_utils">
```
Tobski: We should change this one to "Promoted By" according to the new definitions.

xml/vk.xml
```diff
@@ -6916,7 +6916,7 @@ server.
                 <enum value="&quot;VK_AMD_shader_explicit_vertex_parameter&quot;" name="VK_AMD_SHADER_EXPLICIT_VERTEX_PARAMETER_EXTENSION_NAME"/>
             </require>
         </extension>
-        <extension name="VK_EXT_debug_marker" number="23" type="device" requires="VK_EXT_debug_report" author="Baldur Karlsson" contact="Baldur Karlsson @baldurk" supported="vulkan">
+        <extension name="VK_EXT_debug_marker" number="23" type="device" requires="VK_EXT_debug_report" author="Baldur Karlsson" contact="Baldur Karlsson @baldurk" supported="vulkan" obsoletedby="VK_EXT_debug_utils">
```
Tobski: This one is definitely a promotion.

appendices/VK_IMG_format_pvrtc.txt
krOoze: Nah, just sounds worse **not** to include the IP status.

The stub is important for the script though. It assumes appendix for each extension exists. Would otherwise need to add exception mechanism.

xml/vk.xml
Tobski: (I know there's some weirdness with "number of object handles returned" but given the old functionality is definitely a subset, it seems fine to say promotion - i.e. you could use this in the same way if you ignore extra objects)

krOoze: That is stretching the definition a lot.
- flags are split in two types
- `vkDebugReportMessageEXT` and `vkSubmitDebugUtilsMessageEXT` seems completely different
- `PFN_vkDebugReportCallbackEXT` and `PFN_vkDebugUtilsMessengerCallbackEXT` seem completelly different too

At the same time I don't see a harm as having it as `deprecatedby`. Promotion is just a more restricted type of deprecation.

appendices/extensions.txt
```diff
@@ -58,528 +58,26 @@ not be equivalent to `KHX`, but it is likely to indicate a lesser degree of
...
+include::meta/relevant_extension_appendices.txt[]
+
+
+ifdef::HAS_OBSOLETE_EXTENSIONS[]
+[[succeeded-extension-appendices-list]]
+== List of Promoted and Obsolete Extensions
+include::meta/obsolete_extension_appendices_toc.txt[]
+<<<
+include::meta/obsolete_extension_appendices.txt[]
+endif::HAS_OBSOLETE_EXTENSIONS[]
```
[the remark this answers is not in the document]
krOoze: That's problematic, because I have no way of knowing at the time of the script running, which extensions are included and which vk version is being compiled. HAS_OBSOLETE_EXTENSIONS variable is computed at `asciidoctor` run time to do the magic.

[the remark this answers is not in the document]
krOoze: PS: I can hide the magic and the titles into the metafiles though.

Does it make sense to list deprecated and promoted extensions separately? When I decided to list them together, I thought listing them alphabetically together is more useful to the reader than two lists.

xml/vk.xml
Tobski: Yea I double checked what's changed and you're probably right. Let's call it deprecated.

krOoze:
Updated.

Tobski:
FYI, the definitions have now been pushed out to the latest spec:

https://github.com/KhronosGroup/Vulkan-Docs/blob/ff9357a4bad5fa49b99b9d3f69d254e2a3f0f575/appendices/glossary.txt#L233

https://github.com/KhronosGroup/Vulkan-Docs/blob/ff9357a4bad5fa49b99b9d3f69d254e2a3f0f575/appendices/glossary.txt#L814

https://github.com/KhronosGroup/Vulkan-Docs/blob/ff9357a4bad5fa49b99b9d3f69d254e2a3f0f575/appendices/glossary.txt#L948

oddhack:
I'm going to accept this, then make a few tweaks:

* We shouldn't publish even a partial VK_IMG_format_pvrtc appendix. IMG is working through their internal process to be able to include the spec in this repository, and anticipating them in this fashion won't accomplish anything good. Unfortunately just removing the appendix 
* The registry.txt updates weren't accompanied by a changelog comment.

oddhack:
Done. Turns out I had to recreate an empty pvrtc appendix file to avoid an asciidoc error showing up in the output. Also there's a dead link from the extensions list, which is unavoidable without further tagging of the XML which is really not worth it for this weird one-off case  that will eventually be resolved.

oddhack:
@MarkY-LunarG this functionality is in vulkan at present, if you want to incorporate it in OpenXR tools per previous comment (n.b. we should probably discuss that a bit, since I think there have been a number of other schema and script changes in the vulkan side since you branched off - would be nice to sync back up again, if possible).

PR #691, opened 2018-03-26
Fix typo: Change 'FENCE_FD' to 'SYNC_FD'

ascent12:
When this enum was renamed, this must have been overlooked.


drakos-amd:
The change looks fine.

PR #659, opened 2018-02-13
Require instance extension dependencies on vkCreateDevice

krOoze:
- add VU requiring `VkDeviceCreateInfo::ppEnabledExtensionNames` to have its instance extension dependencies met

PR #693, opened 2018-03-26
Refine extension dependencies

krOoze:
- requires impls to honor resolution in #659
- addresses some confusion in https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/2366
- require impls to support dependencies

### DIFF
--- a/chapters/extensions.txt
+++ b/chapters/extensions.txt
@@ -223,6 +223,15 @@ wrapped by the validation layers.
 See the <<LoaderAndValidationLayers, "`Vulkan Loader Specification and
 Architecture Overview`">> document for additional information.
 
+[NOTE]
+.Note
+====
+Valid Usage sections for individual commands and structures do not currently
+contain which extensions have to be enabled in order to make their use
+valid, although it might do so in the future.
+It is defined only in the <<fundamentals-validusage-extensions>> section.
+====
+
 [open,refpage='vkEnumerateInstanceExtensionProperties',desc='Returns up to requested number of global extension properties',type='protos']
 --
 
@@ -344,21 +353,22 @@ endif::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
 == Extension Dependencies
 
 Some extensions are dependent on other extensions to function.
-To use extensions with dependencies, such _required extensions_ must: also
-be enabled through the same API mechanisms when creating an instance with
-flink:vkCreateInstance or a device with flink:vkCreateDevice.
+To enable extensions with dependencies, such _required extensions_ must:
+also be enabled through the same API mechanisms when creating an instance
+with flink:vkCreateInstance or a device with flink:vkCreateDevice.
 Each extension which has such dependencies documents them in the
 <<extensions, appendix summarizing that extension>>.
 
-[NOTE]
-.Note
-====
-The Specification does not currently include required extensions in Valid
-Usage statements for individual commands and structures, although it might
-do so in the future.
-Nonetheless, applications must: not use any extension functionality if
-dependencies of that extension are not enabled.
-====
+If an extension is supported (as queried by
+flink:vkEnumerateInstanceExtensionProperties or
+flink:vkEnumerateDeviceExtensionProperties), then _required extensions_ of
+that extension must: also be supported for the same instance or
+physical device. 
+
+Any device extension that has an instance extension dependency that is not
+enabled by flink:vkCreateInstance is considered to be unsupported, hence it
+must: not be returned by flink:vkEnumerateDeviceExtensionProperties for any
+slink:VkPhysicalDevice child of the instance.
 
 
 [[extended-functionality-extensions-compatibility]]
--- a/appendices/glossary.txt
+++ b/appendices/glossary.txt
@@ -1021,7 +1021,7 @@ Render Pass Instance::
     A use of a render pass in a command buffer.
 
 Required Extensions::
-    Extensions which must: be enabled to use a specific enabled extension
+    Extensions that must: be enabled alongside extensions dependent on them
     (see <<extended-functionality-extensions-dependencies, Extension
     Dependencies>>).
 


doc/specs/vulkan/chapters/extensions.txt
```diff
@@ -344,21 +354,16 @@ endif::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
...
 
-[NOTE]
-.Note
-====
-The Specification does not currently include required extensions in Valid
-Usage statements for individual commands and structures, although it might
-do so in the future.
-Nonetheless, applications must: not use any extension functionality if
-dependencies of that extension are not enabled.
-====
```
drakos-amd: Not sure why you moved this note, I think its current location makes more sense.

Otherwise the proposed changes seem fine to me.

krOoze: I sorta repurposed it. It no longer describes extension dependencies (which are undefined behavior per VUs), so it no longer belongs in the Extension Dependencies chapter.

I made it instead to link to the Extension Implicit VU chapter.

doc/specs/vulkan/appendices/glossary.txt
```diff
@@ -1021,8 +1021,9 @@ Render Pass Instance::
     A use of a render pass in a command buffer.
 
 Required Extensions::
-    Extensions which must: be enabled to use a specific enabled extension
-    (see <<extended-functionality-extensions-dependencies, Extension
+    Extensions that must: be enabled to be able to enable specific
+    extension dependent on them (see
```
ianelliottus: I think it will read more clearly if you change this sentence to something like:

> Extensions that must: be enabled in order to enable another extension that is dependent on them (see ...

doc/specs/vulkan/chapters/extensions.txt
```diff
@@ -223,6 +223,16 @@ wrapped by the validation layers.
 See the <<LoaderAndValidationLayers, "`Vulkan Loader Specification and
 Architecture Overview`">> document for additional information.
 
+[NOTE]
+.Note
+====
+The Specification does not currently include necessary extensions in Valid
```
ianelliottus: Not sure why you changed "required extensions" (defined in the glossary) to "necessary extensions" (which is not used elsewhere in the spec)..

doc/specs/vulkan/chapters/extensions.txt
```diff
@@ -223,6 +223,16 @@ wrapped by the validation layers.
...
+Usage statements for individual commands and structures, although it might
+do so in the future.
+It is instead defined in the <<For fundamentals-validusage-extensions,
+Valid Usage for Extensions>> section.
```
ianelliottus: The source for the spec is intended for use in building both 1.0 and 1.1 specifications.  What you are referring to is only available in the 1.1 spec, but you didn't use an ifdef.

doc/specs/vulkan/chapters/extensions.txt
```diff
@@ -223,6 +223,16 @@ wrapped by the validation layers.
 See the <<LoaderAndValidationLayers, "`Vulkan Loader Specification and
 Architecture Overview`">> document for additional information.
 
+[NOTE]
```
ianelliottus: I'm not sure why you are moving this note up sooner, and out of the Extension Dependencies section.

ianelliottus: I see you and @drakos-amd discussed this below.  I'll comment further below ....

doc/specs/vulkan/chapters/extensions.txt
ianelliottus: I've looked at your new language a lot, and perhaps I'm starting to see what your intention is, and how you're re-purposing the language.  It's using a non-defined, unique term ("necessary extensions"), that didn't convey to me what I now think you're trying to do.  I'll suggest some language to help convey what I think you want.

In addition, I want you to keep the first sentence of the old note here.  The 2nd sentence is repetitive with the paragraph above it; and is normative, which means it doesn't belong in a note.  Still, keep the first sentence (it serves a purpose internal to Khronos).

doc/specs/vulkan/chapters/extensions.txt
krOoze: Mostly because I do not like to throw keywords around. "**Required** extension" to me sounds like the implementation **must** support it. You are right though it is not changed elsewhere (Glossary).

doc/specs/vulkan/appendices/glossary.txt
krOoze: Both sound horrible to me...  :/

doc/specs/vulkan/chapters/extensions.txt
krOoze: The first sentence does not really make sense to me here.
It is already undefined behavior on `vkCreateInstance\Device`. So those commands are already undefined behavior, if you use null or invalide instace\device\child-of-those. It would make no sense to ever add Required Extension VUs to those, because we are already in undefined behavior teritory.

doc/specs/vulkan/appendices/glossary.txt
cubanismo: Yes, all the forms of this language are pretty awkward.  I don't really like either update suggestion, nor the original.  Perhaps it warrants a more thorough rewording?

doc/specs/vulkan/chapters/extensions.txt
ianelliottus: Oh, we fixed that.  Originally (when I wrote those) it was just in the 1.1 spec.  I thought that we had published a 1.0 spec (which I was looking at) based on the common branch, but I guess we haven't.  Sorry!

doc/specs/vulkan/appendices/glossary.txt
ianelliottus: Hmm.  Looking at this discussion, and below, I agree with James.  The "required extensions" term seems wrong, because it's defined narrowly here, and yet a looser-defined version of the term gets thrown around internally (e.g. the goal of making it clear which core version and/or extension is required in order to use a command, struct, etc.).  Bottom line (might not be saying it correctly), but I agree with James that we should step back and do a more thorough rewording.

doc/specs/vulkan/chapters/extensions.txt
oddhack: @ianelliottus the published 1.0.71 spec is built from master branch along with 1.1.71.

ianelliottus: @oddhack interesting.  I've been bringing up: https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html

That is still 1.0.69, and seems based on the old branch.  I just decided to navigate a bit, and found:

- 1.0.72 in: https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html
- 1.0.69 in: https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/html/vkspec.html
- 1.0.69 in: https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html
- 1.1.72 in: https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html
- 1.1.72 in: https://www.khronos.org/registry/vulkan/specs/1.1-khr-extensions/html/vkspec.html
- 1.1.72 in: https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html

For some issues/PRs/MRs, I compare the 1.0 and 1.1 specs.  Since I do so much with extensions, I always look at the version with all extensions (with the 1.0 spec being stuck:-).

krOoze:
Updated with some suggestions in mind
\+ newly added a paragraph requiring implementations to support the dependencies (aka "required extensions")

chapters/extensions.txt
```diff
@@ -223,6 +223,16 @@ wrapped by the validation layers.
 See the <<LoaderAndValidationLayers, "`Vulkan Loader Specification and
 Architecture Overview`">> document for additional information.
 
+[NOTE]
+.Note
+====
+Valid Usage sections for individual commands and structures do not currently
+contain which extensions have to be enabled in order to make their use
+valid, although it might do so in the future.
+It is defined only in the <<For fundamentals-validusage-extensions,
```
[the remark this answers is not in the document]
krOoze: weird... fixed

chapters/extensions.txt
```diff
@@ -223,6 +223,15 @@ wrapped by the validation layers.
 See the <<LoaderAndValidationLayers, "`Vulkan Loader Specification and
 Architecture Overview`">> document for additional information.
 
+[NOTE]
+.Note
+====
+Valid Usage sections for individual commands and structures do not currently
+contain which extensions have to be enabled in order to make their use
+valid, although it might do so in the future.
+It is defined only in the <<fundamentals-validusage-extensions>> section.
```
ianelliottus: Looks like you lost "Valid Usage for Extensions" in the link.

krOoze: That's intentional; it should automagically use the section name, which is what I want.

cubanismo:
New language looks OK to me.

ianelliottus:
LGTM too.  Will recommend merging.

PR #700, opened 2018-04-10
Remove C99/C++ style '//' comments

luckyxxl:
For a project I need to Vulkan headers with a C90 compiler. This works fine except for a few occurrences of C99/C++ style `//` comments. This PR replaces those with `/* ... */` comment blocks.


luckyxxl:
Ok, no problem!

Thank you very much for your consideration and discussing!

PR #702, opened 2018-04-14
Optimize page load time by hiding content during load

krOoze:
The specification html page load time have again gotten too long. It is a travesty that in 21st century with our buncha-core gazillion Hz CPUs it is in order of tens of seconds to load a page. So I once again look if there is some low hanging fruit to help the situation.

This hides the page until everything is ready, to prevent the browser from trying to do incremental display of partially loaded page. At least for me the browser is largely unresponsive during the load time anyway, so it is no great loss.

Informal measurements [s]:

Browser | Before | #702 | #704 | #708 | #702 + #704  | #702 +  #708 | #704 + #708 | #702 +  #704 + #708 
---------|-------- | ------|------  | ------| --------------- | ----------------| -------------- | ---------------------
Firefox *  | 28        | 12     |  9       |  28   | **6**               | 12                    |   7                  | **6**
Chrome| 30        | 21     |   23    |  23    | 18                  | 14                    |   12                | **10**

\* The Firefox measurements are invalid. I measured offline, and some security policy prevents loading fonts from superfolder in that case.

I also welcome some advice for aditional optimization from someone unlike me that knows and is friends with web technologies.

Some notes for future consideration:
- Math translation takes ~ 7 s &mdash; that can be eliminated by doing it offline (#704)
- The style and layout is calulated twice in Chrome &mdash; I have no idea why at the moment. It wastes ~8 s. (#708)
- I have a suspicion the TOC sidebar gives browsers some trouple. There may also be other tricky elements to layout (tables inside tables?)
- The binary images are inline in the html file. That is bound to create some performance problems...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ ATTRIBOPTS   = -a revnumber="$(SPECREVISION)" \
 ADOCEXTS     = -r $(CURDIR)/config/vulkan-macros.rb -r $(CURDIR)/config/tilde_open_block.rb
 ADOCOPTS     = -d book $(ATTRIBOPTS) $(NOTEOPTS) $(VERBOSE) $(ADOCEXTS)
 
-ADOCHTMLEXTS = -r $(CURDIR)/config/katex_replace.rb
+ADOCHTMLEXTS = -r $(CURDIR)/config/katex_replace.rb -r $(CURDIR)/config/loadable_html.rb
 
 # ADOCHTMLOPTS relies on the relative runtime path from the output HTML
 # file to the katex scripts being set with KATEXDIR. This is overridden
--- a/config/loadable_html.rb
+++ b/config/loadable_html.rb
@@ -1,0 +1,21 @@
+# Copyright (c) 2016-2018 The Khronos Group Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#require 'asciidoctor/extensions' unless RUBY_ENGINE == 'opal'
+RUBY_ENGINE == 'opal' ? (require 'loadable_html/extension') : (require_relative 'loadable_html/extension')
+
+# All the inline macros we need
+Asciidoctor::Extensions.register do
+    postprocessor MakeHtmlLoadable
+end
--- a/config/loadable_html/extension.rb
+++ b/config/loadable_html/extension.rb
@@ -1,0 +1,62 @@
+# Copyright (c) 2016-2018 The Khronos Group Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'asciidoctor/extensions' unless RUBY_ENGINE == 'opal'
+
+include ::Asciidoctor
+
+class MakeHtmlLoadable < Extensions::Postprocessor
+
+  def process document, output
+
+    if document.attr? 'stem'
+
+      loading_msg = '
+<div id="loading_msg"><p>Loading... please wait.</p></div>
+'
+      loadable_class = 'class="loadable"'
+
+      loaded_script = '
+<style>
+    #loading_msg {
+        width: 100%;
+        margin-left: auto;
+        margin-right: auto;
+        margin-top: 1ex;
+        margin-bottom: 1ex;
+        max-width: 62.5em;
+        position: relative;
+        padding-left: 1.5em;
+        padding-right: 1.5em;
+    }
+    .loadable {display: none !important;}
+</style>
+<script>
+    function unhideLoadableContent(){
+        document.getElementById("loading_msg").style.display = "none";
+        var loadables = document.getElementsByClassName("loadable");
+        for( var i = 0; i < loadables.length; ++i ) loadables[i].classList.remove("loadable");
+    }
+
+    window.addEventListener("load", unhideLoadableContent);
+</script>
+'
+
+      output.sub! /(?=<\/head>)/, loaded_script
+      output.sub! /(<div id="content")/, '\1' + " " + loadable_class + " "
+      output.sub! /(?=<div id="content")/, loading_msg
+    end
+    output
+  end
+end


oddhack:
Why should having the images imbedded cause problems?

krOoze:
@oddhack Because they need to be processed before the browser can continue constructing the HTML DOM and calculating layout. Theoretically the browser only needs to know the image dimensions to start calculating layout. So having whole image upfront may theoretically delay the start of the lengthy layout calculation.

krOoze:
Updated with better handling of hidden style. Restricted only to `content` (shows header + toc immediately). Style the loading message.

krOoze:
Fixed IE compatibility.

krOoze:
I set up [krOoze/Vulkan-Docs/optimize_AOTA](https://github.com/krOoze/Vulkan-Docs/tree/optimize_AOTA) branch with all of the PRs.

And I set up unofficial [hosting](http://vulkan.hys.cz/1.1-extensions/vkspec.html) (which I believe is allowed under the conditional copyright). Test your favorite browser. Do **not** link to it &mdash; it is only temporary.

kainino0x:
@oddhack Any chance landing some of these changes could get prioritized? I'm willing to help if there are any further changes needed to the PRs. There's also a chance I could ask some knowledgeable Chrome folks about optimizing this page further.

I also have a small, non-intrusive change that significantly improves the katex loading time, here:
https://github.com/KhronosGroup/Vulkan-Docs/compare/master...kainino0x:katex-faster-auto-render
Which I'll open a PR for after I poke at it some more.

krOoze:
> There's also a chance I could ask some knowledgeable Chrome folks about optimizing this page further.

OMG, yes, please, please. With the above PRs it is almost there. Though in my naivette I don't see a fundamental reason why it should take >1 s to load.

> I also have a small, non-intrusive change that significantly improves the katex loading time, here:
master...kainino0x:katex-faster-auto-render

Should probably get rid of the added `<span>` when replaced.
Block equations should be `<div>`, not `<span>`?

Anyway, PR #704 would get rid of all of that by translating it offline\server-side. Should also avoid triggering style and layout recalculation.

kainino0x:
> In my naivette I don't see a fundamental reason why it should take >1 s to load though.

I think it is inevitable at the sheer size of the generated DOM, except, probably, by hiding sections by default. However, it's possible that the complexity of layout could be reduced enough to make it quicker.

> Should probably get rid of the added <span> when replaced.

This might affect the surrounding DOM element, which could make it a lot more expensive. However, I will look into it.

> Block equations should be <div>, not <span>?

They're separately surrounded by a div (`<div><span></span></div>`) I believe, so this isn't necessary.

> Anyway, PR #704 would get rid of all of that by translating it offline\server-side. Should also avoid triggering style and layout recalculation.

I agree that it is probably a better solution - I intend to take a closer look at that patch. However I'd love to get a "quick fix" in sooner (doesn't require Node, for one thing).

krOoze:
> I think it is inevitable at the sheer size of the generated DOM

I would understand it on a cellphone or tablet. But on desktop processing measly 8 MB of data almost seems like one of the easiest types of tasks... Not to mention that HTML size is quite inflated by being human-readableish.

kainino0x:
on my (fast) workstation:
* With (my) script, with style: ~6.4s
* No script, with style: ~3.2s
* No script, no style (except for the inline style attributes): ~2.5s

So improving the styles could saves some time, but the real killer is the script itself and relayout caused by the script.

I do think it might be preferable to un-inline the images, if we could. I read something yesterday saying that data: uri images require re-decoding on every load, which I think could particularly impact closing and reopening vkspec in a single browser session (something I do a lot). It might reduce parsing time a little; not sure.

kainino0x:
> > Should probably get rid of the added `<span>` when replaced.

This turns out to be cheap, and seems to save about 15-20% of relayout time! Thanks for the tip!

kainino0x:
no script, img src="data:...": ~3.1s
no script, img src="": ~2.6s

This could be because the `` tags in vkspec do not have width and height. This is bad because it causes layout to change when the image is loaded and the browser suddenly knows the width and height of the image.

Fixing this would require some automation, so not sure how to do it.

krOoze:
@kainino0x The `asciidoctor` should be doing it itself if you remove the `:data-uri:` at https://github.com/KhronosGroup/Vulkan-Docs/blob/master/vkspec.txt#L7

kainino0x:
I actually meant that adding the width= and height= attributes would require automation - I don't know if asciidoctor can do that, plus it's a bit tricky to make it work in a way that is "responsive" to the window size.

But thanks for the pointer! I checked the performance of de-inlining the images, and I also found [`[%inline]`](https://github.com/asciidoctor/asciidoctor-bespoke#inserting-svgs).

no script, img src="data:...": 3.1s
no script, img src="images/...": 2.9s
no script, `[%inline]`: 3.05?s

So there's a tiny improvement here. This is on `localhost`, but with Chrome's "Fast 3G" emulation:

3g, no script, img src="data:...": 30.1s
3g, no script, img src="images/...": 28.2s
3g, no script, `[%inline]`: 28.0s

This is probably because the plaintext svg images are smaller than the base64 encoded images, so that might be the whole savings.
(I don't have gzip on my local server, which would also change the results.)

It seems from this that the `[%inline]` method would be the way to go.

[Inline SVG](https://caniuse.com/#feat=svg-html5) is about as well supported as [SVG itself](https://caniuse.com/#feat=svg).

---

I also tried [`svgo`](https://www.npmjs.com/package/svgo), an SVG optimizer, to minify all of the SVGs:

no script, img src="data:...": vkspec.html=5445607 B (697726 B gz)
no script, `[%inline]`: vkspec.html=5041131 B (609636 B gz)
no script, `[%inline]`, svgo: vkspec.html=4321077 B (546450 B gz)

And [`html-minifier`](http://kangax.github.io/html-minifier/) to minify the inline HTML/CSS/JS:

no script, `[%inline]`, svgo + html-minifier: vkspec.html=4000019 B (509007 B gz)

The minifiers seem to give a tiny load time improvement even without 3G emulation.

kainino0x:
Opened https://github.com/KhronosGroup/Vulkan-Docs/pull/734 for `[%inline]`.

kainino0x:
I think my preference would be to land #704 but not #702. Reason being:

#704 (at least for me) significantly reduces the page load time. However the most important thing it does is avoid ever forcing a relayout - layout happens incrementally while the page loads, and the page never freezes (stops being scrollable).

#702 (at least for me, in Chrome 67) causes a huge relayout AFTER the page has loaded - because the entire page still has to be laid out (it couldn't be done incrementally). Plus, it makes the page unusable (and unsearchable) while loading, so I think we shouldn't use it.

#708 also prevents some big relayouts (freezes) while the page is loading, especially on a slower connection where it takes longer to get the fonts in the first place. It doesn't improve loading time a lot for me, but it does greatly improve responsiveness.

@krOoze, what do you think?

krOoze:
@kainino0x 
#704 seems like a obvious low hanging fruit to me. It drags in a dependency on JS interpreter. Alternatively `asciidoctor-mathematical` can be used, but I find the output uglier. To remove the dependency I tried, and failed with `Js2Py` (which translates to Python directly). There are other JS interpreters for Python or Perl, but they drag in the Node in anyway. Depends how sensitive are you to add this dependency directly...

For #702 my original reasoning was that the page is unusable during the load anyway. When I tested, the page did not jump to clicked links, or requested anchor anyway during load, and the experience was laggy. This may change with browser updates though...

#708 should halve loading time -- saves one relayout; if the browser chooses wrong strategy to load the page in the first place. Browsers update faster than I change underwear, so this might change; but there should be no harm in doing this.

kainino0x:
#704: Personally I think it's fine to drag in the dependencies - I just wanted to also make a quicker fix we could land more easily. The dependencies only affect people who are building documentation, which is very few. I wouldn't want to do anything that hurts rendering quality.

#702: In Chrome 67, at least, the page works well while loading. I can scroll reasonably smoothly, click TOC links, and find-in-page.

#708: Yes, I think it's the principled, correct approach - so I don't think things could be worse with #708 than without it.

kainino0x:
I've put up a copy of vkspec.html with #704, #708, #734, and `svgo -f images/` here:
https://kai.graphics/vkspec/ ([repo](https://github.com/kainino0x/vkspec))
(mainly for personal use so that I can use it now instead of waiting for PRs to be merged).

FTR, I also just learned that running svgo multiple times results in slight filesize improvements. (1586 bytes total)

oddhack:

krOoze:
@oddhack Yay!

Not sure about this one (#702). It breaks the `chunked` (fix soon). Also since I originally made this, the browser updates were fighting it; they seem to do incremental present no matter what anyway so not sure if there is any worthwhile benefit.

#704 and #708 will be great though!

oddhack:
> @oddhack Yay!
> 
> Not sure about this one (#702). It breaks the `chunked` (fix soon).

I don't see any breakage in the chunked build on Linux Chrome or Firefox. What are you encountering?

> Also since I originally made this, the browser updates were fighting it; they seem to do incremental present no matter what anyway so not sure if there is any worthwhile benefit anymore.

#702 actually seemed to be giving the bulk of the performance increase for me. Oddly MacOS seems to get no benefit for anything except Safari, according to our one tester on that platform.

The font preloading as hard-wiring in font paths seems non-robust against future KaTeX versions, but I suppose that can just be dealt with when/it it occurs.

My main takeaway from our testing is that Chromium is doing something very wrong compared to other browsers.

krOoze:
@oddhack The script tries to "load" Preamble chapter, which is unnecessary. Also the Preamble chapter shows in 7.1. Render Pass Creation chapter for some reason. Not _that_ fatal.

krOoze:
> #702 actually seemed to be giving the bulk of the performance increase for me.

Huh. Which browser?
#704 + #708 give the most advante for me. The measurements at #704 are semi-updated for Firefox and Chrome.

> The font preloading hard-wiring in font paths seems non-robust against future KaTeX versions, but I suppose that can just be dealt with when/if it occurs.

Yea, that is annoying. But I am not too worried about it. They should not be changing fonts. The problem with it is we are only preloading subset of them, so some added latexmath may need need more fonts. E.g. it would be nice if Vulkan entities were in tt font in the math, which would necessitate adding to this preload.

oddhack:
> Huh. Which browser?

Current Linux Chromium.
 
> Yea, that is annoying. But I am not too worried about it. They should not be changing fonts. The problem with it is we are only preloading subset of them, so some added latexmath may need need more fonts. E.g. it would be nice if Vulkan entities were in tt font in the math, which would necessitate adding to this preload.

I'd like to get the asciidoc math ([eq] style) to use the same fonts / styles as the latexmath so there aren't jarring inconsistencies between them. Unfortunately this requires synthesizing fonts that contain some glyphs in roman style and some in italic from the KaTeX font set, AFAICT, and that's a big enough project not to be any kind of priority.

> Quite honestly it feels little bit that all the browsers are doing something wrong. Maybe I am bit naive, but 10 MB (human-readable) html is maybe like 2 MB real entropy (probably less). Feels like modern RAM and CPU should be able to deal with this under a second.

The Chromium codebase download was well over 1 GB when I built it a few years ago, IIRC. I doubt any human really understands what's going on inside more than a tiny part of it at this point.

oddhack:
> @oddhack The script tries to "load" Preamble chapter, which is unnecessary. Also the Preamble chapter shows in 7.1. Render Pass Creation chapter for some reason. Not _that_ fatal.

How peculiar. I just updated my build tree to use the current toolchain and that might be the cause, although the chunker itself hasn't updated in a very long time. I don't think any of your MRs would cause this. I hope it's not the chunker at issue because that's effectively a dead project at this point, AFAICT.

krOoze:
> I'd like to get the asciidoc math ([eq] style) to use the same fonts / styles as the latexmath so there aren't jarring inconsistencies between them.

I am not sure to what degree it is worth-while to still keep two math markups. Everything could just be latexmath now I think.

> I doubt any human really understands what's going on inside more than a tiny part of it at this point.

Yea, though even the "good" browser feels slow considering we have 1000x faster PCs now. Might also be partially a fault of webtech generally and HTML+CSS failure as a layouting language.

> I don't think any of your MRs would cause this.

I am adding to the HTML in #702, so it might. I will see when I get to it.

kainino0x:
This is amazing news! Thanks to @krOoze and the editors for getting these various improvements out!

kainino0x:
If you're encountering issues where Chromium performs worse than other browsers, please file them on crbug.com/new and send them to me to make sure they get properly routed.

oddhack:
@kainino0x Chrome / Chromium perform dramatically worse at loading the Vulkan 1.2 + extensions spec in spec updates 1.2.136 and earlier relative to Firefox on Linux / Windows (and apparently also Safari on MacOS, though I can't speak personally to that). With the changes included in 1.2.137 it's still generally slower, but down to maybe 1.5-2x instead of 3x. If you want to file an issue to that effect that would be great, but I can't imagine that this is going to get prioritized over the million other things people want to do to the browser so am not going to spend time on jumping through all the hoops that would be required to properly document this.

kainino0x:
Thanks for the info. I'll see if I can investigate it a little bit.

PR #704, opened 2018-04-16
Make latexmath translated offline

krOoze:
This translates latexmath offline during build using Node.js with KaTeX.

It improves loading time of the html, as well as responsivness of the browser during that load. Firefox especially seems to like it.

PS, some load time remeasurements (1.0 no extensions):

| Browser          | original | #704 (this)  |  #708 (font preload)    |   #704 + #708 |
|---------------|:--------:|:------------:|:---------:|:---------------:|
| Firefox 73.0.1 |  ~11 s  |  ~ 4 s          | ~10 s     | **~4 s**                |
| Chrome 80.0 |  ~21s    |  ~ 23 s (!)    | ~11.5 s   | **~9.5 s**            |


oddhack:
What's the interaction between this and #702, if any? Do you think they both make sense?

BUILD.adoc
```diff
@@ -537,6 +540,11 @@ recent version.
     included in .gitattributes as well.
 ====
 
+Node.js is best obtained through `nvm`.
+Follow installation instructions at https://github.com/creationix/nvm#install-script .
+Then install Node.js by running `nvm install stable`.
+Verify your installation by `node --version`.
+Install `escape-string-regexp` package by `npm install escape-string-regexp`.
```
oddhack: npm, or nvm?

krOoze: Both. `npm` installs packages, `nvm` manages Node.js versions.

krOoze:
@oddhack I haven't really tested it much, but they seem complementary.

But firefox seem to react to this surprisingly well and the "Loading" message stays only for few seconds. In chrome it seems to help more.

On the other hand, while it loads longer, the browser is much more responsive (usable) during that time, so it is something to consider too.

krOoze:
PS: Firefox seems to do something akin to #702 by default if you open new tab through link.
For Chrome though the responsiveness seem useless; it tries to incrementally render the page, but the content is useless to the user, because it shows the first page and jumps to the anchor he requested only after everything is loaded.

krOoze:
There's some problem making some html entities are visible :/

krOoze:
Fixed.

krOoze:
Alternatively, it is possible to simply use `asciidoctor-mathematical` as for the pdf target.

BUILD.adoc
```diff
@@ -537,6 +540,11 @@ recent version.
     included in .gitattributes as well.
 ====
 
+Node.js is best obtained through `nvm`.
+Follow installation instructions at https://github.com/creationix/nvm#install-script .
+Then install Node.js by running `nvm install stable`.
+Verify your installation by `node --version`.
+Install necessary Node.js packages by `npm install escape-string-regexp he`.
```
kainino0x: It might be better to add a `package.json`/`package-lock.json` to the repo, that way the modules get installed locally instead of globally.

BUILD.adoc
```diff
@@ -537,6 +540,11 @@ recent version.
     included in .gitattributes as well.
 ====
 
+Node.js is best obtained through `nvm`.
```
kainino0x: I don't think it's that important to give guidance on how to obtain npm - it's easy enough to search online, and pretty much any npm distribution should work fine. But I don't feel strongly.

BUILD.adoc
```diff
@@ -436,6 +438,7 @@ parts you don't use) completely before trying to install.
   * KaTeX distribution (version 0.7.0 from https://github.com/Khan/KaTeX .
     This is cached under `katex/`, and need not be
     installed from github.
+  * `escape-string-regexp` and `he` Node.js package
```
kainino0x: Wouldn't be necessary if using `package.json` (see below)

BUILD.adoc
```diff
@@ -418,6 +418,8 @@ work at least as well.
     This is for the PDF build, and it can still progress without it.
     Ghostscript is used to optimize the size of the PDF, so will be a lot
     smaller if it is included.
+  * Node.js
+    This is for HTML build to translate latexmath.
```
kainino0x: And npm, if using `package.json` (see below)

translate_math.js
```diff
@@ -0,0 +1,26 @@
+const katex = require("./katex/katex.js");
+const fs = require("fs");
+const escapeRegex = require("escape-string-regexp");
```
kainino0x: If we wanted to remove a dependency, this one would be easy to get rid of - could just replace `left: "\\["` with `leftEscaped: "\\\\["`, etc.

BUILD.adoc
krOoze: They are installed locally either way.
TBH I did not worry much about how to work with those files. I guess it might be somewhat nicer to require just `npm install` to install the dependencies...

BUILD.adoc
krOoze: Depends on the audience.
Personally, I prefer `BUILD.md`s to be written in such a way that a braindead person could build the project in 1 minute...

kainino0x: That's fair, it's always nice when all I have to do is follow some instructions without googling anything.

oddhack:
Finally had a chance to look at this. Is it supposed to be translating the latexmath constructs to MathML, which is still interpreted by katex at runtime? The load time on the resulting HTML in my Chromium instance doesn't seem significantly different from the non-translated HTML, though that's fairly subjective.

krOoze:
@oddhack IIRC KaTeX uses plain HTML and only adds hidden MathML for the readers. But yeah, it translates it on build time. It should only need the CSS and fonts on runtime.

It shaves several seconds for me. It removes all the JS computation on runtime(which also tends to freeze the browser), as well as remove a need for one page relayout (assuming all else loads well). It should definitely be felt (similarly to #733).
 
Please give #732 a look. It causes unpredictable perf for me, which does make comparisons hard.

BUILD.adoc
krOoze: Added the package.json defining the dependencies.

krOoze:
There also seem to be a `asciidoctor-katex` gem doing this...

krOoze:
@oddhack So, um, the LunarG survey once more bashes the poor single-page. Could we merge this? This is basically 2/3rds of the problem with load time. And the Node.js was already sneaked in anyway with `chunked`...

jeremyong:
Khronos please :)

The docs as they are are in a deplorable state and need to be prioritized higher. As things currently stand, the documentation continues to link the full spec page in a number of areas (typically, whenever a topic/chapter is referenced) and frankly even the chunked spec pages load too slowly for my liking. Relying on third-party hosting isn't an acceptable substitute, and its a shame as the documentation itself is well-written. I'm sure there are many things on your plate that are important to get to, but ... my two cents is that tasks such as these should really should supplant all that.

oddhack:

PR #708, opened 2018-04-22
Speed up Chrome by preloading fonts

krOoze:
For some reason Chrome does not include fonts in its `load` event, otherwise #702 would show better improvements. As a result it calculates the CSS DOM and page layout twice (it defers the font loading after the first frame is finished, and the fonts cause another recalculation &mdash; it might be reasonable strategy for pages that do not load several seconds).

This forces Chrome to preload the pesky fonts, and so remove one layout recalculation.

See #704 for measurements.

Note: Does not work with `file::/` (offline) if it contains `crossorigin`. But complains online (`http://`) without `crossorigin` :/.


oddhack:

PR #709, opened 2018-04-24
Update KaTeX to 0.10.0

krOoze:
TODO: diff page for rendered results for comparison


oddhack:
What does this enable, or fix, relative to the version we're using now?

krOoze:
@oddhack Sorry, just testing internally.

Which version is the current version anyway?

krOoze:
Ok, the changes are relatively minor.

For some reason, the root symbol is broken if rendered offline (using #704), e.g:
root_old2
vs
root_new2

There are some nice changes to the spacing:
spacing_old3
vs
spacing_new3

And switch brace size is corrected:
paren_old
vs
paren_new

kainino0x:
FYI, I did a load time test between the current KaTeX version and 0.10.0-beta, and I wasn't able to see a significant difference.

krOoze:
@kainino0x Yup, same for the 0.9 as I recall, but worth a try.
So it's just my bias towards updated software (assuming it has less bugs and rough edges).
I am not even sure what the current version is, so might be nice to have this stable-release version in.

kainino0x:
I do agree it would be nice to upgrade KaTeX. There are some concrete benefits, at least the spacing fix you mentioned above, which affects, for example, the math at the end of [15.2.1. RGB to Shared Exponent Conversion](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#textures-RGB-sexp).

old:
2018-06-29-160913_320x160_scrot

v0.10.0-beta:
2018-06-29-160923_317x168_scrot

oddhack:
I've updated to 10.0, but via an internal MR, so closing this.

PR #712, opened 2018-05-04
Fixed small typos

Whenning42:
Fixed some duplicate words and some spellings of "stages" as "stagess".


ianelliottus:
Seems like a no-brainer.

PR #731, opened 2018-06-28
Make offline the dependency on Font Awesome

krOoze:
Make the dependency on Font Awesome offline, so that the HTML can be viewed locally.


krOoze:
Updated with #770 icons. Had to change the script a bit, because it seems to break asciidoctor.

krOoze:
Not quite happy it adds MBs to html, I now made the icons loaded from file instead of inline.

PR #732, opened 2018-06-28
Reremove MathJax

krOoze:
make the removal script version and config agnostic


krOoze:
@oddhack Can I have this merged? Seems like a no-brainer bugfix. Currently it tends to run MathJax, as some versions of asciidoc updated the version.

PR #733, opened 2018-06-28
Replace katex's auto-render with something faster+simpler

kainino0x:
This replaces the `\(\)`/`\[\]` delimiters with `<span>`s, which allows the browser's parser to create the root nodes for math rendering. Without this, a lot of nodes end up getting split into several nodes (at each `\(\)`/`\[\]`) in order to insert the math root nodes - which has a huge cost in the browser (when manipulating these nodes and when performing layout on the changed DOM).

Shaves roughly 25~35% off of page load time in Chrome, loading from localhost, depending on the machine and environmental conditions.

See also discussion here: https://github.com/KhronosGroup/Vulkan-Docs/pull/702#issuecomment-400879142

### DIFF
--- a/config/katex_replace/extension.rb
+++ b/config/katex_replace/extension.rb
@@ -28,28 +28,31 @@ class ReplaceMathjaxWithKatex < Extensions::Postprocessor
 
       katexScript = '<link rel="stylesheet" href="' + katexpath + '/katex.min.css">
 <script src="' + katexpath + '/katex.min.js"></script>
-<script src="' + katexpath + '/contrib/auto-render.min.js"></script>
-    <!-- Use KaTeX to render math once document is loaded, see
-         https://github.com/Khan/KaTeX/tree/master/contrib/auto-render -->
 <script>
+    "use strict";
+    function renderMathForHTMLCollection(htmlCollection, displayMode) {
+        // Clone elements into an array so it cannot change underneath us:
+        var elements = Array.prototype.slice.call(htmlCollection);
+        for (var i = 0; i < elements.length; ++i) {
+            var m = elements[i];
+            katex.render(m.textContent, m, { displayMode: displayMode });
+            m.outerHTML = m.innerHTML;
+        }
+    }
+
     document.addEventListener("DOMContentLoaded", function () {
-        renderMathInElement(
-            document.body,
-            {
-                delimiters: [
-                    { left: "$$", right: "$$", display: true},
-                    { left: "\\\\\[", right: "\\\\\]", display: true},
-                    { left: "$", right: "$", display: false},
-                    { left: "\\\\\(", right: "\\\\\)", display: false}
-                ]
-            }
-        );
+        renderMathForHTMLCollection(
+            document.getElementsByClassName("latex-math-inline"), false);
+        renderMathForHTMLCollection(
+            document.getElementsByClassName("latex-math-display"), true);
     });
 </script>'
 
       output.sub! MathJaXScript, ''
       output.sub! MathJaXCDN, ''
       output.sub! /(?=<\/head>)/, katexScript
+      output.gsub! /\\\((.*?)\\\)/m, '<span class="latex-math-inline">\1</span>'
+      output.gsub! /\\\[(.*?)\\\]/m, '<span class="latex-math-display">\1</span>'
     end
     output
   end


krOoze:
Ah, the math can contain `//[` as in here:
https://github.com/KhronosGroup/Vulkan-Docs/blob/5300d9f05e03f721812273d5f9f9676c3227f056/chapters/textures.txt#L1963

You must replace the delimiters by a regex similarly as I did in https://github.com/KhronosGroup/Vulkan-Docs/pull/704/files#diff-89dc4f21591a8d9d864f2cc4228ba554. Basically `opendelim + \w+? + close delim`

krOoze:
PS: I will also try mesuring it later when you fix it. But there's a tip: cherrypick the #732 bugfix. It pretty much eliminates the time variance.

kainino0x:
> Ah, the math can contain another `\\[` as in here:

Oof, I forgot to check for cases like this. I wonder why I did not get any console errors, or anything like that.

Thanks for the tip on timing, I'll try it.

config/katex_replace/extension.rb
```diff
@@ -28,28 +28,37 @@ def process document, output
...
-                    { left: "\\\\\(", right: "\\\\\)", display: false}
-                ]
+        var elements, m;
+
+        elements = document.getElementsByClassName("latex-math-inline");
+        for (var i = 0; i < elements.length; ++i) {
+            m = elements[i];
+            katex.render(m.textContent, m, { displayMode: false });
+            if (m.childNodes.length == 1) {
+                m.parentNode.replaceChild(m.firstChild, m);
```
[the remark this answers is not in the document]
kainino0x: Oops, I think it was actually just faster because it was causing the HTMLCollection (result of the getElementsByClassName) to change underneath us (as elements were removed). So it was only rendering half of the elements.

kainino0x:
Unfortunately #732 didn't eliminate the loading time variance for me. I suspect that most of the variance I see is coming from GC nondeterminism (different memory pressure at different times, or something).

kainino0x:
OK, done.

Hopefully this is good - I don't want to invest too much more time into it. I really think that #704 is the way to go ultimately.

@krOoze you can take another look if you want - thanks so much for the feedback and reviews!

@oddhack PTAL - I think it would be best to merge #704 eventually, but do you think we could apply this fix in the meantime?

katex/contrib/auto-render.min.js
```diff
@@ -1 +0,0 @@
...ue},{left:"\\(",right:"\\)",display:false}],ignoredTags:["script","noscript","style","textarea","pre","code"]};var f=function(e){var t;var r;for(var n=1,a=arguments.length;n<a;n++){t=arguments[n];for(r in t){if(Object.prototype.hasOwnProperty.call(t,r)){e[r]=t[r]}}}return e};var d=function(e,t){if(!e){throw new Error("No element provided to render")}t=f({},l,t);o(e,t.delimiters,t.ignoredTags)};t.exports=d},{"./splitAtDelimiters":2}],2:[function(e,t,r){var n=function(e,t,r){var n=r;var a=0;var i=e.length;while(n<t.length){var o=t[n];if(a<=0&&t.slice(n,n+i)===e){return n}else if(o==="\\"){n++}else if(o==="{"){a++}else if(o==="}"){a--}n++}return-1};var a=function(e,t,r,a){var i=[];for(var o=0;o<e.length;o++){if(e[o].type==="text"){var l=e[o].data;var f=true;var d=0;var s;s=l.indexOf(t);if(s!==-1){d=s;i.push({type:"text",data:l.slice(0,d)});f=false}while(true){if(f){s=l.indexOf(t,d);if(s===-1){break}i.push({type:"text",data:l.slice(d,s)});d=s}else{s=n(r,l,d+t.length);if(s===-1){break}i.push({type:"math",data:l.slice(d+t.length,s),rawData:l.slice(d,s+r.length),display:a});d=s+r.length}f=!f}i.push({type:"text",data:l.slice(d)})}else{i.push(e[o])}}return i};t.exports=a},{}]},{},[1])(1)});
```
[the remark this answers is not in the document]
kainino0x: Ok, fair. I thought that all of the other contrib scripts had been removed in this snapshot, but actually there just weren't any other contrib scripts in katex at that time.

config/katex_replace/extension.rb
```diff
@@ -28,28 +28,31 @@ def process document, output
...
+        renderMathForHTMLCollection(
+            document.getElementsByClassName("latex-math-display"), true);
     });
 </script>'
 
       output.sub! MathJaXScript, ''
       output.sub! MathJaXCDN, ''
       output.sub! /(?=<\/head>)/, katexScript
+      output.gsub! /\\\((.*?)\\\)/m, '<span class="latex-math-inline">\1</span>'
+      output.gsub! /\\\[(.*?)\\\]/m, '<span class="latex-math-display">\1</span>'
```
krOoze: `.` is not supposed to be matching newlines according to the internet...

kainino0x: I didn't think it would, but it seems to work. Specifically, the `/m` (multiline?) at the end was necessary.

krOoze: Right, apparently it's some Ruby-specific thing...

krOoze:
@kainino0x May be, though for me #732 bug confuses Chrome profiler.
With the fix and some warmup load I get replicable ~31&plusmn;1 s (for the original w\ all ext). 
With this PR I get ~21 s. It reduces the math recalculation from ~ 5 s to 1 s, and somehow shortens one of the layout recalculations 👍  Firefox seems to like this too.

kainino0x:
Awesome, thanks for the feedback!

krOoze:
@kainino0x @oddhack #704 was accepted, so this can be closed I think.

kainino0x:
Agree!

PR #734, opened 2018-06-29
Use [%inline] for all SVGs

kainino0x:
This inlines `<svg>` elements in the HTML, instead of using `` tags with `src="data:..."`.

Provides a 12.6% size savings for vkspec.html
(697770 B -> 609679 B, when served gzipped, as on khronos.org), because:
1. plaintext is smaller than base64-encoded plaintext, and
2. base64 does not gzip as well as plaintext SVG.

Documentation on `[%inline]` can be found here:
https://github.com/asciidoctor/asciidoctor-bespoke#inserting-svgs

### DIFF
--- a/chapters/cmdbuffers.txt
+++ b/chapters/cmdbuffers.txt
@@ -101,6 +101,7 @@ Invalid::
     Command buffers in the invalid state can: only be reset or freed.
 
 [[commandbuffer-lifecycle-diagram]]
+[%inline]
 image::images/commandbuffer_lifecycle.svg[title="Lifecycle of a command buffer",align="center"]
 
 Any given command that operates on a command buffer has its own requirements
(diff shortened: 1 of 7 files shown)


oddhack:
The result doesn't look objectionable, but I don't understand either why it works (e.g. where is the inline option for the image directive documented? Not in the user's guide, AFAICT), why the inline option differs from the image: (single :) directive, or what the upside of increasing the compressibility is.

ratchetfreak:
> or what the upside of increasing the compressibility is

less bandwidth used when using a server that will provide a compressed data content-encoding option, also less disk space when the browser uses compression for its cache store.

For example the khronos.org server will serve the spec with `Content-Encoding: gzip`

krOoze:
@oddhack It was linked in the other thread. Here you go:
https://github.com/asciidoctor/asciidoctor-bespoke#inserting-svgs

kainino0x:
Updated the summary to include the above points.

One more thing - it doesn't JUST improve compressibility, it also significantly improves the uncompressed filesize (also by about 12.5%). I included the numbers for compressed filesize because that is what the Khronos website uses.

PR #742, opened 2018-07-12
Fix VkSubpassDependency self-dependency stage VU

krOoze:
Fix VkSubpassDependency self-dependency stage VU

- Fix `ifdef` inside VU entry, which e.g. breaks the `validusage.json`.
- Allow multiple stage bits to be set, not just one. (Other VUs seem to expect multiple bits, and there's no reason to ban it I think).

### DIFF
--- a/chapters/synchronization.txt
+++ b/chapters/synchronization.txt
@@ -549,8 +549,9 @@ guaranteed:
   * ename:VK_PIPELINE_STAGE_HOST_BIT
 
 ifdef::VK_EXT_conditional_rendering[]
-For conditional rendering operations, the pipeline stage where predicate
-read happens has no particular order relative to other stages.
+The conditional rendering stage is formally part of both the graphics, and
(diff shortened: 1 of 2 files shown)


chapters/synchronization.txt
```diff
@@ -549,8 +549,9 @@ guaranteed:
   * ename:VK_PIPELINE_STAGE_HOST_BIT
 
 ifdef::VK_EXT_conditional_rendering[]
-For conditional rendering operations, the pipeline stage where predicate
-read happens has no particular order relative to other stages.
+Conditional rendering stage is considered part of the graphics, and compute
```
oddhack: "The conditional rendering..."

oddhack: "the graphics and compute pipelines".

krOoze: I attempted some rewording here...

PR #746, opened 2018-07-20
Delete wrong <unused> tag

krOoze:
`-12` is actually used, and the tag does not seem to be useful for anything...


xml/vk.xml
```diff
@@ -3725,7 +3725,6 @@ server.
         <enum value="-10"   name="VK_ERROR_TOO_MANY_OBJECTS" comment="Too many objects of the type have already been created"/>
         <enum value="-11"   name="VK_ERROR_FORMAT_NOT_SUPPORTED" comment="Requested format is not supported on this device"/>
         <enum value="-12"   name="VK_ERROR_FRAGMENTED_POOL" comment="A requested pool allocation has failed due to fragmentation of the pool's memory"/>
-            <unused start="-12"/>
```
oddhack: We could change it to something like

> <unused start="-13" comment="This is the next unused available error code"/>

but I want to retain the tag, as it's an unusual case of allocating enums in increasing negative value.

krOoze: Alright, changed that way.
Though one should not be adding enums to this list. And the one time it happened, the author did not read the `<unused>` tag anyway...

PR #748, opened 2018-07-23
Fix metadoc script showing non-selected extensions

krOoze:
mostly added comments
`doc += content` -> `doc += innerdoc ` does the fix


oddhack:
LGTM.

PR #749, opened 2018-07-23
Reapply #742:  Make STAGE_CONDITIONAL_RENDERING part of graphics pipe

krOoze:
` VK_PIPELINE_STAGE_CONDITIONAL_RENDERING_BIT_EXT` has to be defined as part of graphics pipeline in order for VU modified in #742 to be accurate

### DIFF
--- a/chapters/synchronization.txt
+++ b/chapters/synchronization.txt
@@ -537,6 +537,14 @@ For the compute pipeline, the following stages occur in this order:
   * ename:VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT
   * ename:VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT
 
+ifdef::VK_EXT_conditional_rendering[]
+The conditional rendering stage is formally part of both the graphics, and
+the compute pipeline. The pipeline stage where the predicate read happens
+has unspecified order relative to other stages of these pipelines:
+
+  * ename:VK_PIPELINE_STAGE_CONDITIONAL_RENDERING_BIT_EXT
+endif::VK_EXT_conditional_rendering[]
+
 For the transfer pipeline, the following stages occur in this order:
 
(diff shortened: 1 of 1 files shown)


chapters/synchronization.txt
```diff
@@ -549,8 +549,9 @@ guaranteed:
   * ename:VK_PIPELINE_STAGE_HOST_BIT
 
 ifdef::VK_EXT_conditional_rendering[]
-For conditional rendering operations, the pipeline stage where predicate
-read happens has no particular order relative to other stages.
+The conditional rendering stage is formally part of both the graphics, and
```
oddhack: I see that #742 wasn't squashed to a single commit when I cherry-picked it into our internal gitlab, which is why this didn't make it in. That said, we think it would be clearer to actually list this bit as part of both the graphics and compute stage lists, probably with a reference to the "unspecified order" comment here rather than repeating it in both lists.

krOoze: Alternatively, it can be another pseudo-pipeline, like device-generated commands pipeline... seems very similar case.

BTW shouldn't `VK_PIPELINE_STAGE_COMMAND_PROCESS_BIT_NVX` be allowed in `VkSubpassDependency` too?

vikramkushwaha: We considered calling it a pseudo stage initially (merge request 2676) but later decided against it. I am fine with this change.

krOoze: @oddhack I am not sure how to practically wordsmith that. The graphics and compute pipelines say "the following stages occur in this order:". I am not sure how to clearly list this bit there as unordered.

Some improvement could be to move this paragraph below COMPUTE, so it is at least closer to those related pipelines.

vikramkushwaha: Yes that would be a slight improvement. I dont think its necessary to add the bit to either graphics or compute pipeline stages in the paragraph above.

krOoze: I made it so then.

PR #759, opened 2018-08-06
Add some missing "optional" attributes

Ralith:
The comments on these members suggest that it is sometimes legal for them to contain `VK_NULL_HANDLE`, or in some cases an undefined value. Unless I've misunderstood the schema, it seems like they should therefore be marked as optional.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -747,9 +747,9 @@ server.
             <member><type>VkDeviceSize</type>           <name>range</name><comment>Size in bytes of the buffer resource for this descriptor update.</comment></member>
         </type>
         <type category="struct" name="VkDescriptorImageInfo">
-            <member noautovalidity="true"><type>VkSampler</type>       <name>sampler</name><comment>Sampler to write to the descriptor in case it is a SAMPLER or COMBINED_IMAGE_SAMPLER descriptor. Ignored otherwise.</comment></member>
-            <member noautovalidity="true"><type>VkImageView</type>     <name>imageView</name><comment>Image view to write to the descriptor in case it is a SAMPLED_IMAGE, STORAGE_IMAGE, COMBINED_IMAGE_SAMPLER, or INPUT_ATTACHMENT descriptor. Ignored otherwise.</comment></member>
-            <member noautovalidity="true"><type>VkImageLayout</type>   <name>imageLayout</name><comment>Layout the image is expected to be in when accessed using this descriptor (only used if imageView is not VK_NULL_HANDLE).</comment></member>
+            <member noautovalidity="true" optional="true"><type>VkSampler</type>       <name>sampler</name><comment>Sampler to write to the descriptor in case it is a SAMPLER or COMBINED_IMAGE_SAMPLER descriptor. Ignored otherwise.</comment></member>
+            <member noautovalidity="true" optional="true"><type>VkImageView</type>     <name>imageView</name><comment>Image view to write to the descriptor in case it is a SAMPLED_IMAGE, STORAGE_IMAGE, COMBINED_IMAGE_SAMPLER, or INPUT_ATTACHMENT descriptor. Ignored otherwise.</comment></member>
+            <member noautovalidity="true" optional="true"><type>VkImageLayout</type>   <name>imageLayout</name><comment>Layout the image is expected to be in when accessed using this descriptor (only used if imageView is not VK_NULL_HANDLE).</comment></member>
         </type>
         <type category="struct" name="VkWriteDescriptorSet">
             <member values="VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET"><type>VkStructureType</type> <name>sType</name></member>
@@ -1558,7 +1558,7 @@ server.
             <member><type>VkBool32</type>                         <name>persistentContent</name><comment>VK_TRUE if this is a "smart" display that supports self-refresh/internal buffering.</comment></member>
         </type>
         <type category="struct" name="VkDisplayPlanePropertiesKHR" returnedonly="true">
-            <member><type>VkDisplayKHR</type>                     <name>currentDisplay</name><comment>Display the plane is currently associated with.  Will be VK_NULL_HANDLE if the plane is not in use.</comment></member>
+            <member optional="true"><type>VkDisplayKHR</type>                     <name>currentDisplay</name><comment>Display the plane is currently associated with.  Will be VK_NULL_HANDLE if the plane is not in use.</comment></member>
             <member><type>uint32_t</type>                         <name>currentStackIndex</name><comment>Current z-order of the plane.</comment></member>
         </type>
         <type category="struct" name="VkDisplayModeParametersKHR">
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -738,7 +738,7 @@ server.
             <member><type>VkExtent3D</type>             <name>maxExtent</name><comment>max image dimensions for this resource type</comment></member>
             <member><type>uint32_t</type>               <name>maxMipLevels</name><comment>max number of mipmap levels for this resource type</comment></member>
             <member><type>uint32_t</type>               <name>maxArrayLayers</name><comment>max array size for this resource type</comment></member>
-            <member optional="true"><type>VkSampleCountFlags</type>     <name>sampleCounts</name><comment>supported sample counts for this resource type</comment></member>
+            <member><type>VkSampleCountFlags</type>     <name>sampleCounts</name><comment>supported sample counts for this resource type</comment></member>
             <member><type>VkDeviceSize</type>           <name>maxResourceSize</name><comment>max size (in bytes) of this resource type</comment></member>
         </type>
         <type category="struct" name="VkDescriptorBufferInfo">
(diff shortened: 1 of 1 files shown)


xml/vk.xml
```diff
@@ -742,14 +742,14 @@ server.
...
-            <member noautovalidity="true"><type>VkImageLayout</type>   <name>imageLayout</name><comment>Layout the image is expected to be in when accessed using this descriptor (only used if imageView is not VK_NULL_HANDLE).</comment></member>
+            <member noautovalidity="true" optional="true"><type>VkSampler</type>       <name>sampler</name><comment>Sampler to write to the descriptor in case it is a SAMPLER or COMBINED_IMAGE_SAMPLER descriptor. Ignored otherwise.</comment></member>
+            <member noautovalidity="true" optional="true"><type>VkImageView</type>     <name>imageView</name><comment>Image view to write to the descriptor in case it is a SAMPLED_IMAGE, STORAGE_IMAGE, COMBINED_IMAGE_SAMPLER, or INPUT_ATTACHMENT descriptor. Ignored otherwise.</comment></member>
+            <member noautovalidity="true" optional="true"><type>VkImageLayout</type>   <name>imageLayout</name><comment>Layout the image is expected to be in when accessed using this descriptor (only used if imageView is not VK_NULL_HANDLE).</comment></member>
```
krOoze: So, `noautovalidity` already implies it can be anything. `optional` is used by autovalidity to create an implicit VU.

It somewhat makes sense, but I don't think these cases have `optional` throughout the XML.

xml/vk.xml
```diff
@@ -1558,7 +1558,7 @@ server.
             <member><type>VkBool32</type>                         <name>persistentContent</name><comment>VK_TRUE if this is a "smart" display that supports self-refresh/internal buffering.</comment></member>
         </type>
         <type category="struct" name="VkDisplayPlanePropertiesKHR" returnedonly="true">
-            <member><type>VkDisplayKHR</type>                     <name>currentDisplay</name><comment>Display the plane is currently associated with.  Will be VK_NULL_HANDLE if the plane is not in use.</comment></member>
+            <member optional="true"><type>VkDisplayKHR</type>                     <name>currentDisplay</name><comment>Display the plane is currently associated with.  Will be VK_NULL_HANDLE if the plane is not in use.</comment></member>
```
krOoze: It is `returnedonly` which already implies `optional` (actually more than that, the value is ignored).

xml/vk.xml
```diff
@@ -742,14 +742,14 @@ server.
             <member><type>VkDeviceSize</type>           <name>maxResourceSize</name><comment>max size (in bytes) of this resource type</comment></member>
         </type>
         <type category="struct" name="VkDescriptorBufferInfo">
-            <member><type>VkBuffer</type>               <name>buffer</name><comment>Buffer used for this descriptor slot when the descriptor is UNIFORM_BUFFER[_DYNAMIC] or STORAGE_BUFFER[_DYNAMIC]. VK_NULL_HANDLE otherwise.</comment></member>
+            <member optional="true"><type>VkBuffer</type>               <name>buffer</name><comment>Buffer used for this descriptor slot when the descriptor is UNIFORM_BUFFER[_DYNAMIC] or STORAGE_BUFFER[_DYNAMIC]. VK_NULL_HANDLE otherwise.</comment></member>
```
krOoze: I think the comment might be wrong, or misleading. Comments tend to be evil this way...

From the spec I get the impression that it either has to be valid, or the whole structure array is ignored completely.

xml/vk.xml
Ralith: The value is ignored by the implementation, but probably not by an application. For context, I'm trying to use `optional="true"` attributes to improve static safety of nullable vulkan handles in a generated binding via types (see https://github.com/MaikKlein/ash/pull/96), so cases where the implementation may return a null handle are expressly of interest, since an objective is to force the caller to check the value in those cases rather than assuming it's nonnull.

This seems compatible with the definition of the attribute given in [the registry documentation](https://www.khronos.org/registry/vulkan/specs/1.1/registry.html), and I don't see where `returnedonly` is given to imply optionality. Having this information available with fine-grained detail rather than treating any returned handle at all as possibly null would be nice.

xml/vk.xml
Ralith: Grep shows there are 23 pre-existing cases of a member that has both `noautovalidity="true"` and `optional="true"`, so I'm not the first to think this way.

krOoze: Well, should be harmless in that case...

xml/vk.xml
krOoze: That's laudable, though it is a stretch that it applies to post-conditions.
Though there are precedences in the XML (especially flags). There is no Diff in validity, so I guess this is harmless too...

Ralith: > though it is a stretch that the attribute currently applies to post-conditions

I think this provides a pretty strong motivating case that it should. Regardless, this too is not the first example of such a pattern. Pre-existing optional members in returned-only structs include:
- `VkQueueFamilyProperties::queueFlags`
- `VkSparseImageFormatProperties::aspectMask` and `::flags`
- `VkMemoryType::propertyFlags`
- `VkImageFormatProperties::sampleCounts`

and many others.

kvark:
@krOoze thank you for taking a quick look! It appears to me that review got stalled. Do you have any unresolved concerns?

krOoze:
@kvark I personally do not object to this. But caveat emptor, as these `optional` attributes are not used for anything visible, I would think they are probably wrong several times across the XML (or missing where they should be).

Also it is not obvious to me from schema doc that they should be used this way, and semantics of such use if they are (especially `returnedonly` structs). Probably needs some clarifications there later...

Someone from Khronos still needs to review this (I am not, if that was not clear...), so they will have some authoritative answers.

krOoze:
To see what I mean, for example these `returnedonly` structs:

 - `VkImageFormatProperties::sampleCounts` must be at least one bit, but is marked `optional`
 - `VkMemoryHeap"::flags` can be 0, so is marked `optional`
 - `VkDisplayPropertiesKHR::display` must return valid `VkDisplayKHR`, but it lacks `optional` attribute (neither `true` nor `false`)
 - And, as you discovered, `VkDisplayPlanePropertiesKHR::currentDisplay` can return null handle, but also lacks `optional` attribute (neither `true` nor `false`)

That's all four possible interpretations of `optional` attribute right there.

Ralith:
We should try to move towards a consistent and meaningful use of the attribute, otherwise it's not worth having at all. I think we're most of the way there already, so it makes sense to clean up the remaining errors. The information is manifestly useful.

> VkImageFormatProperties::sampleCounts must be at least one bit, but is marked optional="true"

Good catch; I'll add this to the fixes in this PR.

> VkDisplayPropertiesKHR::display must return valid VkDisplayKHR, but it lacks optional attribute (neither true nor false)

It's reasonable to judge that members/arguments not explicitly marked as optional are indeed not optional, as this is by far the common case.

krOoze:
> We should try to move towards a consistent and meaningful use of the attribute, otherwise it's not worth having at all.

It is worth having, if there is **not** `returnedonly` or `noautovalidity`. It is clear in those cases. And that is what it is used for in this repo.

> Good catch; I'll add this to the fixes in this PR.

There's more if you want. E.g. IIRC sample flags in physical device feature struct. I will give it a look if you want an exhaustive list (there's ~ 100 `returnedonly` structs. I have not checked `noautovalidity` yet though).

I still think `optional` should not apply to post-conditions though. It would be slightly confusing. And also cannot be used for `struct`s that are **not** `returnedonly`, but are sometimes returned from the API (i.e. in that case the `optional` already applies to pre-conditions\VU, but not necessarily to the retured struct too). If needed, new attribute should perhaps be added to the XML...

> It's reasonable to judge that members/arguments not explicitly marked as optional are indeed not optional, as this is by far the common case.

Reasonable. But, I opened #769 for this.

Ralith:
> It is worth having, if there is not returnedonly or noautovalidity. It is clear in those cases. And that is what it is used for in this repo.

There's value in having it in those cases as well, and no shortage of precedent.

> There's more if you want. E.g. IIRC sample flags in physical device feature struct. I will give it a look if you want an exhaustive list (there's ~ 100 returnedonly structs. I have not checked noautovalidity yet though).

That would be great! You might want to wait until we have feedback from Khronos, although many of these seem to be bugs regardless of preferred outcome.

> It would be slightly confusing.

Many restrictions are already specified for returned members, and it is understood that this applies to them as returned. Nullability doesn't seem categorically different to me.

> cannot be used for structs that are not returnedonly, but are sometimes returned from the API (i.e. in that case the optional already applies to pre-conditions\VU, but not necessarily to the retured struct too).

Even if such a member actually exists, it would be useful to classify it as optional to distinguish it from members which must always be nonnull. We can't specify the entire semantics of Vulkan in a machine-readable way, but increasing the detail that can be specified is still valuable to applications such as diagnostic tools and richly-typed bindings.

oddhack:
Assigned to @Tobski for feedback on the attribute.

xml/vk.xml
Tobski: I would imagine (I haven't looked) that those are probably leftovers from when we introduced the noautovalidity statement. As @krOoze suggests these have no effect - I'd rather not add more of these as it confuses things. If you wanted to remove the others please be my guest 😄

xml/vk.xml
Tobski: So the optional attribute is indeed, as @krOoze suggests, only really meant for validation on the implementation side. 

I'm not sure how comfortable I am with this as at some point I would like to wrench out all of the autovalidity from the XML and put it directly into the spec instead; if there's a vestigial utility of the optional attribute I don't know what we'd do with it. I thus would prefer this to be a separate attribute, rather than muddying the water on what's already there.

It's also worth pointing out that the optional attribute came before the readonly attribute; the examples above are just things I missed when I added the "readonly" attribute, and don't serve any functional purpose.

Ralith:
> So the optional attribute is indeed, as @krOoze suggests, only really meant for validation on the implementation side.

I see. As I outlined above, there is utility in encoding a superset of this information: both whether input fields may be null when supplied to the implementation, and whether output fields might be set to null by the implementation. This allows generated bindings such as the one I'm working on to improve the statically-checked correctness of applications.

If vk.xml's metadata is intended to be used to generate third-party bindings in addition to its role in generating official documentation, this type of metadata seems worth having provided that the maintenance burden isn't unreasonable. Metadata that has no effect on the generated documentation can still be valuable for other purposes. Overloading the existing optional attribute to have these semantics doesn't seem ambiguous to me, but if replacing it outright is preferable, that works too.

Otherwise, if it's preferred not to maintain metadata that isn't consumed internally, then removing the dead attributes would reduce future confusion.

Tobski:
> If vk.xml's metadata is intended to be used to generate third-party bindings in addition to its role in generating official documentation, this type of metadata seems worth having provided that the maintenance burden isn't unreasonable. 

In general we've resisted additions that are *only* for external toolchains, mostly based on the fact that if we haven't got anything relying on it, it won't get tested and will almost definitely fall out of date.

I personally think this piece of information is probably somewhat useful to put into the xml though, and we might consider incorporating this information into the spec at some point too (giving it coverage).

> Overloading the existing optional attribute to have these semantics doesn't seem ambiguous to me, but if replacing it outright is preferable, that works too.

Again, my main reason for wanting to use a different attribute is down to wanting to avoid issues later when I get around to moving the VU stuff out of the xml - then at least there's then a clean separation without having to go back and forth to figure out which "optional" attribute is only for VU, and which ones affect your toolchain.

PR #760, opened 2018-08-06
Fix small typos

Whenning42:
Change occurrences of "is an Vk*" to "is a Vk*"

### DIFF
--- a/chapters/VK_EXT_debug_utils.txt
+++ b/chapters/VK_EXT_debug_utils.txt
@@ -553,7 +553,7 @@ include::../api/structs/VkDebugUtilsMessengerCallbackDataEXT.txt[]
   * pname:sType is the type of this structure.
   * pname:pNext is `NULL` or a pointer to an extension-specific structure.
   * pname:flags is 0 and reserved for future use.
-  * pname:pMessageIdName is a null-terminated string that identifies the the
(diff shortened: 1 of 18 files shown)


krOoze:
While at it would you mind adding more cases, or should I?

Should actually be straightforward by a text search for "a [vowel]" and "an [consonant]" +- the grammar weirdnesses.

Whenning42:
I think this fixes most of the typos that are of the form "a [vowel]" and "an [consonant]", however there are still probably some typos of the form "a [consonant that sounds like vowel]" and "an [vowel that sounds like consonant]".

chapters/VK_EXT_debug_utils.txt
```diff
@@ -553,7 +553,7 @@ include::../api/structs/VkDebugUtilsMessengerCallbackDataEXT.txt[]
   * pname:sType is the type of this structure.
   * pname:pNext is `NULL` or a pointer to an extension-specific structure.
   * pname:flags is 0 and reserved for future use.
-  * pname:pMessageIdName is a null-terminated string that identifies the the
+  * pname:pMessageIdName is a null-terminated string that identifies the
```
krOoze: good too. You used regex to find those duplicates?

PR #762, opened 2018-08-07
Remove misleading wording in VkDescriptorBufferInfo::buffer comment

Ralith:
This wording appears to refer to the conditions under which e.g. the `pBufferInfo` member of `VkWriteDescriptorSet` is not ignored, rather than the field in question, which to my reading must always be set to a valid buffer in a non-ignored `VkDescriptorBufferInfo`.

CC @krOoze


krOoze:
Yes, it does makes more sense to me this way. Valid buffer VU should stay (and the XML comments should reflect it), becuase:

1) Other VUs of the `VkDescriptorBufferInfo` do not really expect `VK_NULL_HANDLE` for `buffer`
2) The `VkDescriptorImageInfo` analog works that way too (i.e. value is ignored, or must be valid)

PR #764, opened 2018-08-08
WIP: Allow additional handle type safety in C++

krOoze:
Allow additional handle type safety in C++ through use of classes.
Inspired by https://stackoverflow.com/questions/51743500/why-do-non-dispatchable-handles-use-a-ptr-on-64bit-platofrms which claimed "C and **C++** don't have strong typedefs".

Pros:
 - proper static handle types
 - control over what operations and conversions are allowed over handles
 - Actually seems to work on Windows (https://docs.microsoft.com/en-us/cpp/build/parameter-passing claims 64 bit structs behave as integers in ABI)
 - prolly can be same for 32 and 64 bits platform

Cons:
 - the XML mess
 - overhead risk ? (if compiler is dumb)
 - ?

WIP:
 - investigate ABI on other platforms (and gate this appropriatelly in the header)
 - investigate what operations people use (or would like to) on handles
 - make similar stuff for dispatchable handles
 - needs testing

Anyway, I would like your thoughts in the meantime.

### DIFF
--- a/include/vulkan/vulkan_core.h
+++ b/include/vulkan/vulkan_core.h
@@ -46,7 +46,12 @@ extern "C" {
 #define VK_HEADER_VERSION 82
 
 
-#define VK_NULL_HANDLE 0
+#if defined(__cplusplus) && ( __cplusplus >= 201100 || _MSC_VER >= 1800)
+    struct Null_t{ explicit operator uint64_t() const { return 0; } };
+    #define VK_NULL_HANDLE Null_t{}
+#else
+    #define VK_NULL_HANDLE 0
+#endif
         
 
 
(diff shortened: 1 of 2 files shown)


include/vulkan/vulkan_core.h
```diff
@@ -46,15 +46,36 @@ extern "C" {
 #define VK_HEADER_VERSION 82
 
 
-#define VK_NULL_HANDLE 0
+#ifdef __cplusplus
+    struct Null_t{ explicit operator uint64_t() const { return 0; } };
```
Ralith: Could this be `std::nullptr_t`?

include/vulkan/vulkan_core.h
```diff
@@ -46,15 +46,36 @@ extern "C" {
...
+    #define VK_NULL_HANDLE Null_t{}
```
Ralith: Could this be `std::nullptr_t`?

include/vulkan/vulkan_core.h
```diff
@@ -46,15 +46,36 @@ extern "C" {
...
+#ifdef __cplusplus
+        #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) \
+            class object{ \
+            private: \
+                uint64_t h; \
+            public: \
+                object(){}; \
+                object( const Null_t& nh ){ h = static_cast<uint64_t>(nh); } \
+                explicit object( const uint64_t rh ){ h = rh; } \
+                operator bool() const{ return this->h; } \
```
Ralith: Would it be better for this to be `explicit`? Implicit conversion to bool can sometimes be very surprising.

include/vulkan/vulkan_core.h
krOoze: Well, it is semantically a different thing. Handle is not a pointer. Besides pointers are platform dependent things. Non-dispatchable Vulkan handles are 64-bit integral values.

include/vulkan/vulkan_core.h
krOoze: Personally, I love to be able to do `if( handle ) dosomething();`. So implicit conversion to **some** _integral type_ has to be there to allow it, right?

Ralith: Actually, [`explicit bool` works in that case](https://godbolt.org/g/tAhq9K). See [discussion of the problem](https://en.cppreference.com/w/cpp/language/implicit_conversion#The_safe_bool_problem).

krOoze: thanks, seems so, it is C++11 though. Not sure how much should I worry about C++98\03. I can probably guard the whole thing for C++11, and people using obsolete compilers can just get the C version...

krOoze: done

xml/vk.xml
```diff
@@ -154,16 +154,37 @@ server.
...
         <type category="define" name="VK_DEFINE_NON_DISPATCHABLE_HANDLE">
 #if !defined(VK_DEFINE_NON_DISPATCHABLE_HANDLE)
-#if defined(__LP64__) || defined(_WIN64) || (defined(__x86_64__) &amp;&amp; !defined(__ILP32__) ) || defined(_M_X64) || defined(__ia64) || defined (_M_IA64) || defined(__aarch64__) || defined(__powerpc64__)
+#if defined(__cplusplus) &amp;&amp; ( __cplusplus >= 201100 || _MSC_VER >= 1800)
+        #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) \
+            class object{ \
+            private: \
+                uint64_t h; \
+            public: \
+                object(){}; \
```
jeffbolznv: The Vulkan WG had some code very similar to this in the header before 1.0 was released, but we ultimately ditched it because it introduced too many problems.

One specific bug I recall is that this default constructor needs to initialize h to zero. This is the constructor that gets called when the handle is initialized by aggregate initialization, so leaving it uninitialized here breaks code that uses the following style to zero-init structures:

// renderPass and framebuffer would no longer be zero-initialized after this change
VkRenderPassBeginInfo beginInfo = { VK_STRUCTURE_TYPE_RENDER_PASS_BEGIN_INFO };

This specific problem is obviously fixable, but nonetheless I don't think the WG will want to make a change like this.

krOoze: Not so obviously fixable. That is if I don't want unnecessarily initialized handles.
That would mean I have to restrict it to aggregate type (forgo non-defaulted constructors), and restrict it to C++14 (that should force aggregate initialization, instead of value initialization... I think).

jeffbolznv:
Closing based on my previous feedback. We briefly touched on this in the WG meeting and I think folks generally agree that this is hard/fragile.

PR #781, opened 2018-08-28
Specify the units of buffer{RowLength,imageHeight}. Issue 752.

MarkCallow:


MarkCallow:
Note I have not tried building the spec. with this change. I expected that you would have CI set up to test build PRs. You don't. It would be a good idea to do so.

In this case, the change is simple enough that it is extremely unlikely to have broken the build.

Issue #790, opened 2018-09-07
`VkDebugUtilsMessengerCallbackDataEXT` array members


krOoze:
1) Why do `VkDebugUtilsMessengerCallbackDataEXT` array members have missing VUs (except `objectCount`)

2) Why are not the array/pointers `const`?

oddhack:
Will ask Mark to take a look at this, if he has time.

MarkY-LunarG:
Not sure why, but these had "noatuovalidity" in their tags so they didn't expand.  `const` isn't needed because the only places `VkDebugUtilsMessengerCallbackDataEXT` is used, it itself is a pointer and `const`.  If anything, the `const` in front of `pMessageIdName` is overkill.  I could remove that if necessary, but I'm not sure it's worth changing the interface, do you?

MarkY-LunarG:
I'm going to fix the "noautovalidity" issue in a PR shortly.

krOoze:
That array VU looks like the wrong type (intended for arrays that return `count`).

Per the `const` issue, I spam it everywhere, so I run into this issue when trying to assingn const variable into the non-const pointer, ergo compiler says nope.

MarkY-LunarG:
> That array VUs look like the wrong type (intended for arrays that return `count`).

That's actually the VUs automatically generated by just remove the "noautovalidity" flag on the XML.  I can manually add the VU statement.  It seems like we want this type of statement, though, any time the application passes in this struct, but not when we return it.

> Per the `const` issue, I spam it everywhere, so I run into this issue when trying to assingn const variable into the non-const pointer, ergo compiler says nope. PS: it would not be so api breaking change to add as const can be assigned non-const var.

My concern was mostly related to the loader and layers that implement this extension.  I'll have to look into those separately.  I also know AMD implements this, so they'd have to potentially change as well.

krOoze:
> That's actually the VUs automatically generated by just remove the "noautovalidity" flag on the XML. 

It's finicky, IIRC one has to get the `optional` attributes of both the array and the count correctly. The VU(I) then looks e.g. like this:

> If `dependencyCount` is not `0`, `pDependencies` **must** be a valid pointer to an array of `dependencyCount` valid `VkSubpassDependency` structures

The XML for above is:

```
<member optional="true"><type>uint32_t</type>       <name>dependencyCount</name></member>
<member len="dependencyCount">const <type>VkSubpassDependency</type>* <name>pDependencies</name></member>
```

[...]

Issue #789, opened 2018-09-07
`VkDebugUtilsObjectNameInfoEXT` overrestricted?


krOoze:
`VkDebugUtilsObjectNameInfoEXT` requires real `objectHandle`.
`VkDebugUtilsMessengerCallbackDataEXT` requires `objectCount > 0` (BTW `pObjects` array VU is missing &mdash; weird)

So when I do `vkSubmitDebugUtilsMessageEXT` I am required to provide valid object handle.
Considering the layers themselves do return `UNKNOWN` and `0` sometimes (and it even makes sense e.g. in creation functions, where there's no valid handle), isn't this overrestricted?

oddhack:
Will ask Mark to take a look at this, if he has time.

MarkY-LunarG:
`objectCount` and `pObjects` in `VkDebugUtilsMessengerCallbackDataEXT` should be optional (and able to be `0` and `NULL`).  I'll modify the XML.

As far as the objects themselves go, there were two cases to the `VkDebugUtilsObjectNameInfoEXT`.

1. For use in `vkSetDebugUtilsObjectNameEXT` we thought that passing in `UNKNOWN` and `NULL` information was invalid.
2. When returning the information to a callback from inside the layers, our original thought was that the layers would know what they were dealing with.  But, it still might not for certain extension and so `UNKNOWN` should still have been allowed for this case.

Because of this, I'll move the "can't be `UNKNOWN`" VU from the `VkDebugUtilsObjectNameInfoEXT` up to the `vkSetDebugUtilsObjectNameEXT` section, which is where it belongs.

krOoze:
Would it make sense to allow 0 Object count and keep object type VU?

MarkY-LunarG:
Here are the new sections:

![screenshot from 2018-09-10 12-25-09](https://user-images.githubusercontent.com/16328578/45316683-fb8ad600-b4f4-11e8-84fc-53cbe348ee33.png)

![screenshot from 2018-09-10 12-25-41](https://user-images.githubusercontent.com/16328578/45316692-ff1e5d00-b4f4-11e8-808e-584d3da3ec9f.png)

MarkY-LunarG:
I also added a VU to `vkSubmitDebugUtilsMessageEXT` for the same kind of restriction I put on `vkSetDebugUtilsObjectNameEXT`.

So, `vkSubmitDebugUtilsMessageEXT` now has the manual VU info:

```
[...]

PR #794, opened 2018-09-10
debug_utils: Fix issues discovered by @krOoze

MarkY-LunarG:
Fixing a few items discovered by @krOoze with the EXT_debug_utils
extension spec language validation.  Namely in two issues:

Issue #789:
 The problem was the manual VUs should have been applied to
 the `vkSetDebugUtilsObjectNameEXT` command and not the
 general `VkDebugUtilsObjectNameInfoEXT` structure.

Issue #790:
 Here, we needed to remove the "noautovalidity" tag on
 the XML and add the "optional" tag to `pObjects` and
 `objectCount`.

### DIFF
--- a/chapters/VK_EXT_debug_utils.txt
+++ b/chapters/VK_EXT_debug_utils.txt
@@ -42,6 +42,12 @@ include::../api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
     slink:VkDebugUtilsObjectNameInfoEXT structure specifying the parameters
     of the name to set on the object.
 
+.Valid Usage
+****
+  * pname:pNameInfo\->pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
+  * pname:pNameInfo\->pname:objectHandle must: not be dlink:VK_NULL_HANDLE
+****
+
 include::../validity/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
 --
@@ -68,13 +74,14 @@ removed.
 
 .Valid Usage
 ****
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectType-01905]]
-    pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectHandle-01906]]
+  * If pname:objectType is ename:VK_OBJECT_TYPE_UNKNOWN,
     pname:objectHandle must: not be dlink:VK_NULL_HANDLE
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectHandle-01907]]
-    pname:objectHandle must: be a Vulkan object of the type associated with
-    pname:objectType as defined in <<debugging-object-types>>.
+  * If pname:objectType is not ename:VK_OBJECT_TYPE_UNKNOWN,
+    pname:objectHandle must: be dlink:VK_NULL_HANDLE or
+    a valid Vulkan handle of the type associated with pname:objectType
+    as defined in the
+    <<debugging-object-types, VkObjectType and Vulkan Handle Relationship>>
+    table
 ****
 
 include::../validity/structs/VkDebugUtilsObjectNameInfoEXT.txt[]
@@ -134,11 +141,11 @@ be used by that implementation.
 ****
   * [[VUID-VkDebugUtilsObjectTagInfoEXT-objectType-01908]]
     pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
-  * [[VUID-VkDebugUtilsObjectTagInfoEXT-objectHandle-01909]]
-    pname:objectHandle must: not be dlink:VK_NULL_HANDLE
   * [[VUID-VkDebugUtilsObjectTagInfoEXT-objectHandle-01910]]
-    pname:objectHandle must: be a Vulkan object of the type associated with
-    pname:objectType as defined in <<debugging-object-types>>.
+    pname:objectHandle must: be a valid Vulkan handle of the type associated
+    with pname:objectType as defined in the
+    <<debugging-object-types, VkObjectType and Vulkan Handle Relationship>>
+    table
 ****
 
 include::../validity/structs/VkDebugUtilsObjectTagInfoEXT.txt[]
@@ -377,6 +384,10 @@ include::../api/protos/vkCreateDebugUtilsMessengerEXT.txt[]
     sname:VkDebugUtilsMessengerEXT object created.
 
 include::../validity/protos/vkCreateDebugUtilsMessengerEXT.txt[]
+
+The application must: ensure that flink:vkCreateDebugUtilsMessengerEXT is not
+executed in parallel with any Vulkan command that is also called with
+pname:instance or child of pname:instance as the dispatchable argument.
 
 --
 
@@ -648,7 +659,15 @@ The parameters are passed on to the callback in addition to the
 pname:pUserData value that was defined at the time the messenger was
 registered.
 
+.Valid Usage
+****
+  * pname:objectType member of each element of
+    pname:pCallbackData\->pname:pObjects must: not be
+    ename:VK_OBJECT_TYPE_UNKNOWN
+****
+
 include::../validity/protos/vkSubmitDebugUtilsMessageEXT.txt[]
+
 
 --
 
@@ -679,6 +698,10 @@ include::../api/protos/vkDestroyDebugUtilsMessengerEXT.txt[]
 ****
 
 include::../validity/protos/vkDestroyDebugUtilsMessengerEXT.txt[]
+
+The application must: ensure that flink:vkDestroyDebugUtilsMessengerEXT is not
+executed in parallel with any Vulkan command that is also called with
+pname:instance or child of pname:instance as the dispatchable argument.
 
 --
 
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -2998,11 +2998,11 @@ server.
             <member optional="true"><type>int32_t</type>                                                            <name>messageIdNumber</name></member>
             <member len="null-terminated">const <type>char</type>*                                                  <name>pMessage</name></member>
             <member optional="true"><type>uint32_t</type>                                                           <name>queueLabelCount</name></member>
-            <member noautovalidity="true" optional="true" len="queueLabelCount"><type>VkDebugUtilsLabelEXT</type>*  <name>pQueueLabels</name></member>
+            <member len="queueLabelCount">const <type>VkDebugUtilsLabelEXT</type>*                  <name>pQueueLabels</name></member>
             <member optional="true"><type>uint32_t</type>                                                           <name>cmdBufLabelCount</name></member>
-            <member noautovalidity="true" optional="true" len="cmdBufLabelCount"><type>VkDebugUtilsLabelEXT</type>* <name>pCmdBufLabels</name></member>
-            <member><type>uint32_t</type>                                                                           <name>objectCount</name></member>
-            <member noautovalidity="true" len="objectCount"><type>VkDebugUtilsObjectNameInfoEXT</type>*             <name>pObjects</name></member>
+            <member len="cmdBufLabelCount">const <type>VkDebugUtilsLabelEXT</type>*                 <name>pCmdBufLabels</name></member>
+            <member optional="true"><type>uint32_t</type>                                                           <name>objectCount</name></member>
+            <member len="objectCount">const <type>VkDebugUtilsObjectNameInfoEXT</type>*             <name>pObjects</name></member>
         </type>
         <type category="struct" name="VkImportMemoryHostPointerInfoEXT" structextends="VkMemoryAllocateInfo">
             <member values="VK_STRUCTURE_TYPE_IMPORT_MEMORY_HOST_POINTER_INFO_EXT"><type>VkStructureType</type> <name>sType</name></member>


chapters/VK_EXT_debug_utils.txt
```diff
@@ -44,6 +44,14 @@ include::../api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
 include::../validity/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
+.Valid Usage
+****
+  * pname:pNameInfo\->objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
```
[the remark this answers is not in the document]
MarkY-LunarG: If I change it, won't it break as soon as that change goes in?  Instead, if I leave it like this, it will just start working since that PR is likely going to go in soon, right?

chapters/VK_EXT_debug_utils.txt
```diff
@@ -647,8 +644,20 @@ The parameters are passed on to the callback in addition to the
 pname:pUserData value that was defined at the time the messenger was
 registered.
 
+.Valid Usage
+****
+* If pname:pCallbackData\->objectCount is greater than 0, then for
+  each object in pname:pCallbackData\->objects, the value of
```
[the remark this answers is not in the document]
MarkY-LunarG: :+1:

chapters/VK_EXT_debug_utils.txt
```diff
@@ -44,6 +44,14 @@ include::../api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
 include::../validity/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
+.Valid Usage
+****
+  * pname:pNameInfo\->objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
+  * pname:pNameInfo\->objectHandle must: not be dlink:VK_NULL_HANDLE
+  * pname:pNameInfo\->objectHandle must: be a Vulkan object of the type associated with
```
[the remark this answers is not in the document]
MarkY-LunarG: :+1:

chapters/VK_EXT_debug_utils.txt
```diff
@@ -44,6 +44,14 @@ include::../api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
 include::../validity/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
+.Valid Usage
+****
+  * pname:pNameInfo\->objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
+  * pname:pNameInfo\->objectHandle must: not be dlink:VK_NULL_HANDLE
```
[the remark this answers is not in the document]
MarkY-LunarG: Could be.  This was how it was with the Debug Marker version.  I could remove it since a valid handle seems to imply non-NULL.  But I didn't want to assume that without looking through the spec first.  Where I see VK_NULL_HANDLE being used, it does seem to be an independent value from "a valid handle of type XXX".  So, if we do state that it must be a valid handle, I think that's good enough.  :+1:

chapters/VK_EXT_debug_utils.txt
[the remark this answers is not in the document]
krOoze: That should be so. The PR is ancient though. After merge need to check for visible `pname`s anyway...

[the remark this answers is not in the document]
MarkY-LunarG: Okay, I'll switch it, then.

chapters/VK_EXT_debug_utils.txt
```diff
@@ -44,6 +44,14 @@ include::../api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
 include::../validity/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
+.Valid Usage
+****
+  * pname:pNameInfo\->pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
+  * pname:pNameInfo\->pname:objectHandle must: be a Vulkan handle of the type associated with
+    pname:pNameInfo\->pname:objectType as defined in
+    <<debugging-object-types, VkObjectType and Vulkan Handle Relationship>>.
```
krOoze: "in the ... table"?
\+ No period after VU

chapters/VK_EXT_debug_utils.txt
```diff
@@ -66,17 +74,6 @@ calling fname:vkSetDebugUtilsObjectNameEXT again with a new string.
...
 
-.Valid Usage
-****
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectType-01905]]
-    pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectHandle-01906]]
-    pname:objectHandle must: not be dlink:VK_NULL_HANDLE
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectHandle-01907]]
-    pname:objectHandle must: be a Vulkan object of the type associated with
-    pname:objectType as defined in <<debugging-object-types>>.
```
krOoze: Should still say "if not unknown type, must be valid handle". It might come up in some obscure case, e.g. `api_dump` layer after `unique_objects` layer -- i.e. the handle must be properly translatable without crashes.

chapters/VK_EXT_debug_utils.txt
```diff
@@ -66,17 +74,6 @@ calling fname:vkSetDebugUtilsObjectNameEXT again with a new string.
 If pname:pObjectName is an empty string, then any previously set name is
 removed.
 
-.Valid Usage
-****
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectType-01905]]
-    pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectHandle-01906]]
-    pname:objectHandle must: not be dlink:VK_NULL_HANDLE
```
krOoze: Might still make sense to ban Unknown+Null. The app  an just pass 0 object count for that case.

MarkY-LunarG: How about:

```
  * If pname:objectType is ename:VK_OBJECT_TYPE_UNKNOWN,
    pname:objectHandle must: not be dlink:VK_NULL_HANDLE
```

chapters/VK_EXT_debug_utils.txt
MarkY-LunarG: How about this:

```
  * If pname:objectType is not ename:VK_OBJECT_TYPE_UNKNOWN,
    pname:objectHandle must: be a Vulkan handle of the type associated with
    pname:objectType as defined in the
    <<debugging-object-types, VkObjectType and Vulkan Handle Relationship>>
    table
```
That would allow me to remove the VU statement we just added at `vkSetDebugUtilsObjectNameEXT` since that covers this case.  The only VU statement left for `vkSetDebugUtilsObjectNameEXT` would be this:

```
  * pname:pNameInfo\->pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
```

Since we don't want UNKNOWN there.

chapters/VK_EXT_debug_utils.txt
MarkY-LunarG: :+1:

chapters/VK_EXT_debug_utils.txt
krOoze: Wait I am starting to get lost. Let's recapitulate.

There are two users of `VkDebugUtilsObjectNameInfoEXT`:  `vkSetDebugUtilsObjectNameEXT` and `VkDebugUtilsMessengerCallbackDataEXT` (which in turn is used only by `vkSubmitDebugUtilsMessageEXT`; and by user's callback but that does not count towards VU).

For `vkSetDebugUtilsObjectNameEXT` we always want the and only valid type **and** valid handle.

For `vkSubmitDebugUtilsMessageEXT` I think we agreed it should be valid to provide:
- valid type + valid handle
- valid type + null handle

(
The invalid usage for `vkSubmitDebugUtilsMessageEXT` therefore should be:
- unknown type
- valid type + invalid non-null handle

)

That about right?

<hr />

So if we do GCD, then `VkDebugUtilsObjectNameInfoEXT` can have the
"must not be `UNKNOWN_TYPE`" VU,

The `vkSetDebugUtilsObjectNameEXT` then needs
"`object` must be valid `objectType` handle" VU.

And the `vkSubmitDebugUtilsMessageEXT` would need
"If `object` is not NULL, `object` must be valid `objectType` handle" VU.

(Which alternatively could be moved to `VkDebugUtilsObjectNameInfoEXT`, and then `vkSetDebugUtilsObjectNameEXT` would not need to note the handle type: "`object` must be a valid Vulkan handle")

chapters/VK_EXT_debug_utils.txt
krOoze: Actually that would ban exactly the case we wanted to allow (i.e. allow NULL_HANDLE object from `vkSubmitDebugUtilsMessageEXT`)?

krOoze: I think we agreed that we do not want `VK_OBJECT_TYPE_UNKNOWN` cases with valid handle, or did you reconsider?

MarkY-LunarG: No, I'm still leaning towards allowing it because the validation layers could return it.  If it's in a valid usage statement on the object, people might think the layers are behaving incorrectly.

In the case of an extension not supported by the layers, that adds a new type, it may call a supported command with the new type value which won't be known resulting in an error message with `UNKNOWN`.

chapters/VK_EXT_debug_utils.txt
MarkY-LunarG: We might have `VK_OBJECT_TYPE_UNKNOWN` type for `VkDebugUtilsObjectNameInfoEXT` when it's returned in the callback data if the layers fail to determine a handle in some cases.  This can be the case when an extension adds a type and isn't implemented in the layers (but is in the driver). 

So, I don't think "must not be UNKNOWN_TYPE" VU, belongs on `VkDebugUtilsObjectNameInfoEXT`.  Instead, I think it needs to be called out during the two commands.

This is the way I see it:

`VkDebugUtilsObjectNameInfoEXT`
  * If type == unknown, handle must not be NULL

`vkSubmitDebugUtilsMessageEXT`
  * For each object in pCallbackData->pObjects, type must: not be Unknown
  * For each object in pCallbackData->pObjects, if handle not NULL must be a valid handle

`vkSetDebugUtilsObjectNameEXT`
 * pNameInfo->objectType must: not be Unknown
 * pNameInfo->objectHandle must: be a valid handle

krOoze: Hmm, I never considered VUs to apply to stuff the Implementation returns (or calls calbacks with)...
I think returned values (and parameters from Implementation to the Applications callback) needs its own restrictions clerly addressed to the Implementers.

MarkY-LunarG: `vkSubmitDebugUtilsMessageEXT` is stricter because the application should know everything about the objects that underlying layers/ICDs may not.  That way, we force them to pass in useful data.

krOoze: Then again, why should we care what the app passes to its own callback? Especially if the callback must already be able to handle the unrestricted cases.

Besides "application" may be just another layer. layer 2 (some library) may not know much more than the layers and ICD does.

And it is also weird to me that the layers/ICDs would not know. Either they do not know a thing is a handle\object in the first place, or they do know the type, right?

chapters/VK_EXT_debug_utils.txt
```diff
@@ -44,6 +44,12 @@ include::../api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
 include::../validity/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
+.Valid Usage
+****
+  * pname:pNameInfo\->pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
+  * pname:pNameInfo\->pname:objectHandle must: not be dlink:VK_NULL_HANDLE
+****
```
krOoze: This VU block is in the wrong place. Should be above Implicit VU.

chapters/VK_EXT_debug_utils.txt
```diff
@@ -647,8 +654,15 @@ The parameters are passed on to the callback in addition to the
 pname:pUserData value that was defined at the time the messenger was
 registered.
 
+.Valid Usage
+****
+  * For each object in pname:pCallbackData\->pname:pObjects,
+    pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN.
```
krOoze: maybe
> pname:objectType member of each element of pname:pCallbackData\\->pname:pObjects must: not be ename:VK_OBJECT_TYPE_UNKNOWN

?
Also, no period for consistency.

MarkY-LunarG: :+1:

chapters/VK_EXT_debug_utils.txt
```diff
@@ -68,13 +74,14 @@ removed.
...
-    pname:objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectHandle-01906]]
+  * If pname:objectType is ename:VK_OBJECT_TYPE_UNKNOWN,
     pname:objectHandle must: not be dlink:VK_NULL_HANDLE
-  * [[VUID-VkDebugUtilsObjectNameInfoEXT-objectHandle-01907]]
-    pname:objectHandle must: be a Vulkan object of the type associated with
-    pname:objectType as defined in <<debugging-object-types>>.
+  * If pname:objectType is not ename:VK_OBJECT_TYPE_UNKNOWN,
+    pname:objectHandle must: be dlink:VK_NULL_HANDLE or
+    a Vulkan handle of the type associated with pname:objectType
```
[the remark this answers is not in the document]
MarkY-LunarG: :+1:

chapters/VK_EXT_debug_utils.txt
MarkY-LunarG: :+1:

xml/vk.xml
```diff
@@ -6213,20 +6213,20 @@ server.
         </command>
         <command successcodes="VK_SUCCESS" errorcodes="VK_ERROR_OUT_OF_HOST_MEMORY">
             <proto><type>VkResult</type> <name>vkCreateDebugUtilsMessengerEXT</name></proto>
-            <param><type>VkInstance</type> <name>instance</name></param>
+            <param externsync="true"><type>VkInstance</type> <name>instance</name></param>
```
krOoze: I think we would also need to add implicit externsync for every child of Instance. That seems unsustainable. There's also a subtle case that we probably only care about the dispatch parameters.

Maybe just add it as a freeform paragraph below `vkCreateDebugUtilsMessengerEXT`, e.g.:

> The application **must** ensure that `vkCreateDebugUtilsMessengerEXT` duration does not execute in parallel with duration of any Vulkan command that has been called with `instance` or child of `instance` as the dispatchable argument.

xml/vk.xml
```diff
@@ -6213,20 +6213,20 @@ server.
...
             <proto><type>void</type> <name>vkDestroyDebugUtilsMessengerEXT</name></proto>
-            <param><type>VkInstance</type> <name>instance</name></param>
+            <param externsync="true"><type>VkInstance</type> <name>instance</name></param>
             <param externsync="true"><type>VkDebugUtilsMessengerEXT</type> <name>messenger</name></param>
             <param optional="true">const <type>VkAllocationCallbacks</type>* <name>pAllocator</name></param>
         </command>
         <command>
             <proto><type>void</type> <name>vkSubmitDebugUtilsMessageEXT</name></proto>
-            <param><type>VkInstance</type> <name>instance</name></param>
+            <param externsync="true"><type>VkInstance</type> <name>instance</name></param>
```
krOoze: Not sure this needs `externsync`. The Impl probably does not need to modify instance here (only read the callback pointer, and get the filter parameters).

xml/vk.xml
MarkY-LunarG: I think you're statement is a little heavy handed.  "Child of `instance`" is basically any object.  I don't think it will impact anything if we just stated `instance`.

krOoze: > Child of instance" is basically any object.

Well that's intentional, no? Debug messengers affect any and all objects. E.g. command called with `device`, where that `device` is child of that `instance`.

For comparison wording of the same situation in `vkGetInstanceProcAddr`:

> The returned function pointer **must** only be called with a dispatchable object (the first parameter) that is `instance` or a child of `instance`, e.g. `VkInstance`, `VkPhysicalDevice`, `VkDevice`, `VkQueue`, or `VkCommandBuffer`.

MarkY-LunarG: Ok, I don't think it's going to matter that much, but better safe than sorry.  I did clean up your wording a little, too.

krOoze:
BTW, I have gone ahead and posted https://github.com/KhronosGroup/Vulkan-Loader/issues/86. May I assume making the `objectCount` `optional` is intentional?

chapters/debugging.txt
```diff
@@ -68,9 +68,6 @@ ifdef::VK_NVX_device_generated_commands[]
 | ename:VK_OBJECT_TYPE_OBJECT_TABLE_NVX | slink:VkObjectTableNVX
 | ename:VK_OBJECT_TYPE_INDIRECT_COMMANDS_LAYOUT_NVX | slink:VkIndirectCommandsLayoutNVX
 endif::VK_NVX_device_generated_commands[]
-ifdef::VK_EXT_debug_utils[]
-| ename:VK_OBJECT_TYPE_DEBUG_UTILS_MESSENGER_EXT | slink:VkDebugUtilsMessengerEXT
-endif::VK_EXT_debug_utils[]
```
krOoze: Why is this removed?

chapters/VK_EXT_debug_utils.txt
```diff
@@ -378,6 +385,14 @@ include::../api/protos/vkCreateDebugUtilsMessengerEXT.txt[]
 
 include::../validity/protos/vkCreateDebugUtilsMessengerEXT.txt[]
 
+[NOTE]
+.Note
+====
```
krOoze: Why as a Note? Specifying language should be outside a note IMO. Notes should only reiterate information that is already there but easy to miss from a formal description.

chapters/debugging.txt
MarkY-LunarG: You told me to remove it since you were commiting this change in another submission.

krOoze: Right. Weird it shows in the diff then.
The change was merged, so I think you should rebase, just to be safe.

MarkY-LunarG: Weird, looks like it got messed up even after a rebase.  I fixed it.

PR #800, opened 2018-09-13
Fix #788: make clear messageTypes is a bitmask + name fix

### LINKED ISSUE #788: Ambiguous PFN_vkDebugUtilsMessengerCallbackEXT type parameter
In `PFN_vkDebugUtilsMessengerCallbackEXT` type parameter is declared as:
`VkDebugUtilsMessageTypeFlagsEXT messageType`,
but in the parameter descriptions it says:
> `messageTypes` [sic] specifies the `VkDebugUtilsMessageTypeFlagBitsEXT` that triggered this callback.

So, is it a bitmask, or a single bit?

PS: `vkSubmitDebugUtilsMessageEXT` clearly has bitmask, so that is probably the correct case.

krOoze:
fixes #788 

Decided to change the name of the callback parameter in the XML. The name of a parameter should not matter in a function pointer, I think.


oddhack:
@mikes-lunarg could we get your feedback on this and hopefully sign it off?

oddhack:
@mikes-lunarg could we get your feedback on this and hopefully sign it off?

PR #810, opened 2018-09-27
Added missing pool implicitexternsyncparams

### LINKED ISSUE #802: External synchronization on command buffer reset
The specification states in the 'Threading Behavior' section:
```
In addition, there are some implicit parameters that need to be externally synchronized. For
example, all commandBuffer parameters that need to be externally synchronized imply that the
commandPool that was passed in when creating that command buffer also needs to be externally
synchronized.
```
 The following list of implicitly synchronized parameters doesn't contain the pool when reseting command buffers via `vkResetCommandBuffer` even though the `commandBuffer` parameter there needs to be externally synchronized.

Is this an oversight or intentional as the text implies it would also include this function.

krOoze:
Added some `implicitexternsyncparams` that I feel were missing.

Also fixes #802

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -6665,6 +6665,9 @@ typedef void <name>CAMetalLayer</name>;
             <param><type>VkDevice</type> <name>device</name></param>
             <param optional="true" externsync="true"><type>VkDescriptorPool</type> <name>descriptorPool</name></param>
             <param optional="true">const <type>VkAllocationCallbacks</type>* <name>pAllocator</name></param>
+            <implicitexternsyncparams>
+                <param>all sname:VkDescriptorSet objects allocated from pname:descriptorPool</param>
+            </implicitexternsyncparams>
         </command>
         <command successcodes="VK_SUCCESS">
             <proto><type>VkResult</type> <name>vkResetDescriptorPool</name></proto>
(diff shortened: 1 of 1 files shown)


xml/vk.xml
```diff
@@ -5129,6 +5135,10 @@ server.
             <param len="descriptorWriteCount" externsync="pDescriptorWrites[].dstSet">const <type>VkWriteDescriptorSet</type>* <name>pDescriptorWrites</name></param>
             <param optional="true"><type>uint32_t</type> <name>descriptorCopyCount</name></param>
             <param len="descriptorCopyCount" externsync="pDescriptorCopies[].dstSet">const <type>VkCopyDescriptorSet</type>* <name>pDescriptorCopies</name></param>
+            <implicitexternsyncparams>
+                <param>the sname:VkDescriptorPool that the pname:dstSet member of each pname:pDescriptorWrites element was allocated from</param>
+                <param>the sname:VkDescriptorPool that the pname:dstSet member of each pname:pDescriptorCopies element was allocated from</param>
+            </implicitexternsyncparams>
```
jeffbolznv: I don't think these two are needed. The descriptor sets have already been allocated and these should just be writing to their allocations, which doesn't require synchronizing the pool.

And, practically speaking, if we made descriptor updates externsync on a common object it would make multithreading significantly harder.

xml/vk.xml
```diff
@@ -6207,6 +6226,9 @@ server.
             <param externsync="true"><type>VkDescriptorSet</type> <name>descriptorSet</name></param>
             <param><type>VkDescriptorUpdateTemplate</type> <name>descriptorUpdateTemplate</name></param>
             <param noautovalidity="true">const <type>void</type>* <name>pData</name></param>
+            <implicitexternsyncparams>
+                <param>the sname:VkDescriptorPool that pname:descriptorSet was allocated from</param>
+            </implicitexternsyncparams>
```
jeffbolznv: Similar comment to vkUpdateDescriptorsets.

Other than these two, the rest look fine to me.

xml/vk.xml
```diff
@@ -1851,6 +1851,9 @@ server.
             <member optional="true"><type>VkDeviceSize</type>                             <name>sequencesCountOffset</name></member>
             <member optional="true"><type>VkBuffer</type>                                 <name>sequencesIndexBuffer</name></member>
             <member optional="true"><type>VkDeviceSize</type>                             <name>sequencesIndexOffset</name></member>
+            <implicitexternsyncparams>
+                <param>the sname:VkCommandPool that pname:targetCommandBuffer was allocated from</param>
+            </implicitexternsyncparams>
```
krOoze: Not sure this makes sense either. The `targetCommandBuffer` is supposed to already have a space reserved.

xml/vk.xml
krOoze: Agreed. It returns `void`, which probably means it cannot ask the pool for additional memory.

krOoze:
@jeffbolznv Anything else to be done here?

krOoze:
Rebased. Pls reviw 😿

PR #829, opened 2018-10-17
Fix public issue 763

MarkCallow:


oddhack:
@MarkCallow we have a broader terminology fix underway in internal MR 2881, which you should be able to see and comment on. We might fold elements of this PR in, such as the introductory paragraph in the VkFormat section, but the group's discussion ended up with wanting to call "data element" -> "texel block", and "element size" -> "texel block size".

Issue #832, opened 2018-10-22
Ambiguous Unsupported Limit


krOoze:
In the Table 48. Required Limits Unsupported Limit column, it mostly seems to me that it lists the exact value the limit member should have when the feature is unsupported.

But the text strongly implies the limit is either `min` or `max` type.

E.g. for `maxViewports` value `1` it is ambiguous whether if the feature is unsupported the value "must be 1" or the value "must be greater or equal to 1".

Furthermore some Unsupported Limit values seem to be missing (`-`). E.g. `primitiveOverestimationSize`, and I am unsure what that means (can be any value??).

TomOlson:
We've just started looking at this. Agreed that there are a number of inconsistencies and quirks in the required limits table. (The fact that the table has nine footnotes may be an indication that we should rethink the design.) There are multiple problems here, affecting spec language written by many different people, so I'd like to take this back to the working group for further discussion. Because changes here would impact a lot of language, it's likely to take some time to find fixes that work for every case.

BTW it looks like you tried to address some of these problems in PR #833, which is officially about html table formatting. I'll comment separately over there, but in general it's better to split concerns into independent issues and PRs - that makes them easier and faster to review.

PR #833, opened 2018-10-23
Try to fix html page overflow

krOoze:
Prevents showing slider (tested on 16:10 display), but still many things extends into the bounds (probably ok enough).

### DIFF
--- a/chapters/features.txt
+++ b/chapters/features.txt
@@ -7632,7 +7632,7 @@ following table:
 [[external-memory-handle-types-compatibility]]
 .External memory handle types compatibility
 |====
-| Handle type | sname:VkPhysicalDeviceIDProperties::pname:driverUUID | sname:VkPhysicalDeviceIDProperties::pname:deviceUUID
+| Handle type | sname:VkPhysicalDeviceIDProperties{wbro}::pname:driverUUID | sname:VkPhysicalDeviceIDProperties{wbro}::pname:deviceUUID
 | ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT | Must match | Must match
 | ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT | Must match | Must match
 | ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_KMT_BIT | Must match | Must match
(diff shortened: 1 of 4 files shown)


krOoze:
PS: also fixes Ghostscript error:
```
   **** Error reading a content stream. The page may be incomplete.
               Output may be incorrect.
```

caused by the long `textures.txt` equation. The equation is silently missing in PDF when using updated tooling. And probably causes some corruption on which ghostscript then trips.

krOoze:
All of those should reduce table width.

"External memory handle types compatibility" table overflows page.
I think the rest only overflows container (and extend into the borders), which, I guess, can be lived with...

TomOlson:
We'd like to support the goal of this PR, but we'd prefer to separate pure formatting concerns from the table content and semantics issues you raise in #832. Probably best to let us address that issue and settle what the table should say, before trying to figure out how to format it. You could if you like separate out changes that don't affect semantics, such as mapping "implementation dependent" to "any" in the limits table. However, we may end up getting rid of the "type" field, so not sure it is worth the effort.

chapters/features.txt
```diff
@@ -7596,24 +7598,31 @@ following table:
...
-| ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP_BIT | Must match | Must match
-| ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE_BIT | Must match | Must match
+| Handle Type +
+  etext:VK_EXTERNAL_MEMORY_HANDLE_TYPE_*
+
+| sname:VkPhysicalDeviceIDProperties::{zwsp}pname:driverUUID
+| sname:VkPhysicalDeviceIDProperties::{zwsp}pname:deviceUUID
+
+
+| etext:OPAQUE_FD_BIT | Must match | Must match
```
TomOlson: I appreciate what you're trying to do, but replacing enames with etext shorthand makes me nervous (and violates the style guide). It can't be validated, and searching for the full enumerant name won't find it.  Yes, some of the names are very long, but we'd like to find another way to address the problem.

krOoze: Ok. It should end up in `<code>`, so I may attempt a CSS solution.
The PDF does something like that and line-breaks even withing the keyword, if it has to.

I was not happy with it either for searchability sake. `*text` macros though seem for exactly this purpose &mdash; formating as `*name` things that are for various reasons not Vulkan entities (e.g. wildcarded names).

chapters/primsrast.txt
```diff
@@ -1970,17 +1970,15 @@ The bias value [eq]#o# for a polygon is
 
 [latexmath]
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-o =
+\begin{aligned}
+o = \operatorname{dbclamp}( m \times \mathtt{depthBiasSlopeFactor} + r \times \mathtt{depthBiasConstantFactor} ) \\
```
krOoze: Causes Parse Error, to be fixed momentarily.

PR #839, opened 2018-10-28
Restore support for regular manpage generation

Ryp:
* Rename make rules to remove ambiguous meaning, namely manpages and
manhtmlpages.
* The support is not full yet, there is still html specific constructs that
end up in the man.

This is not shipping-quality yet, but it does produce usable man pages already. If someone would spearhead the support of this useful format, that would be great. Otherwise I can try to find some spare time and modify the custom markup generation tools to have perfect man pages.

I accompagny this PR with a [usable AUR package](https://aur.archlinux.org/packages/vulkan-man-git/) that I will point to the upstream repo if my PR gets merged.

### DIFF
--- a/BUILD.adoc
+++ b/BUILD.adoc
@@ -28,8 +28,8 @@ builds an HTML5 specification output.
 
     $ make all
 
-builds the spec targets `html`, `pdf`, `styleguide`, `manhtml`, `manpdf`,
-`manhtmlpages`, `checkinc`, and `checklinks`.
+builds the spec targets `html`, `pdf`, `styleguide`, `manpages`,
+`manhtml`, `manpdf`, `manhtmlpages`, `checkinc`, and `checklinks`.
 
 [NOTE]
 .Notes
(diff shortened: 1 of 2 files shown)


oddhack:
I think the best path is for you to host these changes externally in a fork of this repository, and feed manpage packages from that to the distributions wanting them. We've discussed this previously and don't want to maintain this format ourselves, or spend time on toolchain issues like the math path. Enabling this path in a Khronos-hosted repo would lead people to think we'll support them going forward.

This is exactly the scenario we wanted to enable with the CC-BY licensing scheme - letting other people support output formats we don't want to support ourselves.

Ryp:
Is there a closed issue discussing the reasons or was that private? I find it interesting that the man page style is still used across the API documentation even though the real format support is dropped.

I intend to at least support the bare minimum for these manpages to still be supported, but I don't want to be drowned in the noise if people can't see manpage support on the official repo. Would it be possible mention manpage support as unofficial with a link to my fork from the official docs?

oddhack:
There was some internal discussion 2+ years ago, mostly motivated by being unable to provide a fully functional set of ref pages due to the math rendering issue, and LunarG observing that they'd had all of 1 request for the .3 format pages in their SDK distribution. We kept the target in the Makefile but unused for a while IIRC, and at some point stripped it out.

I think we'd be happy to provide a link from the registry index page and the toplevel README or build instructions.

Ryp:
Fair enough, I'm closing the PR then. I'll be in touch for the rest. Thanks!

PR #840, opened 2018-10-31
Fix math markup in Anisotropic Filtering ch.

krOoze:


chapters/textures.txt
```diff
@@ -2472,8 +2472,8 @@ to ename:VK_SAMPLER_REDUCTION_MODE_WEIGHTED_AVERAGE_EXT, the above summation
 is used.
 If the reduction mode is ename:VK_SAMPLER_REDUCTION_MODE_MIN_EXT or
 ename:VK_SAMPLER_REDUCTION_MODE_MAX_EXT, then the value is instead computed
-as [eq]#\tau_{2Daniso} = reduce(\tau_1, ..., \tau_N)#, combining all texel
-values with non-zero weights.
+as latexmath:[\tau_{2Daniso} = \mathrm{reduce}(\tau_1, ..., \tau_N)],
```
oddhack: Needs fixing, but please fix it as [eq] math that's marked up properly instead of latexmath, e.g. {tau}\~2Daniso~ = reduce({tau}\~1~ ..., etc.

krOoze: OK. BTW what is the reasoning for having two types of math markup?

krOoze: done

oddhack: We try to only use latexmath markup when the constructs are beyond the scope of straightforward asciidoc markup. Hopefully, eventually there will be asciimath markup powerful enough to express the conditional equations, matrices, etc. that are hard to handle in straight asciidoc.

PR #852, opened 2018-11-12
Fix vkCmdClearDepthStencilImage::aspectMask VU

krOoze:
The original VUs said "must: each only include X". That makes the two VUs contradictory for depth+stencil format.

New VUs should allow `DEPTH` (for depth, or depth+stencil format). `STENCIL` (for stencil,  or depth+stencil format), or `DEPTH | STENCIL` (for depth+stencil format).

### DIFF
--- a/chapters/clears.txt
+++ b/chapters/clears.txt
@@ -162,16 +162,19 @@ endif::VK_VERSION_1_1,VK_KHR_maintenance1[]
     pname:imageLayout must: be either of
     ename:VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL or
     ename:VK_IMAGE_LAYOUT_GENERAL
+  * The slink:VkImageSubresourceRange::pname:aspectMask member of each
+    element of the pname:pRanges array must: not include bits other than
+    ename:VK_IMAGE_ASPECT_DEPTH_BIT or ename:VK_IMAGE_ASPECT_STENCIL_BIT
   * [[VUID-vkCmdClearDepthStencilImage-aspectMask-02499]]
-    The slink:VkImageSubresourceRange::pname:aspectMask members of the
-    elements of the pname:pRanges array must: each only include
-    ename:VK_IMAGE_ASPECT_DEPTH_BIT if the image format has a depth
-    component
+    If the pname:image's format does not have a stencil component, then the
+    slink:VkImageSubresourceRange::pname:aspectMask member of each element
+    of the pname:pRanges array must: not include the
+    ename:VK_IMAGE_ASPECT_STENCIL_BIT bit
   * [[VUID-vkCmdClearDepthStencilImage-aspectMask-02500]]
-    The slink:VkImageSubresourceRange::pname:aspectMask members of the
-    elements of the pname:pRanges array must: each only include
-    ename:VK_IMAGE_ASPECT_STENCIL_BIT if the image format has a stencil
-    component
+    If the pname:image's format does not have a depth component, then the
+    slink:VkImageSubresourceRange::pname:aspectMask member of each element
+    of the pname:pRanges array must: not include the
+    ename:VK_IMAGE_ASPECT_DEPTH_BIT bit
   * [[VUID-vkCmdClearDepthStencilImage-baseMipLevel-01474]]
     The slink:VkImageSubresourceRange::pname:baseMipLevel members of the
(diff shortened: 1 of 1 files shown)


chapters/clears.txt
```diff
@@ -162,16 +164,8 @@ endif::VK_VERSION_1_1,VK_KHR_maintenance1[]
...
-  * [[VUID-vkCmdClearDepthStencilImage-aspectMask-02499]]
-    The slink:VkImageSubresourceRange::pname:aspectMask members of the
-    elements of the pname:pRanges array must: each only include
-    ename:VK_IMAGE_ASPECT_DEPTH_BIT if the image format has a depth
-    component
-  * [[VUID-vkCmdClearDepthStencilImage-aspectMask-02500]]
-    The slink:VkImageSubresourceRange::pname:aspectMask members of the
-    elements of the pname:pRanges array must: each only include
-    ename:VK_IMAGE_ASPECT_STENCIL_BIT if the image format has a stencil
-    component
```
Tobski: So I think this is a case of unclear english language usage - it's meant to be like "only include these ASPECT bits if there's the right component" as opposed to "include only this ASPECT bit if the component matches". Language.

I think this can be clarified by just flipping the existing text condition around, so that it says "must: not include <X> if the image format does not have a <Y> component. I.e.:

```
  * [[VUID-vkCmdClearDepthStencilImage-aspectMask-02499]]
    The slink:VkImageSubresourceRange::pname:aspectMask members of the
    elements of the pname:pRanges array must: not include
    ename:VK_IMAGE_ASPECT_DEPTH_BIT if the image format does not have a
    depth component
  * [[VUID-vkCmdClearDepthStencilImage-aspectMask-02500]]
    The slink:VkImageSubresourceRange::pname:aspectMask members of the
    elements of the pname:pRanges array must: not include
    ename:VK_IMAGE_ASPECT_STENCIL_BIT if the image format does not have a
    stencil component
```

This retains the original meaning/VUIDs/precision of the existing VUs.

chapters/clears.txt
```diff
@@ -154,6 +154,8 @@ endif::VK_VERSION_1_1,VK_KHR_maintenance1[]
   * [[VUID-vkCmdClearDepthStencilImage-image-00010]]
     If pname:image is non-sparse then it must: be bound completely and
     contiguously to a single sname:VkDeviceMemory object
+  * [[VUID-vkCmdClearDepthStencilImage-image-00014]]
+    pname:image must: have a depth/stencil format
```
Tobski: I'm not clear why you moved this VU? Just seems like churn to me?

chapters/clears.txt
krOoze: Hmm, what about all the other aspects?

chapters/clears.txt
krOoze: Probably esthetics. Seems better with the other `image` VUs, and before all the longwinded VUs...

chapters/clears.txt
Tobski: This command is already restricted to depth/stencil formats, which takes care of the rest.

chapters/clears.txt
Tobski: In general I'd rather not move individual VUs around outside of a general "tidy up" pass, because it just means more chance for conflicts with other in-flight proposals.

chapters/clears.txt
krOoze: yes, but not to aspects.

Tobski: Oh I see. I think then to remain precise we just add:

```
  * slink:VkImageSubresourceRange::pname:aspectMask must: not include bits other than ename:VK_IMAGE_ASPECT_STENCIL_BIT or ename:VK_IMAGE_ASPECT_DEPTH_BIT.
```

krOoze: Are you sure none of these new aspects apply here? I am not familiar with them much...

krOoze: Right, if you prefer.

chapters/clears.txt
krOoze: As you wish. Though having barriers against tidying usually means it is virtually never done...

chapters/clears.txt
krOoze: done

krOoze:
@Tobski So what remains to be done here?

oddhack:
We'll accept this and then I'll strip the VUIDs.

PR #879, opened 2018-12-13
Fix links in PDF build

### LINKED ISSUE #877: Links in PDF specs not working
In https://github.com/KhronosGroup/Vulkan-Web-Registry/commit/ddf69d4a2215e16be78c9bf7ac4bd10bf58eadcd#commitcomment-31641862, @michaelforney noticed that links from the PDF documents weren't resolving, which I've verified.

I updated to Ruby 2.5.3 and slightly newer versions of some of the Ruby packages out of necessity due to a deep DSO versioning conflict after updating my Debian system (one that actually made it impossible to even do a 'gem update' because it was unhappy about the installed version of the ssl/crypto libraries), and can't easily back off to the older Ruby version / components I was using. That  will make diagnosing this difficult, but I'll try to figure it out - assistance welcome. The asciidoctor-pdf toolchain is a giant black box from my POV so there's a limited amount of diagnosis available.

krOoze:
fixes #877 

Not quite sure what the flag is supposed to be doing. Based on https://bugs.ghostscript.com/show_bug.cgi?id=699830.

Alternatively downgrade ghostscript to <= 9.23.

### DIFF
--- a/config/optimize-pdf
+++ b/config/optimize-pdf
@@ -72,6 +72,7 @@ ERRFILE=ERRS.optimize
 "$GS" -q -dNOPAUSE -dBATCH -dSAFER -dNOOUTERSAVE \
   -sDEVICE=pdfwrite \
   -dPDFSETTINGS=/prepress \
+  -dPrinted=false \
   -dCannotEmbedFontPolicy=/Warning \
   -dDownsampleColorImages=$DOWNSAMPLE_IMAGES \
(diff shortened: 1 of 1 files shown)


oddhack:
I have no idea how you tracked that down, but thanks - seems to be fine now.

krOoze:
@oddhack Actually quite easy, I built it first without `gs` (coz it is faster), and the links just worked...

Issue #876, opened 2018-12-11
Diagrams for texture coordinates are flipped vertically


francoisbertelatschrodinger:
According to "15.1.1. Texel Coordinate Systems" (spec 1.1.95),
* Normalized texel coordinates:
The t coordinate goes from 0.0 to 1.0, top to bottom.
* Unnormalized texel coordinates:
The v coordinate goes from -1.0 to 5.0, top to bottom. 
* Integer texel coordinates:
The j coordinate goes from -1 to 5, top to bottom.

However figure 3 "Texel Coordinate Systems" ( images/vulkantexture0.svg ), figure 4 "Texel Coordinate Systems" ( images/vulkantexture1.svg) and figure 11 "Implicit Derivatives" (in 15.5 Derivatives Operations, images/vulkantexture2.svg) show a vertical axis going from bottom to top.

oddhack:
Appears to be related to #898?

francoisbertelatschrodinger:
It is! However the commit 31b0c98a504eb3f9dc5a6f7bcf9a130fb09cb376 in #898 is only touching images/vulkantexture0.svg and images/vulkantexture1.svg.
It is missing images/vulkantexture2.svg for figure 11 "Implicit Derivatives".

oddhack:
We just pushed an update that included a bunch of changes to diagrams in this chapter. @francoisbertelatschrodinger could you take a look and see if this issue has been addressed now?

francoisbertelatschrodinger:
Should I look at a specific branch? In the master branch  images/vulkantexture0.svg, images/vulkantexture1.svg and images/vulkantexture2.svg are still from "10 month ago" when browsing through github. And the current pdf (Version 1.1.99, 2019-02-04 09:27:01Z) at https://www.khronos.org/registry/vulkan/specs/1.1-extensions/pdf/vkspec.pdf still have the vertical axes going from bottom to top instead of top to bottom.

oddhack:
I misspoke about the nature of the changes. It's the body text that should be different now, for example it describes "The t coordinate goes from 0.0 to 1.0" without talking about top/bottom. Hopefully that's improved consistency between text and diagrams to the point the arrows are not a problem?

francoisbertelatschrodinger:
I see. Yes it looks consistent now. The spec is orientation agnostic then, right?

oddhack:
[...]

PR #898, opened 2019-01-18
Flip arrow in figures of Texel Coordinate Systems

axinging:
The "15.1.1. Texel Coordinate Systems" says:
For s, u, i, "left to right".
For t, v, j, "top to bottom".
So should flip the arrow direction in both Figure 3. Texel Coordinate Systems
and Figure 4. Texel Coordinate Systems.

Original link:
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#_image_operations_overview

### DIFF
--- a/images/vulkantexture0.svg
+++ b/images/vulkantexture0.svg
@@ -14,7 +14,7 @@
    viewBox="0 0 186.82375 101.62997"
    version="1.1"
    id="svg8"
-   inkscape:version="0.92.2 (5c3e80d, 2017-08-06)"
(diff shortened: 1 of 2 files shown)


oddhack:
We just pushed an update that included a bunch of changes to diagrams and descriptions in this chapter. @axinging  could you take a look and see if this has been addressed now?

oddhack:
@axinging we think this has been addressed now, not by removing the arrows but by changing the body text to avoid the orientation language, so closing this. If you disagree then please comment on #876, or open a new issue.

PR #900, opened 2019-01-22
Fix some markup

krOoze:


chapters/drawing.txt
```diff
@@ -776,10 +776,9 @@ The assembled primitives execute the bound graphics pipeline.
     If the sname:VkPipeline object bound to
     ename:VK_PIPELINE_BIND_POINT_GRAPHICS requires any dynamic state, that
     state must: have been set on the current command buffer
-  * [[VUID-vkCmdDrawIndexed-indexSize-00463]]
```
krOoze: Should I keep this VUID? It contains the `indexSize` name which is not a API thing.

chapters/VK_NVX_device_generated_commands/generation.txt
```diff
@@ -116,42 +116,42 @@ include::../../api/structs/VkCmdProcessCommandsInfoNVX.txt[]
...
+    ename:VK_BUFFER_USAGE_INDIRECT_BUFFER_BIT bit set
   * [[VUID-VkCmdProcessCommandsInfoNVX-sequencesCountBuffer-01339]]
     If pname:sequencesCountBuffer is used, pname:sequencesCountOffset must:
     be aligned to
-    sname:VkDeviceGeneratedCommandsLimitsNVX::pname:minSequenceCountBufferOffsetAlignment.
+    sname:VkDeviceGeneratedCommandsLimitsNVX::pname:minSequenceCountBufferOffsetAlignment
   * [[VUID-VkCmdProcessCommandsInfoNVX-sequencesIndexBuffer-01340]]
     If pname:sequencesIndexBuffer is used, its usage flag must: have
-    ename:VK_BUFFER_USAGE_INDIRECT_BUFFER_BIT bit set.
+    ename:VK_BUFFER_USAGE_INDIRECT_BUFFER_BIT bit set
```
oddhack: Suggest 'must: have *the* ename:blah_BIT bit set'.

chapters/drawing.txt
oddhack: It's harmless to leave so long as the VU statement itself hasn't changed. The script just picks off the first pname: and uses it in the VUID, which is obviously not 100% reliable. Sometimes it can't find one and uses 'None', which is also OK.

krOoze: OK, then. The VU statement did not change meaning, only markup.

krOoze:
@oddhack I am adding `pname` to parameters inside a link. The `code` tag color overrides link color the user expect so in effect the parameters do not look clickable. So I want CSS to use `link` color for `code` (the items should still be distinguished by the monospace font inside a link).

krOoze:
@oddhack Removed from this PR.

PR #901, opened 2019-01-22
 Make markup inside link look like a link in CSS 

krOoze:
Currently the `code` tags (which `pname` and other markup is translated to) override the color of a link if the are part of one. In effect that makes the links look unclickable (especially if there is no non-markup text in the link too).

So I make the markup tag inside a link inherit link's color, instead using its own. The markup text should still be distinguished by its monospace font...


oddhack:
Closing since all changes were done via equivalent internal MRs.

PR #907, opened 2019-01-28
Remove readme.txt reference

zyrolasting:
I think this is referring to `../registry.txt`, and that's covered in the root README.


zyrolasting:
Force push was to change author email.

PR #912, opened 2019-02-04
Fix VK_IMG_filter_cubic aliases

krOoze:


BeRo1985:
I also noticed the issue when I tried to update the Vulkan pascal header of PasVulkan to 1.x.99.

karl-lunarg:
This breaks the validation layer compilation as well.  @oddhack: Can this make it into the next release?

oddhack:
I have a slightly different internal fix I'll be putting in the next update, so closing this. Thanks!

oddhack:
@karl-lunarg how expensive would it be to run an LVL build as part of the gitlab Vulkan spec repo CI? Seems like that might be a viable way to minimize this type of problem in the future - the XML was all perfectly valid and the header was legal C code, so those sanity checks didn't catch this.

krOoze:
@oddhack Well, technically valid, but the `#define` was shadowing the enum value. And the `define` has wrong type (which is problem for templates).

Probably should not have survived review. This is pretty much what I have seen quick-scrolling through changelog. People, I think, expect higher quality out of `EXT`, so I hope more caution to avoid this can be made in the future...

BTW, I didn't care to check, but I think some of the VUs are off too, especially in other extensions that still use the old `VK_FILTER_CUBIC_IMG` enumerant. They seem not aware of the newly added queries.

karl-lunarg:
Thanks to @krOoze for the fix and to @oddhack for effectively applying it.  This is going to save a lot of grief next week.

@oddhack I tend to agree with @krOoze.  I would add the specific suggestion that as part of the review, the generated header files be compared before and after the change to be sure that the registry change has the intended effect on the headers.  Should be a checklist item, if not already.

Using the V-VL repo as a compile check in GitLab CI is certainly attractive because V-VL has a lot of codgen that generates code that uses pretty close to all the symbols in the headers.  I have a lot of detailed thoughts that I'll post over in GitLab, since we are talking about GitLab CI here.

oddhack:
If it's not enforced by CI, it's inevitably going to fall through the cracks now and then, as happened here. More variety of automated checks is better, so looking forward to your thoughts.

krOoze:
That's true, but CI and automated testing is no replacement for discipline. If a problem I see in 10 seconds can seep in, how can I trust any of the more complex and subtle additions to Vulkan (which are fundamentally untestable).

There's also #664 addressing this particular problem (separates `enum` from define constants in xml).

PR #915, opened 2019-02-08
Correct and complete protected queue creation VUs

krOoze:
1) Let the non-unique family queues in the device creation list share the queue count limit
2) Prevent creation of `PROTECTED` queue from a family that does not support `PROTECTED`
3) Fix preprocessor inside a VU breaking json script + reduce `ifdef`ery by wording in 1.0 compatible manner
4) Prevent creation of `PROTECTED` queue from when `protectedMemory` feature is not being enabled
5) Allow `vkGetDeviceQueue` to fetch any queue that had `flags` set to `0`, and remove ambiguity if this function is valid if `pQueueCreateInfos` contained both protected and non-protected version of the same queue family


krOoze:
@oddhack @Tobski Rebased. Can I has reviews pls.

PR #916, opened 2019-02-10
Make WARNING into NOTE (per styleguide)

krOoze:
also PR #731 requires to only the Note icon be used...


oddhack:
Note that this also was moved around due to an unrelated internal MR.

PR #925, opened 2019-02-20
1.0 vk khr android surface

thananop1987:
hhujjjjjjjjj


oddhack:
What's the purpose of this PR? The 1.0-* branches have been obsolete for a very long time. If there's a change you're proposing, it need to be done as a PR based on the current state of master branch.

PR #927, opened 2019-02-20
Memory Model: Add acyclicity (consistency) axiom

jeffbolznv:
Add acyclicity (consistency) axiom and rephrase Scoped Modification Order Coherence section as a consequence of it.

For https://github.com/KhronosGroup/Vulkan-MemoryModel/issues/13

### DIFF
--- a/appendices/memorymodel.txt
+++ b/appendices/memorymodel.txt
@@ -788,21 +788,53 @@ NOTE: It is possible for there to be a write between X and Y that overwrites
 a subset of the memory locations, but the remaining memory locations (M~2~)
 will still be visible-to Y.
 
+[[memory-model-acyclicity]]
+== Acyclicity
+
+_Reads-from_ is a relation between operations, where the first operation is a
+write, the second operation is a read, and the second operation reads the
+value written by the first operation.
+_From-reads_ is a relation between operations, where the first operation is a
+read, the second operation is a write, and the first operation reads a value
+written earlier than the second operation in the second operation's scoped
+modification order (or the first operation reads from the initial value, and
+the second operation is any write to the same locations).
+
+Then the implementation must: guarantee that no cycles exist in the union of
+the following relations:
+
+  * location-ordered
+  * scoped modification order (over all atomic writes)
+  * reads-from
+  * from-reads
+
+NOTE: This is a "consistency" axiom, which informally guarantees that
+sequences of operations can't violate causality.
+
 [[memory-model-scoped-modification-order-coherence]]
-== Scoped Modification Order Coherence
+=== Scoped Modification Order Coherence
 
-Let A and B be mutually-ordered atomic operations, where A happens-before B,
-and let O be A's scoped modification order.
-Then:
+Let A and B be mutually-ordered atomic operations, where A is
+location-ordered before B.
+Then the following rules are a consequence of acyclicity:
 
-  * If A and B are both writes, then A must: be earlier than B in O
-  * If A and B are both reads, then the write that A takes its value from
-    must: be earlier in O than (or the same as) the write that B takes its
-    value from
+  * If A and B are both reads and A does not read the initial value, then the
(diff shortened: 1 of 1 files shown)


RobinMorisset:
Thank you for this clarification of the spec! It all looks good to me.

Another consequence of the acyclicity axiom is the following: If a is a store and b is an atomic read-modify-write operation, and b reads-from a, and a and b are mutually ordered, then a and b are adjacent in a's scoped modification order.

Proof:
- a and b are mutually ordered, so b is in a's scoped modification order
- If b < a in that order, there would be a loop in scoped modification order + reads-from
- If a < c < b in that order, then we would have b `from-reads` c, giving us another loop
- So a < b with nothing in between.

This property is the equivalent of the "rmw_atomicity" rule in the C++11 model. I don't know if you want it made explicit in your spec, but since you are listing coherency properties that mirror those from C++11, I thought I may mention it.

jeffbolznv:
Thanks, I've added that.

appendices/memorymodel.txt
```diff
@@ -788,21 +788,53 @@ NOTE: It is possible for there to be a write between X and Y that overwrites
...
+  * scoped modification order (over all atomic writes)
+  * reads-from
+  * from-reads
 
-  * If A and B are both writes, then A must: be earlier than B in O
-  * If A and B are both reads, then the write that A takes its value from
-    must: be earlier in O than (or the same as) the write that B takes its
-    value from
+NOTE: This is a "consistency" axiom, which informally guarantees that
+sequences of operations can't violate causality.
```
Tobski: nit: should be a Note block.

appendices/memorymodel.txt
```diff
@@ -788,21 +788,53 @@ NOTE: It is possible for there to be a write between X and Y that overwrites
...
-  * If A is a read and B is a write, then A must: take its value from a
-    write earlier than B in O
+    write later than A in A's scoped modification order (no cycles between
+    location-order, scoped modification order, and from-reads).
+  * If A and B are both writes, then A must: be earlier than B in A's scoped
+    modification order (no cycles between location-order and scoped
+    modification order).
+  * If A is a write and B is a read-modify-write and B reads the value
+    written by A, then B comes immediately after A in A's scoped modification
+    order (no cycles between scoped modification order and from-reads).
```
Tobski: If this is only a *consequence*, and is thus not a normative rule, should we call it out in the spec as such? Other sections are tagged "(Informal)" to denote this.

appendices/memorymodel.txt
```diff
@@ -788,21 +788,53 @@ NOTE: It is possible for there to be a write between X and Y that overwrites
...
 
-  * If A and B are both writes, then A must: be earlier than B in O
-  * If A and B are both reads, then the write that A takes its value from
-    must: be earlier in O than (or the same as) the write that B takes its
-    value from
+NOTE: This is a "consistency" axiom, which informally guarantees that
+sequences of operations can't violate causality.
+
+[[memory-model-scoped-modification-order-coherence]]
+=== Scoped Modification Order Coherence
```
Tobski: nit: I dislike the use of the word "coherence" due to all the reasons we've previously discussed about this term ;). If there's no sane alternative or if this is mirroring some C++ term then it's fine, but I figured it was worth at least mentioning.

appendices/memorymodel.txt
jeffbolznv: I'd like to keep this consistent with the rest of the appendix for now, but we can fix all of them in a later change if you'd like.

appendices/memorymodel.txt
jeffbolznv: This is mirroring C++ terminology, each of these rules has a note like: "[Note: This requirement is known as read-read coherence. —end note]"

appendices/memorymodel.txt
jeffbolznv: It's still normative, it just follows from the previous section.

PR #928, opened 2019-02-21
Fix typo in image markup

krOoze:
removes chunked build warning
makes image into inline svg


oddhack:
Thanks! This was annoying enough in the build process that I just applied it internally, but the PR will be credited in the next change log (probably in about a week).

PR #947, opened 2019-03-26
Generate a quick ref HTML page.

dj2:
This Cl adds a generator to create a quick_ref.html page which links to
the indvidual spec pages. The structure of the generated page is similar
to the vulkan quick reference document.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -208,7 +208,7 @@ chunked: $(HTMLDIR)/vkspec.html $(SPECSRC) $(COMMONDOCS)
 	$(QUIET)$(ROSWELL) $(ROSWELLOPTS) $(CHUNKER) \
 	    $(HTMLDIR)/vkspec.html -o $(HTMLDIR)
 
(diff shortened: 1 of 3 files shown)


dj2:
Does this seem like it would be generally useful? I find it's a lot easier to search in the page of just structs/enums and functions to find what I want and then got to the full page.

oddhack:
Thanks for the suggestion - we can definitely stand to have some navigation. The index in man/apispec.txt is sort of this if it's changed from including the asciidoc refpages to linking to the target page, so I'm tempted to repurpose that (also, it's generated from the same process as the refpages, which guarantees that there won't be any dangling links). We'll take a closer look at this soon.

dj2:
Does it generate an HTML version? I'm mostly looking for something I can use from the browser as I don't use the Vulkan man pages.

dj2:
Going to close this as there doesn't seem to be interest.

PR #948, opened 2019-04-02
Fix the VK_EXT_full_screen_exclusive dependency on VK_KHR_win32_surface

aloucks:
`VK_KHR_win32_surface` is an extension, not a feature


oddhack:
@cubanismo please sign off since it's your extension, but this looks good.

PR #966, opened 2019-05-14
Add VK_INTEL_shader_integer_functions2

ianromanick:
The related SPV extension has already been merged.

@jekstrand


oddhack:
Approved - will go in the next spec update. FYI it's generally a bit easier for process to do new extensions through gitlab, but not a big deal in this case.

PR #995, opened 2019-07-06
Remove redundant render pass VUs

krOoze:
These VUs seem redundant to [VUID-VkRenderPassCreateInfo-pDependencies-00837](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#VUID-VkRenderPassCreateInfo-pDependencies-00837) and its siblings.

There was perhaps some attempt at being more generic. But any pipeline bind point is unlikely to support `VK_PIPELINE_STAGE_HOST_BIT` anyway. And not sure why any future self-dependencies could not theoretically be non-graphic.

### DIFF
--- a/chapters/renderpass.txt
+++ b/chapters/renderpass.txt
@@ -1215,12 +1215,6 @@ layouts as follows:
 
 .Valid Usage
 ****
-  * [[VUID-VkSubpassDependency-srcSubpass-00858]]
-    If pname:srcSubpass is not ename:VK_SUBPASS_EXTERNAL, pname:srcStageMask
-    must: not include ename:VK_PIPELINE_STAGE_HOST_BIT
-  * [[VUID-VkSubpassDependency-dstSubpass-00859]]
-    If pname:dstSubpass is not ename:VK_SUBPASS_EXTERNAL, pname:dstStageMask
-    must: not include ename:VK_PIPELINE_STAGE_HOST_BIT
   * [[VUID-VkSubpassDependency-srcStageMask-00860]]
     If the <<features-geometryShader,geometry shaders>> feature is not
     enabled, pname:srcStageMask must: not contain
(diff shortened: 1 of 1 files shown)


krOoze:
@Tobski Validation is slave to the spec, so that is no surprise.

But they seem fundamentally redundant to me. The VU I am keeping is worded thus (`VkRenderPassCreateInfo`):

> For any element of `pDependencies`, if the `srcSubpass` is not `VK_SUBPASS_EXTERNAL`, all stage flags included in the `srcStageMask` member of that dependency **must** be a pipeline stage supported by the pipeline identified by the `pipelineBindPoint` member of the source subpass

The VU about `VK_PIPELINE_STAGE_HOST_BIT` is clearly redundant to this, as there is no `pipelineBindPoint` that corresponds to it. And it is unlikely there ever will be.

The VU about graphics flags seem pointless speculative promise about the future. It says only graphics subpasses can have self-dependencies.

And considering `pipelineBindPoint` is only ever allowed to be `GRAPHICS`, as it is currently written, it also makes it redundant.

Actually ,if we get back to the layers, you can actually see they assume it is always `GRAPHICS`.

krOoze:
PS: Right. It works in isolation because the layers artificially if-else the case and gives priority to the more specific VUs over the general one.

Actually I even seem to have branch cleaning that up. Probably why I opened this Issue in the first place because tests stopped working for me...

Tobski:
I'm not following - the VUs you've cut and the VUs you're claiming make them redundant are calling out _explicitly_ opposing conditions:

>  * **If srcSubpass is not VK_SUBPASS_EXTERNAL**, srcStageMask
>    must not include VK_PIPELINE_STAGE_HOST_BIT

vs.

> * For any element of pDependencies, **if the srcSubpass is not VK_SUBPASS_EXTERNAL**, all stage flags included in the srcStageMask member of that dependency must be a pipeline stage supported by the pipeline identified by the pipelineBindPoint member of the source subpass.

One covers external dependencies, the other covers non-external dependencies. After looking at it further I don't think these can even be combined - external subpass dependencies can wait on literally anything outside the render pass *except* the host - that's what the first VU is for. Whereas when you wait on another subpass, you're only allowed to wait on pipeline stages which that subpass executes - something that doesn't make sense when your subpass is external.

For the other VU you've cut:

>     If pname:srcSubpass is equal to pname:dstSubpass, pname:srcStageMask and
>    pname:dstStageMask must: not set any bits that are neither
>    ename:VK_PIPELINE_STAGE_ALL_GRAPHICS_BIT, nor one of the
>    <<synchronization-pipeline-stages-types,graphics pipeline stages>>

Yes this is technically redundant. The thought was that if we ever introduced another pipeline bind point, we may not want it to allow subpass self dependencies, and this would cover it. However, right now that's not the case, and we could reintroduce it if we actually added such a thing. I think I'm ok with removing it as long as the validation layers can be coaxed into properly returning an alternative error for the same test that currently exists.

Tobski:
Ugh I'm a derp, I somehow missed the "not" in the first VU. Alright, then I'm happy with this change if you get a patch to the validation layers that makes the tests that currently execute fire an appropriate other VU that exists today.

krOoze:
@Tobski https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1089

PR #1004, opened 2019-07-17
Clarify fptr get tables

### LINKED ISSUE #1003: vkGetInstanceProcAddr behavior missing
I think this is missing the definition for:
- _instance_ + `vkEnumerateInstanceVersion` -> ?
- _instance_ + `vkEnumerateInstanceExtensionProperties` -> ?
- _instance_ + `vkEnumerateInstanceLayerProperties` -> ?
- _instance_ + `vkCreateInstance` -> ?

krOoze:
fixes #1003

### DIFF
--- a/chapters/initialization.txt
+++ b/chapters/initialization.txt
@@ -56,28 +56,31 @@ be cast to the type of the command being queried.
 [width="80%",options="header"]
 |====
 | pname:instance   | pname:pName                                  | return value
-| *                | `NULL`                                       | undefined:
-| invalid instance | *                                            | undefined:
+| *^1^             | `NULL`                                       | undefined:
+| invalid non-`NULL` instance | *^1^                              | undefined:
 ifdef::VK_VERSION_1_1[]
 | `NULL`           | flink:vkEnumerateInstanceVersion             | fp
 endif::VK_VERSION_1_1[]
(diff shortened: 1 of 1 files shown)


chapters/initialization.txt
```diff
@@ -56,28 +56,31 @@ be cast to the type of the command being queried.
...
-| instance         | available device extension^2^ commands for pname:instance | fp^1^
-| instance         | * (any pname:pName not covered above)        | `NULL`
+| instance         | core Vulkan command                          | fp^2^
+| instance         | enabled instance extension commands for pname:instance    | fp^2^
+| instance         | available device extension^3^ commands for pname:instance | fp^2^
+2+|  any other case, not covered above                            | `NULL`
 |====
 
 1::
+    "*" means any conceivable value for the parameter (including valid
```
critsec: nitpick: I'd remove "conceivable", here and in the next table. "Any value" is sufficient, but if we want to reinforce the point we should use "possible" or maybe "representable". The word "conceivable" is more subjective: different people can conceive of different things.

krOoze: Well, there is difference between "conceived" and "conceivable". It does not require real people to actually conceive it.

But far be it for me to stand in the way of wordsmithery. "representable" seems more specific, and fiting this domain.

PR #1006, opened 2019-07-25
Reserve extension 286 for wl_acquire_display

ddevault:
See #1001

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -11223,5 +11223,11 @@ typedef void <name>CAMetalLayer</name>;
                 <enum value="&quot;VK_EXT_extension_285&quot;"              name="VK_EXT_extension_285"/>
             </require>
         </extension>
+        <extension name="VK_EXT_extension_286" number="286" type="instance" author="EXT" contact="Drew DeVault <sir@cmpwn.com>" supported="disabled">
+            <require>
+                <enum value="0"                                             name="VK_EXT_EXTENSION_286_SPEC_VERSION"/>
+                <enum value="&quot;VK_EXT_extension_286&quot;"              name="VK_EXT_extension_286"/>
+            </require>
+        </extension>
     </extensions>
 </registry>


xml/vk.xml
```diff
@@ -11222,6 +11222,12 @@ typedef void <name>CAMetalLayer</name>;
                 <enum value="0"                                             name="VK_EXT_EXTENSION_285_SPEC_VERSION"/>
                 <enum value="&quot;VK_EXT_extension_285&quot;"              name="VK_EXT_extension_285"/>
             </require>
+        </extension>
+		<extension name="VK_EXT_extension_286" number="286" type="instance" author="MESA" contact="Drew DeVault <sir@cmpwn.com>" supported="disabled">
```
jekstrand: I think your text editor had a minor issue here and the indentation got messed up.  We use spaces, not tabs.

ddevault:
Whoops, good catch.

xml/vk.xml
jekstrand: Looks good to me now.

xml/vk.xml
```diff
@@ -11223,5 +11223,11 @@ typedef void <name>CAMetalLayer</name>;
                 <enum value="&quot;VK_EXT_extension_285&quot;"              name="VK_EXT_extension_285"/>
             </require>
         </extension>
+        <extension name="VK_EXT_extension_286" number="286" type="instance" author="MESA" contact="Drew DeVault <sir@cmpwn.com>" supported="disabled">
```
krOoze: Shouldn't it have `MESA` prefix then? `MESA` seem defined in XML as tag.

ddevault: I think EXT is appropriate and MESA is not. I wrote that before I understood the meaning of the field and never went back to change it. Just pushed a fixed version now.

jekstrand: It also really doesn't matter when just reserving an extension number.  We can do "the right thing" when the real version of the extension rolls out.

krOoze: Well yea, though it might be annoying to refactor the whole extension spec with proper tag.

ddevault:
Thanks!

oddhack:
BTW, the angle brackets around the email address are invalid XML syntax - I didn't notice this until I had cherry-picked the commit into our internal gitlab server where we run CI tests. I will be pushing a spec update shortly which just removes the brackets. Also, we prefer to use github IDs here instead of email addresses.

ddevault:
Thanks for catching that! I would prefer to use my email than my GitHub username, though.

PR #1012, opened 2019-07-31
Compatible pathlib for py2 and py3 in script/generator.py

### LINKED ISSUE #1011: Incompatible moudle pathlib for py2 and py3 in script/generator.py
https://github.com/KhronosGroup/Vulkan-Docs/blob/master/scripts/generator.py
> from pathlib import Path

The script throws import error 'No module named pathlib' in environment with python2, becuase pathlib is named pathlib2 in python2.

haoxli:
pathlib name is not compatible in python2 and python3, it's named pathlib2 in python2.

Fixes #1011

### DIFF
--- a/scripts/generator.py
+++ b/scripts/generator.py
@@ -21,7 +21,10 @@ import os
 import re
 import pdb
 import sys
-from pathlib import Path
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
 
 def write( *args, **kwargs ):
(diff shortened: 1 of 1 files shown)


Kangz:
PTAL @oddhack !

oddhack:
We're going to accept this, but be aware that python 2 support is not something we actually care about, and as we understand it, python 2 itself is being deprecated rapidly.

PR #1014, opened 2019-08-05
Clarify the ACCESS_MEMORY bit flags

### LINKED ISSUE #1007: Inconsistent description of VK_ACCESS_MEMORY_*_BIT wrt host
In description of access mask it says:

> `VK_ACCESS_MEMORY_READ_BIT` specifies read access via non-specific entities. These entities include the Vulkan device **and host**, but **may** also include entities external to the Vulkan device or otherwise not part of the core Vulkan pipeline.

But in specific parts describing visibility it says you need `VK_ACCESS_HOST_*_BIT` explicitly. E.g. most importantly:

> If the destination access mask includes `VK_ACCESS_HOST_READ_BIT` or `VK_ACCESS_HOST_WRITE_BIT`, then the dependency includes a memory domain operation from device domain to host domain.

Tobski:
This change makes clear that VK_ACCESS_MEMORY_READ_BIT and VK_ACCESS_MEMORY_WRITE_BIT are meant to be equivalent to setting all applicable READ/WRITE access flags.

Closes #1007

### DIFF
--- a/chapters/synchronization.txt
+++ b/chapters/synchronization.txt
@@ -789,24 +789,12 @@ endif::VK_KHR_depth_stencil_resolve[]
     operation.
     Accesses of this type are not performed through a resource, but directly
     on memory.
-  * ename:VK_ACCESS_MEMORY_READ_BIT specifies read access via non-specific
-    entities.
-    These entities include the Vulkan device and host, but may: also include
-    entities external to the Vulkan device or otherwise not part of the core
-    Vulkan pipeline.
-    When included in a destination access mask, makes all available writes
-    visible to all future read accesses on entities known to the Vulkan
-    device.
-  * ename:VK_ACCESS_MEMORY_WRITE_BIT specifies write access via non-specific
-    entities.
-    These entities include the Vulkan device and host, but may: also include
-    entities external to the Vulkan device or otherwise not part of the core
-    Vulkan pipeline.
-    When included in a source access mask, all writes that are performed by
-    entities known to the Vulkan device are made available.
-    When included in a destination access mask, makes all available writes
-    visible to all future write accesses on entities known to the Vulkan
-    device.
+  * ename:VK_ACCESS_MEMORY_READ_BIT is equivalent to the logical OR of
+    every other etype:READ access flag bit that is supported in the
+    pipeline stage it is used with.
+  * ename:VK_ACCESS_MEMORY_WRITE_BIT is equivalent to the logical OR of
+    every other etype:WRITE access flag bit that is supported in the
+    pipeline stage it is used with.
 ifdef::VK_EXT_conditional_rendering[]
   * ename:VK_ACCESS_CONDITIONAL_RENDERING_READ_BIT_EXT specifies read access
     to a predicate as part of conditional rendering.
--- a/chapters/synchronization.txt
+++ b/chapters/synchronization.txt
@@ -789,12 +789,12 @@ endif::VK_KHR_depth_stencil_resolve[]
     operation.
     Accesses of this type are not performed through a resource, but directly
     on memory.
-  * ename:VK_ACCESS_MEMORY_READ_BIT is equivalent to the logical OR of
-    every other etype:READ access flag bit that is supported in the
-    pipeline stage it is used with.
-  * ename:VK_ACCESS_MEMORY_WRITE_BIT is equivalent to the logical OR of
-    every other etype:WRITE access flag bit that is supported in the
-    pipeline stage it is used with.
+  * ename:VK_ACCESS_MEMORY_READ_BIT specifies all read accesses.
+    It is always valid in any access mask, and is treated as equivalent to setting
+    all etext:READ access flags that are valid where it is used.
+  * ename:VK_ACCESS_MEMORY_WRITE_BIT specifies all write accesses.
+    It is always valid in any access mask, and is treated as equivalent to setting
+    all etext:WRITE access flags that are valid where it is used.
 ifdef::VK_EXT_conditional_rendering[]
   * ename:VK_ACCESS_CONDITIONAL_RENDERING_READ_BIT_EXT specifies read access
     to a predicate as part of conditional rendering.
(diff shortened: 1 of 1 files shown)


Tobski:
@krOoze Can you take a look at this and see if it fixes #1007 sufficiently?

chapters/synchronization.txt
```diff
@@ -789,24 +789,12 @@ endif::VK_KHR_depth_stencil_resolve[]
...
-    These entities include the Vulkan device and host, but may: also include
-    entities external to the Vulkan device or otherwise not part of the core
-    Vulkan pipeline.
-    When included in a source access mask, all writes that are performed by
-    entities known to the Vulkan device are made available.
-    When included in a destination access mask, makes all available writes
-    visible to all future write accesses on entities known to the Vulkan
-    device.
+  * ename:VK_ACCESS_MEMORY_READ_BIT is equivalent to the logical OR of
+    every other etype:READ access flag bit that is supported in the
```
krOoze: What's `etype`? Probably meant to be `etext`.

krOoze:
@Tobski I dislike one thing about this. E.g. the quote in #1007 still only mentions `VK_ACCESS_HOST_*_BIT`. The edits do technically say `VK_ACCESS_MEMORY_*_BIT` is equivalent to OR of everything. But I am not sure this style of writing sufficiently explicit and unambiguous for a specification.

chapters/synchronization.txt
```diff
@@ -789,24 +789,12 @@ endif::VK_KHR_depth_stencil_resolve[]
...
+    pipeline stage it is used with.
+  * ename:VK_ACCESS_MEMORY_WRITE_BIT is equivalent to the logical OR of
+    every other etype:WRITE access flag bit that is supported in the
```
krOoze: Probably doesn't matter, but I think the nomenclature usually used here is "bits being set" rather than "logical OR of values".

Also maybe should qualify "functionally equivalent" as the values are not actually equivalent.

krOoze:
In "Table 4. Supported access types" there is "N\A" for `VK_ACCESS_MEMORY_*_BIT`. Maybe it should say "any" or something like that.

chapters/synchronization.txt
Tobski: Whoops, yep - will fix.

Tobski:
@krOoze I've updated the text to address your feedback - hopefully this makes sense!

Tobski:
(Also sorry it's taken so long - been a busy month...)

chapters/synchronization.txt
```diff
@@ -789,12 +789,12 @@ endif::VK_KHR_depth_stencil_resolve[]
...
     Accesses of this type are not performed through a resource, but directly
     on memory.
-  * ename:VK_ACCESS_MEMORY_READ_BIT is equivalent to the logical OR of
-    every other etype:READ access flag bit that is supported in the
-    pipeline stage it is used with.
-  * ename:VK_ACCESS_MEMORY_WRITE_BIT is equivalent to the logical OR of
-    every other etype:WRITE access flag bit that is supported in the
-    pipeline stage it is used with.
+  * ename:VK_ACCESS_MEMORY_READ_BIT specifies all read accesses.
+    It is always valid in any access mask, and is treated as equivalent to setting
```
krOoze: On the other hand this may be too brittle.
If some VU said the access mask must be `0`, then it would not be "always" valid.

chapters/synchronization.txt
```diff
@@ -894,8 +894,8 @@ endif::VK_NV_ray_tracing[]
 |ename:VK_ACCESS_TRANSFER_WRITE_BIT                           | ename:VK_PIPELINE_STAGE_TRANSFER_BIT
 |ename:VK_ACCESS_HOST_READ_BIT                                | ename:VK_PIPELINE_STAGE_HOST_BIT
 |ename:VK_ACCESS_HOST_WRITE_BIT                               | ename:VK_PIPELINE_STAGE_HOST_BIT
-|ename:VK_ACCESS_MEMORY_READ_BIT                              | N/A
-|ename:VK_ACCESS_MEMORY_WRITE_BIT                             | N/A
+|ename:VK_ACCESS_MEMORY_READ_BIT                              | Any
+|ename:VK_ACCESS_MEMORY_WRITE_BIT                             | Any
```
krOoze: Even `TOP_OF_PIPE` and `BOTTOM_OF_PIPE` ?

chapters/synchronization.txt
Tobski: > On the other hand this may be too brittle.
> If some VU said the access mask must be 0, then it would not be "always" valid.

Yea I had this thought too, to the point that I figured out the one time this happens in the spec currently - if for some reason you were to submit a command buffer to a sparse-binding-only queue. It being valid remains in-line with the existing description of it, so it's fine like this IMO.

> PS: should be "any access mask" -> "any access mask parameter"?

We refer to "access mask" as a thing many other places in this chapter, so I think this is also fine.

chapters/synchronization.txt
Tobski: Ugh my two least favourite pipeline stages... that's a good question. I'm going to say yes for now, but this probably needs affirmation by the WG.

krOoze: Oh and can `VK_ACCESS_MEMORY_WRITE_BIT` be on any stage that is read only, and vice versa?

Tobski: That one I don't think is contentious - that's already true from the current language IMO

Tobski:
@oddhack This one needs a bit of review from the WG, can you put it on next week's agenda?

chapters/synchronization.txt
krOoze: Just asking. It makes it different from the other access masks. So making sure that is intentional.
E.g. `VK_PIPELINE_STAGE_VERTEX_INPUT_BIT` otherwisely only accepts `VK_ACCESS_*_READ_BIT`. So `VK_ACCESS_MEMORY_WRITE_BIT` is valid there anyway?

Tobski: Vulkan WG agreed with my assertions - these are wildcards, valid anywhere you can stick an access flag.

chapters/synchronization.txt
Tobski: As per comment below, WG agreed with my assertions.

PR #1023, opened 2019-08-13
Fix pNext features not accessible via VkPhysicalDeviceFeatures2

krOoze:
Require all pNext features to be accesible via VkPhysicalDeviceFeatures2::pNext.

Would it be a reasonable restriction?

BTW: For some reason `VkPhysicalDeviceFeatures2` struct does not have generated `pNext` VU...
probably because of its dual use in `vkGet*Features` and `vkCreateDevice`...

### DIFF
--- a/chapters/devsandqueues.txt
+++ b/chapters/devsandqueues.txt
@@ -966,6 +966,10 @@ ifdef::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
   * [[VUID-VkDeviceCreateInfo-pNext-00373]]
     If the pname:pNext chain includes a slink:VkPhysicalDeviceFeatures2
     structure, then pname:pEnabledFeatures must: be `NULL`
+  * If the pname:pNext chain includes a slink:VkPhysicalDeviceFeatures2
+    structure, then pname:pNext chain must: not include any
+    stext:Vk*Features* instances earlier in chain before the
+    slink:VkPhysicalDeviceFeatures2
 endif::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
 ifdef::VK_AMD_negative_viewport_height[]
 ifdef::VK_VERSION_1_1[]


chapters/devsandqueues.txt
```diff
@@ -966,6 +966,10 @@ ifdef::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
   * [[VUID-VkDeviceCreateInfo-pNext-00373]]
     If the pname:pNext chain includes a slink:VkPhysicalDeviceFeatures2
     structure, then pname:pEnabledFeatures must: be `NULL`
+  * If the pname:pNext chain includes a slink:VkPhysicalDeviceFeatures2
+    structure, then pname:pNext chain must: not include any
+    stext:Vk*Features* instances earlier in chain before the
+    slink:VkPhysicalDeviceFeatures2
```
jeffbolznv: We discussed this in the working group today and the group generally doesn't like straying from the more general rule that the order of the pNext chain doesn't matter. Is this related to https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/1156? If so, how bad is it to just handle the case where they're out of order?

krOoze: Well as you can see, I (hopefully) happen to get that right in the layers (otherwise I would not know I am doing it wrong and wouldn't make this Issue). But it served as an inspiration.

I mean, it feels like a gotcha.
If the `VkPhysicalDeviceFeatures2KHR` will not house all the features, then what's the point of having a `pNext` variant of `VkPhysicalDeviceFeatures` in the first place? The design somewhat begs that all the Features should be in `VkPhysicalDeviceFeatures2KHR` chain if the user uses that one instead of the Vulkan 1.0 version.

I guess it could be "solved" by a CTS check instead.

Although it generally feels weird\hacky that you have to somewhat get the original `pNext` chain root, and the extension struct is thus not "self-sufficient".
Though I would think most apps supply it in the "natural" order anyway. It is just that it is a posibility and I had to account for it in the layers just-in-case.

jeffbolznv: The working group discussed this again today and wants to keep the rule that there's no assumed ordering of the pNext chain. So shall we close this PR?

krOoze: Fine with me

Issue #1009, opened 2019-07-29
VK_EXT_buffer_device_address revision history not updated


WoodRain:
Hi @jeffbolznv From [git blame info of VK_EXT_buffer_device_address](https://github.com/KhronosGroup/Vulkan-Docs/blame/master/appendices/VK_EXT_buffer_device_address.txt), the struct name is changed, but the revision number and revision history is not changed accordingly.
By the way, why not changing VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_BUFFER_ADDRESS_FEATURES_EXT to VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_BUFFER_DEVICE_ADDRESS_FEATURES_EXT in the initial version. It seems more suitable together with VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO_EXT.

jeffbolznv:
We've done several of these structure renames recently for consistency, but they're backward compatible and we haven't changed the version numbers for any of them.

> By the way, why not changing VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_BUFFER_ADDRESS_FEATURES_EXT to VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_BUFFER_DEVICE_ADDRESS_FEATURES_EXT in the initial version.

I don't understand the question. We did this rename, what's missing?

oddhack:
If I'm following the question, the answer is that we just didn't catch the inconsistency before the extension shipped.

WoodRain:
Well, I mean the naming changes may be recorded in the spec to avoid some sort of confusion. And people had a look at the earlier editions of those names can still find those in the new spec :)

oddhack:
@WoodRain want to propose a PR suggesting what you would like to see here? Probably something in the Issues section of the appendix would be most appropriate.

WoodRain:
Got it. A PR is issued as #1024 .

PR #1024, opened 2019-08-18
Add issue for name changes in buffer_device_address

WoodRain:
Some structure names are changed in VK_EXT_buffer_device_address for consistency and this PR adds an issue to log it.

### DIFF
--- a/appendices/VK_EXT_buffer_device_address.txt
+++ b/appendices/VK_EXT_buffer_device_address.txt
@@ -66,7 +66,9 @@ None
 
 === Issues
 
-None
+1) Where is VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_BUFFER_ADDRESS_FEATURES_EXT and VkPhysicalDeviceBufferAddressFeaturesEXT?
+
+*RESOLVED*: They were renamed as VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_BUFFER_DEVICE_ADDRESS_FEATURES_EXT and VkPhysicalDeviceBufferDeviceAddressFeaturesEXT accordingly for consistency.
 
 === Version History
 


appendices/VK_EXT_buffer_device_address.txt
```diff
@@ -66,7 +66,9 @@ None
 
 === Issues
 
-None
+1) Where is VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_BUFFER_ADDRESS_FEATURES_EXT and VkPhysicalDeviceBufferAddressFeaturesEXT?
+
+*RESOLVED*: They were renamed as VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_BUFFER_DEVICE_ADDRESS_FEATURES_EXT and VkPhysicalDeviceBufferDeviceAddressFeaturesEXT accordingly for consistency.
```
oddhack: Add something like "the old names are still available, for compatibility".

jeffbolznv: These should use elink/slink macros.

oddhack:
N.b. closes #1009 once accepted.

appendices/VK_EXT_buffer_device_address.txt
krOoze: That somewhat makes the question invalid, if the old names are still available...

krOoze:
I think "revision 2" should be added to the "Version History" to properly address the Issue.

appendices/VK_EXT_buffer_device_address.txt
WoodRain: Add some words for compatibility in the new commit.

oddhack:
> I think "revision 2" should be added to the "Version History" section to properly address the Issue. In XML it is `2`, but in spec it is `1`.

There are some pending internal edits to fix all the appendix / XML version inconsistencies, so I'd leave this alone for now. Otherwise the PR looks good, so merging it.

PR #1030, opened 2019-08-25
In same parent VU do not check validity of ignored values

krOoze:
Ignored value could be accidentally "valid". So when making same-parent VU qualify instead that the handles must be previously required to be valid (, not simply valid).


krOoze:
For example `VkDescriptorUpdateTemplateCreateInfo`.
Diff:
diff-parent

krOoze:
Technically, now reading it, the new wording is not accurate neither (wrt "null or valid handle" VU). Maybe I am overthinking it, since there is only low chance someone will misinterpret the VU anyway...

krOoze:
Attempt nr. deux. Diff above updated. Please re-review.

PR #1031, opened 2019-08-25
Make template descriptorSetLayout noauto

krOoze:
to be consistent with the `pipelineLayout` member in the same struct.


krOoze:
Diff:
diff-noauto

krOoze:
@Tobski Yes, `A` is original, `B` is new.

PR #1046, opened 2019-10-09
Queue transfer ops now refer to access masks

Tobski:
Previously suggested stage masks were ignored, which is incorrect - only the access masks are ignored.


chapters/synchronization.txt
```diff
@@ -4855,7 +4855,7 @@ family.
 The pname:srcQueueFamilyIndex parameter of the barrier must: be set to the
 source queue family index, and the pname:dstQueueFamilyIndex parameter to
 the destination queue family index.
-pname:dstStageMask is ignored for such a barrier, such that no visibility
+pname:dstAccessMask is ignored for such a barrier, such that no visibility
```
krOoze: I think it makes it unclear what is ignored. Is the mask still considered wrt to VUs?

Elswhere it says:
> [...] and the first access scope includes no access, as if `srcAccessMask` was `0`.

krOoze: At some point should be documented in the VUs...

Tobski: Oof, yes. It looks like right now it's implicitly always got to be a valid set of VUs... I can attempt to fix this too whilst I'm here.

Tobski: Done.

Tobski:
@krOoze Please take another look :)

chapters/synchronization.txt
```diff
@@ -4773,6 +4785,18 @@ ifdef::VK_NV_shading_rate_image[]
     have been created with ename:VK_IMAGE_USAGE_SHADING_RATE_IMAGE_BIT_NV
     set
 endif::VK_NV_shading_rate_image[]
+  * If pname:image was created with a sharing mode of
+    ename:VK_SHARING_MODE_EXCLUSIVE, and pname:srcQueueFamilyIndex and
+    pname:dstQueueFamilyIndex don't describe a
+    <<synchronization-queue-transfers-acquire, queue family acquire
```
krOoze: Per styleguide, should not use contractions. "don't"

The link actually does not specify how to recognize an acquire barrier, so might be useless here. Might be better if the VU is more prescriptive. E.g.: 

> - If `image` was not created with a sharing mode of `VK_SHARING_MODE_EXCLUSIVE` then `srcAccessMask` **must** be a valid combination of `VkAccessFlagBits` values
> -  If `image` was created with a sharing mode of `VK_SHARING_MODE_EXCLUSIVE` and `dstQueueFamilyIndex` is the same as the family of the queue that will execute this barrier, then `srcAccessMask` **must** be a valid combination of `VkAccessFlagBits` values

Which technically is a problem, because we are invoking a state that cannot be validated in this spot.

The ignored parameter also impacts this `vkCmdPipelineBarrier` VU, which has not been touched:

> Each element of pMemoryBarriers, pBufferMemoryBarriers and pImageMemoryBarriers must not have any access flag included in its srcAccessMask member if that bit is not supported by any of the pipeline stages in srcStageMask, as specified in the table of supported access types.

Tobski: > Which technically is a problem, because we are invoking a state that cannot be validated in this spot.

Huh that's true, should've written it out as it would've been more obvious, I guess. Not a huge deal though, I can fix that.

Tobski: > The link actually does not specify how to recognize an acquire barrier, so might be useless here.

I've just re-read that link - why don't you think it specifies how to recognise an acquirebarrier? It says quite clearly:

> An acquire operation is defined by executing a buffer memory barrier (for a buffer range) or an image memory barrier (for an image subresource range), on a queue from the destination queue family. The buffer range or image subresource range specified in an acquire operation must match exactly that of a previous release operation. The srcQueueFamilyIndex parameter of the barrier must be set to the source queue family index, and the dstQueueFamilyIndex parameter to the destination queue family index.

What is missing from that description?

Tobski: And done. Other than the fact that I'm still referencing the link that you seemed to disagree was sufficient. Also note that vkCmdWaitEvents validation doesn't consider queue ownership transfers, which is deliberate, as (spoiler alert!) this is not actually intended to be supported and is about to be clarified in one of the next spec updates.

krOoze: > What is missing from that description?

The fact that description is not a description. It is a requirement. It says what acquire barrier is required to have as arguments. It does not say which arguments say it **is** an acquire barrier

Actual description\identification is specified elsewhere (below `VkImageMemoryBarrier`):

> If `dstQueueFamilyIndex` is not equal to `srcQueueFamilyIndex`, and `dstQueueFamilyIndex` is equal to the current queue family, then the memory barrier defines a _queue family acquire operation_ for the specified image subresource range, [...]

chapters/synchronization.txt
```diff
@@ -4236,6 +4252,52 @@ ifdef::VK_NV_mesh_shader[]
     pname:dstStageMask must: not contain
     ename:VK_PIPELINE_STAGE_TASK_SHADER_BIT_NV
 endif::VK_NV_mesh_shader[]
+  * Each element of pname:pMemoryBarriers must: not have any access flag
+    included in its pname:srcAccessMask member that is not supported by any
+    of the pipeline stages in pname:srcStageMask, as specified in the
+    <<synchronization-access-types-supported, table of supported access
+    types>>
```
krOoze: You forgot do delete the original VUs.

chapters/synchronization.txt
```diff
@@ -4236,6 +4252,52 @@ ifdef::VK_NV_mesh_shader[]
...
+    with a sharing mode of ename:VK_SHARING_MODE_EXCLUSIVE, then its
+    pname:dstAccessMask member must: only contain access flags that are
+    supported by any of the pipeline stages in pname:dstStageMask, as
+    specified in the <<synchronization-access-types-supported, table of
+    supported access types>>
+  * For each element of pname:pBufferMemoryBarriers, if its
+    pname:srcQueueFamilyIndex and pname:dstQueueFamilyIndex
+    members describe a <<synchronization-queue-transfers-acquire, queue
+    family acquire operation>>, and its pname:buffer member was created
+    with a sharing mode of ename:VK_SHARING_MODE_EXCLUSIVE, then its
```
krOoze: This might be technically incorrect, as one of the extensions allow Acquire Barrier even for `VK_SHARING_MODE_CONCURRENT` (foreign queue).

If you intent do go with the link, I suggest you delete unnecessary information as mentioning the `*srcQueueFamilyIndex`. Otherwise it might as well be clearer to explicitly prescribe here what is Acquire Barrier, as I suggested previously.

chapters/synchronization.txt
```diff
@@ -4236,6 +4252,52 @@ ifdef::VK_NV_mesh_shader[]
...
+    members describe a <<synchronization-queue-transfers-release, queue
+    family release operation>>, and its pname:buffer member was created
+    with a sharing mode of ename:VK_SHARING_MODE_EXCLUSIVE, then its
+    pname:dstAccessMask member must: only contain access flags that are
+    supported by any of the pipeline stages in pname:dstStageMask, as
+    specified in the <<synchronization-access-types-supported, table of
+    supported access types>>
+  * For each element of pname:pBufferMemoryBarriers, if its
+    pname:srcQueueFamilyIndex and pname:dstQueueFamilyIndex
+    members describe a <<synchronization-queue-transfers-acquire, queue
```
krOoze: I belive it should be opposite. "If it does **not** describe a queue family acquire operation".

chapters/synchronization.txt
```diff
@@ -3999,6 +3985,36 @@ ifdef::VK_NV_mesh_shader[]
     pname:dstStageMask must: not contain
     ename:VK_PIPELINE_STAGE_TASK_SHADER_BIT_NV
 endif::VK_NV_mesh_shader[]
+  * Each element of pname:pMemoryBarriers must: not have any access flag
+    included in its pname:srcAccessMask member that is not supported by any
+    of the pipeline stages in pname:srcStageMask, as specified in the
```
krOoze: Is the double negative necessary?

"must include only flags supported by pipeline stages"?

chapters/synchronization.txt
```diff
@@ -4236,6 +4252,52 @@ ifdef::VK_NV_mesh_shader[]
...
+    pname:dstAccessMask member must: only contain access flags that are
+    supported by any of the pipeline stages in pname:dstStageMask, as
+    specified in the <<synchronization-access-types-supported, table of
+    supported access types>>
+  * For each element of pname:pBufferMemoryBarriers, if its
+    pname:srcQueueFamilyIndex and pname:dstQueueFamilyIndex
+    members describe a <<synchronization-queue-transfers-acquire, queue
+    family acquire operation>>, and its pname:buffer member was created
+    with a sharing mode of ename:VK_SHARING_MODE_EXCLUSIVE, then its
+    pname:srcAccessMask member must: only contain access flags that are
```
krOoze: This might be insufficient to replace the below removed VUs.
It should  say "valid access flags" otherwise non-enabled extension access masks would still be valid here per the table.

chapters/synchronization.txt
Tobski: Eh, I can just remove the sharing mode exclusive part, as the VU for describing acquire/release barriers already bans cases where you'd specify an qfot when it doesn't make sense. Shall do that.

chapters/synchronization.txt
Tobski: True, was just mirroring existing language, but that would be easier.

chapters/synchronization.txt
Tobski: Sticking the word "valid" on it doesn't really change anything here, this suffers from the existing "don't validate extensions explicitly spec" issue. The correct fix would be to add an explicit VU banning each relevant access type explicitly when its extension/feature is not enabled. And fixing that is definitely beyond the scope of this PR...

chapters/synchronization.txt
Tobski: Whoops

chapters/synchronization.txt
Tobski: Ok I see the complaint now, for some reason I thought EXTERNAL and FOREIGN had additional complications but... looks like they're basically fine 🤷‍♂ 

I was mostly just wary of duplicating info, but a simple "src != dst and {src|dst} == this queue" is fine.

Tobski:
Feedback addressed... @krOoze does it look right to you now?

chapters/synchronization.txt
krOoze: Hm, I somehow assumed that's what "valid flag" means. I guess not.

chapters/synchronization.txt
```diff
@@ -3999,6 +3985,36 @@ ifdef::VK_NV_mesh_shader[]
     pname:dstStageMask must: not contain
     ename:VK_PIPELINE_STAGE_TASK_SHADER_BIT_NV
 endif::VK_NV_mesh_shader[]
+  * Each element of pname:pMemoryBarriers must: only include access
+    flags in its pname:srcStageMask member that are supported by any of
+    the pipeline stages in pname:dstStageMask, as specified in the
```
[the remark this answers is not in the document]
Tobski: 🤦‍♂

Tobski:
Many silly fixes later...

chapters/synchronization.txt
```diff
@@ -3999,6 +3985,36 @@ ifdef::VK_NV_mesh_shader[]
     pname:dstStageMask must: not contain
     ename:VK_PIPELINE_STAGE_TASK_SHADER_BIT_NV
 endif::VK_NV_mesh_shader[]
+  * The pname:srcAccessMask member of any element of
```
krOoze: any -> each, if I am reading the styleguide right. (applies uniformly).

"any" only in those last four VUs, where we pick only elements with a certain quality.

Tobski: Done!

Tobski:
@oddhack this is ready for WG sign off.

chapters/synchronization.txt
```diff
@@ -3999,6 +3985,36 @@ ifdef::VK_NV_mesh_shader[]
     pname:dstStageMask must: not contain
     ename:VK_PIPELINE_STAGE_TASK_SHADER_BIT_NV
 endif::VK_NV_mesh_shader[]
+  * The pname:srcAccessMask member of each element of
+    pname:pMemoryBarriers must: only include access flags that are
+    supported by any of the pipeline stages in pname:srcStageMask, as
```
oddhack: I'd prefer "one or more of" vs. "any of" here and below. The latter just reads weirdly to my brain.

krOoze: Above I was suggesting "at least one pipeline stage in"

Tobski: I went with "one or more" because it was shorter. Fixed now.

krOoze: ??
"one or more of the pipeline stages in"
"at least one pipeline stage in" <==>

chapters/synchronization.txt
```diff
@@ -3966,6 +3966,10 @@ semaphore).
     pname:commandBuffer was allocated from, as specified in the
     <<synchronization-pipeline-stages-supported, table of supported pipeline
     stages>>.
+  * [[VUID--srcQueueFamilyIndex-02803]]
```
oddhack: Where'd the VUID come from?

krOoze: This needs to be rebased. It was fixed in my PR.

Tobski: Just took care of this as I resolved the conflict.

oddhack:
@tobski sorry, overlooked including this for this week and it's too much of a pain to re-roll the entire release. I'll get it next time.

PR #1048, opened 2019-10-09
Removed misleading note as per 812

### LINKED ISSUE #812: Request clarification for interaction between semaphore wait and barrier
I'm trying to understand exactly how the initialLayout transition in a renderpass is sychronised with the WSI semaphore. I encountered the _Swapchain Image Acquire and Present_ section in https://github.com/KhronosGroup/Vulkan-Docs/wiki/Synchronization-Examples which seems to be based on a note in the spec (extended KHR version) section 6.4.2, about using semaphore waits to guarantee that the WSI semaphore wait operation happens-before the layout transition. I am trying to understand how it sets up a dependency such that the transition happens-after the semaphore wait operation, as it isn't immediately obvious to me.

The note says to set both the srcStageMask and dstStageMask to VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT, which it says establishes an execution dependency between the wait and the barrier:

> the semaphore signals after the presentation operation completes, the semaphore wait stalls the VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT stage, and there is a dependency from that same stage to itself with the layout transition performed in between.

The "dependency from that same stage to itself" is the main source of confusion for me, as the rest of the spec talks in terms of synchronization scopes, and it's not immediately clear from the note what exactly is in the first sync scope of the barrier such that it actually happens-after the wait. The strong implication is that the wait operation itself is in the first sync scope, which would imply that the wait operation is defined to execute in the stage specified in the pWaitDstStageMask parameter of vkSubmitInfo, and therefore gets included in the first sync scope of the barrier when that stage is specified as the srcStageMask, but elsewhere in the spec it says it doesn't execute in any defined pipeline stage.

I am struggling to find the language in the spec that concretely clears this up for me, maybe someone can point me in the right direction?

Thanks in advance

Tobski:
Removed misleading note about multiple execution dependencies.

Fixes #812

### DIFF
--- a/chapters/synchronization.txt
+++ b/chapters/synchronization.txt
@@ -97,23 +97,6 @@ execution dependencies:
   * Let *B~S~* be the second synchronization scope of the last command in
     *S*.
 
-[NOTE]
-.Note
-====
-An execution dependency is inherently also multiple execution dependencies -
-a dependency exists between each subset of *A'* and each subset of *B'*, and
-the same is true for execution dependency chains.
-For example, a synchronization command with multiple
-<<synchronization-pipeline-stages,pipeline stages>> in its stage masks
-effectively generates one dependency between each source stage and each
-destination stage.
-This can be useful to think about when considering how execution chains are
-formed if they do not involve all parts of a synchronization command's
-dependency.
-Similarly, any set of adjacent dependencies in an execution dependency chain
-can: be considered an execution dependency chain in its own right.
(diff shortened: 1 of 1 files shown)


Tobski:
@oddhack assigning to you to merge/put on the agenda for next week's meeting.

PR #1049, opened 2019-10-10
Clarify the memory mapping note about invalidation

### LINKED ISSUE #889: Does non-coherent mapped new memory need to be invalidated?
There's this note:

> Note
>
> Mapping non-coherent memory does not implicitly invalidate the mapped memory, and device writes that have not been invalidated **must** be made visible before the host reads or overwrites them.

It is ambiguous if newly allocated memory need to be invalidated too. Technically no device write happened yet, but still the memory may contain undefined values. And the note does force us to invalidate even before host writes.

Secondly, the quote should not be a note, as it gives new information containing a "must" sentence. Or the "must" was meant as "should"?

Tobski:
Made clear that the note is not normative and only points out an instance of a data race that is not satisfied implicitly as some may expect.

Closes #889

### DIFF
--- a/chapters/memory.txt
+++ b/chapters/memory.txt
@@ -2772,9 +2772,10 @@ the last time it was invalidated.
 [NOTE]
 .Note
 ====
-Mapping non-coherent memory does not implicitly invalidate the mapped
-memory, and device writes that have not been invalidated must: be made
-visible before the host reads or overwrites them.
+Mapping non-coherent memory does not implicitly invalidate that memory.
+Device writes to non-coherent memory need to be made visible to the host
+before any further host access to avoid a data race, irrespective of when
+that memory was mapped to the host.
 ====
 
 include::{generated}/validity/protos/vkInvalidateMappedMemoryRanges.txt[]
--- a/chapters/memory.txt
+++ b/chapters/memory.txt
@@ -2773,9 +2773,6 @@ the last time it was invalidated.
 .Note
 ====
 Mapping non-coherent memory does not implicitly invalidate that memory.
-Device writes to non-coherent memory need to be made visible to the host
-before any further host access to avoid a data race, irrespective of when
-that memory was mapped to the host.
 ====
 
 include::{generated}/validity/protos/vkInvalidateMappedMemoryRanges.txt[]


chapters/memory.txt
```diff
@@ -2772,9 +2772,10 @@ the last time it was invalidated.
 [NOTE]
 .Note
 ====
-Mapping non-coherent memory does not implicitly invalidate the mapped
-memory, and device writes that have not been invalidated must: be made
-visible before the host reads or overwrites them.
+Mapping non-coherent memory does not implicitly invalidate that memory.
+Device writes to non-coherent memory need to be made visible to the host
+before any further host access to avoid a data race, irrespective of when
+that memory was mapped to the host.
```
krOoze: - further -> subsequent?
- To avoid a data race, device writes ... ?
- need to be **explicitly** made visible  \ (or: "the App needs to make visible")
- **a** non-coherent memory (I think it is countable noun, as it refers to the one specific object or range, i.e. the parameters of `vkMapMemory`)

- , irrespective of when that memory was mapped to the host
->
always need to be made visible ...
(it is irrelevant whether the memory was mapped after the device write).

Tobski: On second thoughts, I'm just going to suggest we just reduce the note to the first sentence. The rest is just niggle-fuel :P

krOoze: Fine with me

Tobski: Yes, this feels better.

Tobski: > a non-coherent memory (I think it is countable noun, as it refers to the one specific object or range, i.e. the parameters of vkMapMemory)

Oh on this, since I didn't technically resolve it - the usage is correct as-is, it's used a mass noun - it's not a single VkMemory object that gets mapped, rather some range of a larger whole.

Tobski:
Assuming there's no other feedback from @krOoze ... @oddhack I think this is ready to go?

krOoze:
Sorry if it is extending the scope of the PR, but for consistency, shouldn't there be a similar note for `vkUnmapMemory`?

Tobski:
I think the reason it's on MapMemory is because other map commands that exist elsewhere (GL?) do invalidate. And there's something about unmap not flushing as well. We could just state all these points for both commands, although tbh I'm not sure it's worth it. This one is here specifically to point out a historical issue, so I'm not inclined to put it elsewhere.

krOoze:
> And there's something about unmap not flushing as well

That is what I meant.

Ah nevermind, I found it. It is actually not a Note.

PR #1066, opened 2019-10-18
Simplify links to external extensions

krOoze:
plus some collateral damage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,9 +139,6 @@ SPECDATE     = $(shell echo `date -u "+%Y-%m-%d %TZ"`)
 SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD 2> /dev/null || echo Git branch not available`) \
 	     commit: $(shell echo `git log -1 --format="%H" 2> /dev/null || echo Git commit not available`)
 
-# Base path to SPIR-V extensions on the web.
-SPIRVPATH = https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions
-
 # Some of the attributes used in building all spec documents:
 #   chapters - absolute path to chapter sources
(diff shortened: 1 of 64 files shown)


krOoze:
@oddhack Hm, why complicate things though... are you planning to preprocess the links or something? YAGNI

`link:` is the built-in, so probably taken. Would have to mess around in Ruby to figure out how not to break normal links. I am thinking maybe `registry:`?

oddhack:
@krOoze I didn't mean to say the attribute names should be asciidoctor macros, just that they shouldn't claim the base extension name in this namespace. E.g. {link-GL_EXT_blah} would be OK. With that change we could accept this.

krOoze:
@oddhack Hm, why though? The extensions are already namespaced by its prefix, and shouldn't otherwisely occur in the Vulkan spec.

The mangled name is awfully close to a macro. Do you have any esthetical preference before I start? `{link-GL_EXT_blah}` or `reglink:GL_EXT_blah` (or any other macro name that is not already taken).

BTW, if it was somewhat made consistent at what registry given GLSL extension is and if it has `GLSL` or `GL` prefix, then macro would be better and it would not require authors to manually add the links.

oddhack:
I'd prefer link-extensionname. Changing the locations of files in the existing registries would be painful and require setting up rewrite directives in the webserver, which is not work I'm eager to do just to benefit this PR.

krOoze:
That is not a prerequisite for the PR. It just means I would in the meantime still require the authors to manually add the GL links. On the other hand SPV (and others) does not seem to have such problem, so I could lift the requirement there.

Anyway I meant from purely esthetical standpoint. `{link-GL_EXT_blah}` makes it look macroish already, so maybe I could just as well just make it a real macro.

krOoze:
Updated with the suggestion.

PR #1067, opened 2019-10-18
Fix links inside links

krOoze:


oddhack:
Turns out that this fails the 'allchecks' target, which is arguably a bug in the checker rather than the markup, but replacing it with backquote-delimited markup for now.

PR #1071, opened 2019-10-24
Process SVGs for the web

krOoze:
First commit; minor image editing:
- remove `flowRoot` element (resolves warning during a save about Flow Text feature browser compatibility)
- resave images from old Inkscape (resolves warning during a save about changed default DPI)
- normalize arrows in pipeline diagrams (z-order and misplaced arrow in mesh pipeline modification)
- fix z order in texture diagrams

Second commit; produce optimized and valid SVG for HTML inclusion
- optimization which reduces SVG size across all images by 1 MB
- makes a valid inline SVG in HTML (i.e. passing w3c validator as much as possible)
- accompanying styleguide for future images (requires saving copy of the image using special Inkscape settings)


oddhack:
What is the benefit of this MR?

Tobski:
Hey @krOoze, can you write up a description for this including what you've done and why you're doing it? AFAICT it looks like this is doing a minify pass on the images, but it's unclear why you're doing it (html size? load time?), why you're doing it the way you're proposing, and what exactly the way you're doing it is. Any detail here would be helpful in order to get us to evaluate this properly. If you've done any measurement of the impact of this change that would be good too.

krOoze:
The commits explain what they do in their message details.

The first commit fixes some random stuff in the images. `flowRoot` relates to text flow, and Inkscape was complaining when saving that browsers are likely to display it incorrectly. Also it seemed it wasn't used, except that the Root was defined in the SVG tree.

The second commit performs SVG optimization of the images via a feature directly in Inkscape, and the process is described in the accompanying styleguide changes. The size is reduced by more than 1 MB. Also it is possible to configure it in a way that produces valid HTML.
Minification is only a small part of this; I didn't quite care but decided to switch it on. They are from Inkscape, not meant to be human readable anyway.

Why I am doing this is that I like single-page. The version I self-host feels mostly on the border of being usable speedwise for me; so every bit helps. Also I would like to have a HTML valid, before I continue experimenting with "fixing" the single-page spec (now I am annoyed the TOC or something jumps around, probably causing the page to be styled twice). Though none of that relates to you at the moment; still I offer the PR anyway as it is probably the correct™️ way to author SVGs for the web.

krOoze:
PS, settings rationale:
- The page resize changes size to match the image contents+margins. Already mandated by styleguide before.
- the Clean document should clean unused stuff from SVG, Not sure exactly if it does anything, but it does not seem to hurt
- the settings in the ID tab make sure the tag `id`s are unique across the whole HTML
- The XML/HTML attributes failed validation, so I preferred the CSS setting
- XML Declarations setting removes XML declaration and declarations of XML namespaces incompatible with HTML, i,e, making HTML valid
- Disabling viewboxing causes the SVG to have a size, i.e. what we already expect
- rest should reduce the size

oddhack:
Thanks, that's helpful rationale. We will discuss more.

As a general strategy when you're contemplating one of these gigantic global changes, we encourage you to discuss it with us first via an issue rather than dropping dozens or hundreds of changed files and new processes out of nowhere. Almost all of these types of PRs need to have further internal work done for corresponding files on our unreleased feature development branches, so it's very helpful, when possible, to accomplish the changes via scripting tools or other automated ways of making the edits, if we're going to do them at all.

krOoze:
Well, "gigantic"...; it is about half an hour to do it for all the images, when I knew what should be done.  Maybe easier to apologize, than ask for permission, hehe :p

It is no catastrophy if someone releases a SVG that does not follow the process yet. It will simply work as before. And when I notice it in commit diff, I can fix that in 15 s. Automated script is probably the kind of bikesheding I do not need in my life.

oddhack:
That's up to you. But I'm telling you, speaking as the spec editor, that we don't look with enthusiasm on these gigantic PRs because they're a pain to assess and integrate with our internal workflows. Unless there is a clearly elucidated benefit beyond you liking them for your own purposes, they are going to tend to sit there and not get much attention from the WG.

Tobski:
@krOoze thanks for the explanation - the commit history isn't immediately obvious on the PR page so it would be useful in future if you could copy that into the PR description? In general I think we'd like to see/hear the arguments for big changes like this in advance - whether that's in an issue or in the PR description is largely at your discretion.

With regard to this specific PR, now that I understand what it's for I think I'm personally quite happy that this is actually a net benefit and of minimal invasiveness in the long term. Just looks messy without that rationale! Thanks!

PR #1084, opened 2019-11-11
Rename pixel shaders to fragment shaders

### LINKED ISSUE #1082: Pixel shaders should be fragment shaders
The specification uses the terminology "pixel shader" in a few places but does not define them. I believe this is a crossover from DirectX and the intent was to write "fragment shader".

As part of describing the `clipped` member of `VkSwapchainCreateInfoKHR`:

> If set to VK_TRUE, the presentable images associated with the swapchain may not own all of
their pixels. Pixels in the presentable images that correspond to regions of the target surface
obscured by another window on the desktop, or subject to some other clipping mechanism
will have undefined content when read back. **Pixel shaders** may not execute for these pixels,
and thus any side effects they would have had will not occur. VK_TRUE value does not
guarantee any clipping will occur, but allows more optimal presentation methods to be used
on some platforms.

In a note just further below:

> Applications should set this value to VK_TRUE if they do not expect to read back the
content of presentable images before presenting them or after reacquiring them,
and if their **pixel shaders** do not have any side effects that require them to run for
all pixels in the presentable image.

polarina:
Fixes #1082.

### DIFF
--- a/chapters/VK_KHR_swapchain/wsi.txt
+++ b/chapters/VK_KHR_swapchain/wsi.txt
@@ -205,7 +205,7 @@ Usage requirements.
      target surface obscured by another window on the desktop, or subject to
      some other clipping mechanism will have undefined: content when read
      back.
-     Pixel shaders may: not execute for these pixels, and thus any side
+     Fragment shaders may: not execute for these pixels, and thus any side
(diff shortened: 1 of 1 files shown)


TomOlson:
Thanks! LGTM

Issue #395, opened 2016-10-12
Inconsistency in the required support for StorageImageExtendedFormats


Dolkar:
From reading the SPIR-V spec and the list of physical device features, it's implied that the support for using images with certain formats as storage images is optional. To be able to use them, the _shaderStorageImageExtendedFormats_ feature needs to be enabled:

> _shaderStorageImageExtendedFormats_ indicates whether the extended storage image formats are available in shader code. If this feature is not enabled, the formats requiring the StorageImageExtendedFormats capability are not supported for storage images.

However, in the Required Format Support section of the Vulkan spec, many of the formats that are supposed to only be available if the _shaderStorageImageExtendedFormats_ feature is enabled have the _VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT_ checked, meaning they are required regardless of the presence of the feature.

For example, as per the SPIR-V spec, the **Rg32f** format requires the _StorageImageExtendedFormats_ shader capability, which in turn needs a device with the _shaderStorageImageExtendedFormats_ feature enabled. According to the Vulkan spec, though, the matching **VK_FORMAT_R32G32_SFLOAT** format is always required to be available for use with storage images, with no footnote or reference to the feature in question.

oddhack:
Closing after merging #1098.

kvark:
Just to get some clarification, if I want to use `Rg32f` storage format, I still have to enable this SPIR-V capability, even if my logical device doesn't have `shaderStorageImageExtendedFormats` feature enabled. Is that right?
If that's right, it's unfortunately very confusing, since the features are named similarly, but in this case I'd have to enable one but not the other...

PR #1098, opened 2019-11-23
Clarify shaderStorageImageExtendedFormats feature

krOoze:
Fix suggestion for #395.

- Add the formats to Required Format tables, with precondition on the feature
- Update feature description with explicitly listed format
- Update description with note about enabling being optional
- Update description with note about discrepancy vs SPIR-V formats

### DIFF
--- a/chapters/features.txt
+++ b/chapters/features.txt
@@ -548,14 +548,53 @@ supported properties of individual formats as normal.
     code:ImageGatherExtended capability.
   * [[features-shaderStorageImageExtendedFormats]]
     pname:shaderStorageImageExtendedFormats specifies whether all the
-    extended storage image formats are available in shader code.
-    If this feature is enabled then the
-    ename:VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT feature must: be supported in
-    pname:optimalTilingFeatures for all of the extended formats.
-    To query for additional properties, or if the feature is not enabled,
-    flink:vkGetPhysicalDeviceFormatProperties and
-    flink:vkGetPhysicalDeviceImageFormatProperties can: be used to check for
-    supported properties of individual formats as normal.
+    "`storage image extended formats`" below are supported; if this feature
+    is supported, then the ename:VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT must:
+    be supported in pname:optimalTilingFeatures for the following formats:
++
+  ** ename:VK_FORMAT_R16G16_SFLOAT
+  ** ename:VK_FORMAT_B10G11R11_UFLOAT_PACK32
+  ** ename:VK_FORMAT_R16_SFLOAT
+  ** ename:VK_FORMAT_R16G16B16A16_UNORM
+  ** ename:VK_FORMAT_A2B10G10R10_UNORM_PACK32
+  ** ename:VK_FORMAT_R16G16_UNORM
+  ** ename:VK_FORMAT_R8G8_UNORM
+  ** ename:VK_FORMAT_R16_UNORM
+  ** ename:VK_FORMAT_R8_UNORM
+  ** ename:VK_FORMAT_R16G16B16A16_SNORM
+  ** ename:VK_FORMAT_R16G16_SNORM
+  ** ename:VK_FORMAT_R8G8_SNORM
+  ** ename:VK_FORMAT_R16_SNORM
+  ** ename:VK_FORMAT_R8_SNORM
+  ** ename:VK_FORMAT_R16G16_SINT
+  ** ename:VK_FORMAT_R8G8_SINT
+  ** ename:VK_FORMAT_R16_SINT
+  ** ename:VK_FORMAT_R8_SINT
+  ** ename:VK_FORMAT_A2B10G10R10_UINT_PACK32
+  ** ename:VK_FORMAT_R16G16_UINT
+  ** ename:VK_FORMAT_R8G8_UINT
+  ** ename:VK_FORMAT_R16_UINT
+  ** ename:VK_FORMAT_R8_UINT
++
+[NOTE]
+.Note
+====
+pname:shaderStorageImageExtendedFormats feature only adds a guarantee of
+format support, which is specified for the whole physical device.
+Therefore enabling or disabling the feature via flink:vkCreateDevice has no
+practical effect.
+
+To query for additional properties, or if the feature is not supported,
+flink:vkGetPhysicalDeviceFormatProperties and
+flink:vkGetPhysicalDeviceImageFormatProperties can: be used to check for
+supported properties of individual formats, as usual rules allow.
+
+ename:VK_FORMAT_R32G32_UINT, ename:VK_FORMAT_R32G32_SINT, and
+ename:VK_FORMAT_R32G32_SFLOAT from code:StorageImageExtendedFormats SPIR-V
+capability, are already covered by core Vulkan
+<<formats-mandatory-features-32bit,mandatory format support>>.
+====
+
   * [[features-shaderStorageImageMultisample]]
     pname:shaderStorageImageMultisample specifies whether multisampled
     storage images are supported.
--- a/chapters/formats.txt
+++ b/chapters/formats.txt
@@ -2634,6 +2634,8 @@ endif::VK_VERSION_1_1,VK_KHR_maintenance1[]
 ^|{sym2} | This feature must: be supported on at least some
 of the named formats, with more information in the table
 where the symbol appears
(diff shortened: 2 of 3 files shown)


appendices/spirvenv.txt
```diff
@@ -304,6 +304,21 @@ ifdef::VK_EXT_shader_demote_to_helper_invocation[]
...
+1::
+    <<features-shaderStorageImageExtendedFormats,pname:shaderStorageImageExtendedFormats>>
+    feature is not strictly necessary to be enabled to use
+    code:StorageImageExtendedFormats SPIR-V capability.
+    _Extended storage image formats_ may: support
+    `VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT` regardless of
+    pname:shaderStorageImageExtendedFormats being supported or enabled.
+    But the application must: enable the feature if it relies only on the
+    <<features-required-format-support,required format support>> of the
+    _extended storage image formats_.
```
jekstrand: I think the first two sentences are fine but I don't know what you mean by the third.

chapters/formats.txt
```diff
@@ -2902,16 +2916,20 @@ s| Format
...
-| ename:VK_FORMAT_R32G32B32_SINT             |   |   |   |   |   |   |   |   |   | {sym1} |   |   |
-| ename:VK_FORMAT_R32G32B32_SFLOAT           |   |   |   |   |   |   |   |   |   | {sym1} |   |   |
-| ename:VK_FORMAT_R32G32B32A32_UINT          | {sym1} | {sym1} |   | {sym1} |   | {sym1} | {sym1} |   |   | {sym1} | {sym1} | {sym1} |
-| ename:VK_FORMAT_R32G32B32A32_SINT          | {sym1} | {sym1} |   | {sym1} |   | {sym1} | {sym1} |   |   | {sym1} | {sym1} | {sym1} |
-| ename:VK_FORMAT_R32G32B32A32_SFLOAT        | {sym1} | {sym1} |   | {sym1} |   | {sym1} | {sym1} |   |   | {sym1} | {sym1} | {sym1} |
+| ename:VK_FORMAT_R32_SFLOAT                 | {sym1} | {sym1} |   | {sym1} |        | {sym1} | {sym1} |   |   | {sym1} | {sym1} | {sym1} |
+| ename:VK_FORMAT_R32G32_UINT                | {sym1} | {sym1} |   | {sym3} |        | {sym1} | {sym1} |   |   | {sym1} | {sym1} | {sym1} |
+| ename:VK_FORMAT_R32G32_SINT                | {sym1} | {sym1} |   | {sym3} |        | {sym1} | {sym1} |   |   | {sym1} | {sym1} | {sym1} |
+| ename:VK_FORMAT_R32G32_SFLOAT              | {sym1} | {sym1} |   | {sym3} |        | {sym1} | {sym1} |   |   | {sym1} | {sym1} | {sym1} |
```
jekstrand: This isn't correct and I'm afraid it somewhat substantially impacts this MR.  We had discussions internally recently about this and re-confirmed that we want the R32G32 formats to be always required.  This means that there is a mismatch between Vulkan's concept of extended formats and the SPIR-V concept, which is a bit unfortunate.  In my view, the right thing to do here is probably to drop the glossary definition and make the table footnotes only refer to the feature bit.

appendices/spirvenv.txt
krOoze: Hm thought I removed that. I will reremove it later. It relates to the confusing part I relay in the "BTW" point in the OP.

chapters/formats.txt
krOoze: I reverted them back to being unconditionally supported

oddhack:
@krooze  did you address @jekstrand's issues? It looks like a merge conflict has affected the PR since then and it's a bit hard to tell what happened.

jekstrand:
> @krOoze did you address @jekstrand's issues? It looks like a merge conflict has affected the PR since then and it's a bit hard to tell what happened.

My bad.  I've not yet taken the time to review his updated version.

krOoze:
@oddhack That should be so. I just need to rebase.

appendices/spirvenv.txt
```diff
@@ -304,6 +304,18 @@ ifdef::VK_EXT_shader_demote_to_helper_invocation[]
...
+[NOTE]
+.Note
+====
+1::
+    <<features-shaderStorageImageExtendedFormats,pname:shaderStorageImageExtendedFormats>>
+    feature is not strictly necessary to be enabled to use
+    code:StorageImageExtendedFormats SPIR-V capability.
+    _Extended storage image formats_ may: support
+    `VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT` regardless of
+    pname:shaderStorageImageExtendedFormats being supported or enabled.
```
jekstrand: I've been rolling this one around in my head and I don't think this note clarifies anything.  The feature doesn't have to be set because it's purely informative.  This is just like `textureCompressionETC2` which you don't have to enable to use ETC2 formats.  If anything, saying it's optional and adding a note is just going to confuse people.

krOoze: I think optionally enablable features are inherently confusing, given usual features do not work that way. If enabling it does nothing it should be noted _somewhere_. Maybe I could say that if it is supported, it is always enabled implicitly?

jekstrand: I 100% agree  The text for `textureCompressionETC2` currently says:

> `textureCompressionETC2` specifies whether all of the ETC2 and EAC compressed texture formats are supported. If this feature is enabled, then the `VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT`, `VK_FORMAT_FEATURE_BLIT_SRC_BIT` and `VK_FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT` features must be supported in optimalTilingFeatures for the following formats:

However, those bits are queried by `vkGetPhysicalDeviceFormatProperties` which is a physical device function and therefore not subject to feature enables.  I think it's better to say that the features are implicitly enabled if they are available.  Maybe @jbreton-nv or @Tobski have an opinion here?

chapters/features.txt
```diff
@@ -548,14 +548,16 @@ supported properties of individual formats as normal.
     code:ImageGatherExtended capability.
   * [[features-shaderStorageImageExtendedFormats]]
     pname:shaderStorageImageExtendedFormats specifies whether all the
-    extended storage image formats are available in shader code.
-    If this feature is enabled then the
-    ename:VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT feature must: be supported in
-    pname:optimalTilingFeatures for all of the extended formats.
-    To query for additional properties, or if the feature is not enabled,
+    SPIR-V's "`extended storage image formats`" are available in shader
+    code.
```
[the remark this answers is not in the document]
krOoze: Sure, the format list is teensy bit long though. I guess we can get away with this with features, which are each basically their own section with an anchor...

appendices/spirvenv.txt
krOoze: ok I tried something else.

chapters/features.txt
```diff
@@ -548,14 +548,53 @@ supported properties of individual formats as normal.
...
+  ** ename:VK_FORMAT_R8G8_UINT
+  ** ename:VK_FORMAT_R16_UINT
+  ** ename:VK_FORMAT_R8_UINT
++
+[NOTE]
+.Note
+====
+pname:shaderStorageImageExtendedFormats feature only adds guarantee of
+format support, which is specified for the whole physical device.
+Therefore enabling or disabling the feature via flink:vkCreateDevice has not
```
jekstrand: Maybe "has no practical effect" instead of "not"?  Other than that, I really like it.

chapters/features.txt
```diff
@@ -548,14 +548,53 @@ supported properties of individual formats as normal.
...
+  ** ename:VK_FORMAT_A2B10G10R10_UINT_PACK32
+  ** ename:VK_FORMAT_R16G16_UINT
+  ** ename:VK_FORMAT_R8G8_UINT
+  ** ename:VK_FORMAT_R16_UINT
+  ** ename:VK_FORMAT_R8_UINT
++
+[NOTE]
+.Note
+====
+pname:shaderStorageImageExtendedFormats feature only adds guarantee of
```
jekstrand: One more language nit: "adds a guarantee".

oddhack:
@krOoze FYI I'm going to be accepting your approved PRs now, but there won't be a new spec update published until sometime after the holiday break - we're going on some WG downtime with many people and companies unavailable.

PR #1101, opened 2019-11-24
Fix minImageCount VU wrt the shared present modes

krOoze:
Most of the drivers seem to report `minImageCount==2`, which would make the swapchain uncreatable due to the another VU requiring the count to be 1 for shared present modes.


oddhack:
Assigned to SI TSG for triage.

PR #1121, opened 2019-11-25
Hide outdated VU in face of timeline semaphores

Tobski:
The following VU added by VK_KHR_timeline_semaphore is accurate, and covers the same invalid cases as this does, but factoring in timeline semaphores.


krOoze:
The term "Binary Semaphore" is established in Glossary. That could be used to make a singular VU.

Tobski:
@krOoze 

> The term "Binary Semaphore" is established in Glossary. That could be used to make a singular VU.

The WG agreed that this would be possible, but figured it'd be better to stay consistent for now and just get this in before it is wrong for too long, and then address this type of change in a future consistency fix. If you wanted to raise an issue, or just tackle it I'd be happy to contribute/review.

PR #1133, opened 2019-12-07
Move extension pipe stages so they are not part of ALL_GRAPHICS

krOoze:


chapters/pipelines.txt
```diff
@@ -504,16 +504,16 @@ include::{generated}/api/enums/VkShaderStageFlagBits.txt[]
...
   * ename:VK_SHADER_STAGE_FRAGMENT_BIT specifies the fragment stage.
   * ename:VK_SHADER_STAGE_COMPUTE_BIT specifies the compute stage.
-ifdef::VK_NV_mesh_shader[]
-  * ename:VK_SHADER_STAGE_TASK_BIT_NV specifies the task stage.
-  * ename:VK_SHADER_STAGE_MESH_BIT_NV specifies the mesh stage.
-endif::VK_NV_mesh_shader[]
   * ename:VK_SHADER_STAGE_ALL_GRAPHICS is a combination of bits used as
     shorthand to specify all graphics stages defined above (excluding the
     compute stage).
   * ename:VK_SHADER_STAGE_ALL is a combination of bits used as shorthand to
```
oddhack: Maybe it would also make sense to move SHADER_STAGE_ALL to the very end of the list to indicate it includes all the stages above.

oddhack: Cancel that, we decided to take it in this form.

Tobski: Except it's in the right place because again, it is an explicit bitmask. We should probably consider deprecating these values...

krOoze: @oddhack I should probably move them in a way that it matches the declaration order though. Is this already merged internally, or can I still mess with this?

oddhack: I'd like to merge this in the approved form and get it off the table, but you could propose another PR rearranging it further. We won't be discussing new PRs until WG activities pick up again in January.

PR #1134, opened 2019-12-07
memory.txt: Clarify memory properties

gurchetansingh:
This may be a bit obvious, but we want to make sure the imported
memory, if it is mapped, as the same memory properties of the
corresponding memory index.

Again, let me know if this is too obvious given VkMemoryFdPropertiesKHR.

### DIFF
--- a/chapters/memory.txt
+++ b/chapters/memory.txt
@@ -1888,6 +1888,11 @@ times into a given Vulkan instance.
 In all cases, each import operation must: create a distinct
 sname:VkDeviceMemory object.
 
+If the import is successful, the implementation must: ensure that any successful
+subsequent call to fname:vkMapMemory results in a host visible pointer that is
+consistent with memory properties corresponding to pname:memoryTypeIndex of the
+slink:VkMemoryAllocateInfo structure.
+
 .Valid Usage
 ****
   * [[VUID-VkImportMemoryFdInfoKHR-handleType-00667]]


chapters/memory.txt
```diff
@@ -1888,6 +1888,10 @@ times into a given Vulkan instance.
 In all cases, each import operation must: create a distinct
 sname:VkDeviceMemory object.
 
+The implementation must: ensure the any subsequent mappings of imported memory
```
krOoze: "the any"
"mappings" sounds ambiguous

krOoze:
One thing I don't understand now is what this got to do with `vkMapMemory` at all.

Maybe should say generally that imported memory does indeed inherit the capabilities of the `memoryType` it was imported as.

gurchetansingh:
> One thing I don't understand now is what this got to do with `vkMapMemory` specifically.

Mainly, we want to ensure `vkMapMemory` returns uncached memory when the memory index points to uncached memory type.  

> Maybe should say generally that imported memory does indeed inherit the capabilities of the `memoryType` it was imported as. Could be a Note; that indeed feels bit obvious given the Memory Type is a parameter to the allocate\import call.

That's true.  If it's too obvious, feel free to close.

cubanismo:
This does seem like it's already clear.  I don't know why anyone would think they can create mappings inconsistent with the memory type specified at import time.  A note could be added if needed I suppose, but this really feels like it's covered by valid usage of the memory imported being compatible with the memory type specified.

stonesthrow:
We are rejecting this change as being redundant. But, it could be a note.

gurchetansingh:
Cool, thanks.  Glad everyone seems to agree the spec is clear on this.

PR #1135, opened 2019-12-07
Clarify surface queries intro paragraph

### LINKED ISSUE #1132: Unclear meaning of independent capabilities
Spec says:

> Capabilities that correspond to image creation parameters are not
independent of each other: combinations of parameters that are not supported
as reported by `vkGetPhysicalDeviceImageFormatProperties` are not
supported by the surface on that physical device, even if the capabilities
taken individually are supported as part of some other parameter
combinations.

I am unable to parse this, and what it is trying to convey. I think it conflates two or three concepts making it illegible.

krOoze:
fix #1132

### DIFF
--- a/chapters/VK_KHR_surface/wsi.txt
+++ b/chapters/VK_KHR_surface/wsi.txt
@@ -275,15 +275,19 @@ endif::VK_NN_vi_surface[]
 The capabilities of a swapchain targeting a surface are the intersection of
 the capabilities of the WSI platform, the native window or display, and the
 physical device.
-The resulting capabilities can be obtained with the queries listed below in
+The resulting capabilities can: be obtained with the queries listed below in
 this section.
-Capabilities that correspond to image creation parameters are not
-independent of each other: combinations of parameters that are not supported
-as reported by flink:vkGetPhysicalDeviceImageFormatProperties are not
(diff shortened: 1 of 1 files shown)


chapters/VK_KHR_surface/wsi.txt
```diff
@@ -275,15 +275,19 @@ endif::VK_NN_vi_surface[]
...
+[NOTE]
+.Note
+====
+In addition to the surface capabilities as obtained by surface queries
+below, swapchain images are also subject to ordinary image creation limits
+as reported by flink:vkGetPhysicalDeviceImageFormatProperties.
+As an application is instructed by the appropriate Valid Usage sections,
+both the surface capabilities and the image creation limits have to be
+satisfied whenever swapchain images are created.
+====
```
cubanismo: I think this should remain normative language.  Do you believe the spec is clear enough for a discerning reader in the absence of this note?  To me making this language a note leaves the actual spec language even more ambiguous than before.

cubanismo: In call, @Tobski found the VU stating the actual requirement normatively already, so I retract this.  I agree a note is appropriate.

PR #1148, opened 2019-12-31
Fix vkGetPhysicalDeviceToolPropertiesEXT

krOoze:
- Fix returned array implicit VUs
diff
- Rewrite text to be more consistent (based on `vkGetPhysicalDeviceQueueFamilyProperties`)
- Explicitly state variability (aka "lifetime") of retrieved results. (I assume active tools list can change; e.g. tools being switched on and off outside the Vulkan scope.)


Tobski:
@oddhack this is fine to merge in whenever.

PR #1149, opened 2019-12-31
Fix c arrow pnames

krOoze:


oddhack:
We picked this up internally in the 1.2.131 spec update, so the PR should be moot now - closing.

PR #1152, opened 2020-01-04
Clarify command buffer state

krOoze:
- do not use "reverts back" with invalid state
- clear out ambiguities in primary-secondary lifecycle linkage
  - include the case when completion moves to the invalid state
  - clarify that moving secondary between the pending and
     execution state does not invalidate any other primary it
     may be recorded in

### DIFF
--- a/chapters/cmdbuffers.txt
+++ b/chapters/cmdbuffers.txt
@@ -63,8 +63,8 @@ Each command buffer is always in one of the following states:
 Initial::
     When a command buffer is <<vkAllocateCommandBuffers, allocated>>, it is
     in the _initial state_.
-    Some commands are able to _reset_ a command buffer, or a set of command
-    buffers, back to this state from any of the executable, recording or
+    Some commands are able to _reset_ a command buffer (or a set of command
+    buffers) back to this state from any of the executable, recording or
     invalid state.
     Command buffers in the initial state can: only be moved to the recording
     state, or freed.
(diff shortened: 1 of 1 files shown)


chapters/cmdbuffers.txt
```diff
@@ -106,25 +107,33 @@ on what state a command buffer must: be in, which are detailed in the valid
...
-within it, and once all executions of each command buffer complete, they
-move to the executable state.
-If a secondary moves to any other state whilst it is recorded to another
-command buffer, the primary moves to the invalid state.
-A primary moving to any other state does not affect the state of the
-secondary.
-Resetting or freeing a primary command buffer removes the linkage to any
-secondary command buffers that were recorded to it.
+recorded to it move to the _pending state_.
+Once execution of the primary completes, so does for any secondary recorded
```
janharaldfredriksen-arm: nit: remove "for"

krOoze: Not sure I can. Execution completes **for** any secondary.

krOoze: I opted for "so it does for any". good?

janharaldfredriksen-arm: yep, that works.

chapters/cmdbuffers.txt
```diff
@@ -106,25 +107,33 @@ on what state a command buffer must: be in, which are detailed in the valid
...
-If a secondary moves to any other state whilst it is recorded to another
-command buffer, the primary moves to the invalid state.
-A primary moving to any other state does not affect the state of the
-secondary.
-Resetting or freeing a primary command buffer removes the linkage to any
-secondary command buffers that were recorded to it.
+recorded to it move to the _pending state_.
+Once execution of the primary completes, so it does for any secondary
+recorded within it.
+And once all executions of each command buffer complete, they each move to
```
Tobski: Nit: Shouldn't start sentence with "And". Maybe: "After execution of a command buffer completes, it moves to its appropriate..."

krOoze: Done.
I kept the plural, because the spec tries to make an important point there with it. Though not sure it is accented enough...

PR #1153, opened 2020-01-05
Remove redundant command buffer SIMULTANEOUS_USE language

krOoze:
- remove VUID requiring primary of a secondary to be pending. I think by definition secondary is pending if primary is, and there is already secondary specific VUID.
- remove VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT language covered by VUIDs

### DIFF
--- a/chapters/cmdbuffers.txt
+++ b/chapters/cmdbuffers.txt
@@ -1585,24 +1585,19 @@ command buffer becomes <<commandbuffers-lifecycle, invalid>>.
     pname:level of ename:VK_COMMAND_BUFFER_LEVEL_SECONDARY
   * [[VUID-vkCmdExecuteCommands-pCommandBuffers-00089]]
     Each element of pname:pCommandBuffers must: be in the
-    <<commandbuffers-lifecycle, pending or executable state>>.
-  * [[VUID-vkCmdExecuteCommands-pCommandBuffers-00090]]
-    If any element of pname:pCommandBuffers was not recorded with the
-    ename:VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT flag, and it was
-    recorded into any other primary command buffer, that primary command
-    buffer must: not be in the <<commandbuffers-lifecycle, pending state>>
+    <<commandbuffers-lifecycle, pending or executable state>>
   * [[VUID-vkCmdExecuteCommands-pCommandBuffers-00091]]
     If any element of pname:pCommandBuffers was not recorded with the
     ename:VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT flag, it must: not be
-    in the <<commandbuffers-lifecycle, pending state>>.
+    in the <<commandbuffers-lifecycle, pending state>>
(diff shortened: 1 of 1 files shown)


chapters/cmdbuffers.txt
```diff
@@ -761,14 +761,6 @@ time.
...
-creating a command buffer, that command buffer must: not be submitted to a
-queue whilst it is already in the <<commandbuffers-lifecycle, pending
-state>>.
-If ename:VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT is not set on a
-secondary command buffer, that command buffer must: not be used more than
-once in a given primary command buffer.
-
 [NOTE]
 .Note
 ====
```
krOoze: BTW, not quite sure what this note is saying about some "patching"...

polarina: @krOoze Patching refers to the driver modifying the data and executable code recorded into the command buffer at queue-submission time or during execution.

krOoze: @polarina Sorry, I know what patching means. What I do not understand is is why it is brought up here with relation to `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT`. One would assume the copy is the real performance problem. Whether one patches the original or the copy as part of the submission, that part is the same, no?

polarina: @krOoze Correct, the copy is the performance problem. It is the use of `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` that forces that copy, the implementation can not patch in-place which it could otherwise do without that bit set.

krOoze: oh, right, the patching is a write access that necessitates the copy in the first place.

chapters/cmdbuffers.txt
```diff
@@ -761,14 +761,6 @@ time.
...
 --
 
-If ename:VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT was not set when
-creating a command buffer, that command buffer must: not be submitted to a
-queue whilst it is already in the <<commandbuffers-lifecycle, pending
-state>>.
-If ename:VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT is not set on a
-secondary command buffer, that command buffer must: not be used more than
-once in a given primary command buffer.
-
```
Tobski: I don't think we should remove this - there's some plan to change/move instances of this type of language fairly globally at some point; but they were added for a reason. This is almost always as a highlight for non-obvious behavior which we expect developers to trip up on; I would be ok changing these to notes though.

krOoze: I dunno, feels obvious because this just reiterates what `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` is defined as.

Tobski: Ah yes, looking at it I think you're right - so I'm fine with accepting this.

krOoze: I already preemptively made it a note. But I cannot pass on deleting stuff, which is what I think you are agreeing with here. So I will do that when I get to PC

PR #1154, opened 2020-01-05
Prevent primary with both ONE_TIME and SIMULTANEOUS_USE

krOoze:
Having primary command buffer with both `VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT` and `VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT` does not make sense. Cannot submit multiple times to make use of simultaneous flag.

For secondaries `SIMULTANEOUS` controls whether they can be recorded multiple times, so the situation is different.

### DIFF
--- a/chapters/cmdbuffers.txt
+++ b/chapters/cmdbuffers.txt
@@ -589,12 +589,12 @@ include::{generated}/api/protos/vkBeginCommandBuffer.txt[]
 ****
   * [[VUID-vkBeginCommandBuffer-commandBuffer-00049]]
     pname:commandBuffer must: not be in the <<commandbuffers-lifecycle,
-    recording or pending state>>.
+    recording or pending state>>
   * [[VUID-vkBeginCommandBuffer-commandBuffer-00050]]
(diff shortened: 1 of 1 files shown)


chrisforbes:
Hang on -- this converts a "probably not useful but well-defined" corner into UB. Today it's perfectly valid to create a primary with both of these flags, and submit it once.

krOoze:
@chrisforbes Yes. Another option could be Best Practices Layer check.

Tobski:
Corner cases like this are usually inadvertent rather than deliberate, and are definitely not tested - it’s thus unclear that this would actually work. So yes you’re right it’s technically defined - but we didn’t really intend it to be. Hence supporting this change.

Tobski:
The Vulkan WG discussed and agreed with my response - in cases like this the choice boils down to "write CTS tests for it" or "make it invalid and write validation for it". We didn't see value in ensuring this behavior worked across all implementations (via CTS testing), so are happy with this change.

PR #1157, opened 2020-01-08
Add anchor handles to chapters and VUIDs

krOoze:
- add anchor handles to chapters
   - chunked already had them sneaked in with the build script, so replace with the asciidoctor feature. Difference is it is on the left of the chapter header instead. Also the symbol is different (but that can be changed in the CSS).
- add similar anchor handles to each VU

It is pretty standard on modern websites (such as here on GitHub when viewing md file). Helps people to link to stuff without having to fish inside the html source for the anchor ids. Additionally makes the VUID of VU more visible (hovering over link usually shows the URL, as well as clicking it should apply it to the address bar).

preview: http://vulkan.hys.cz/orig/pr1157/vkspec.html
hover mouse over a chapter header, or over a VU statement to see a clickable link symbol.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,6 +212,9 @@ GENDEPENDS     = $(APIDEPEND) $(VALIDITYDEPEND) $(HOSTSYNCDEPEND) $(METADEPEND)
 # All non-format-specific dependencies
 COMMONDOCS     = $(SPECFILES) $(GENDEPENDS)
 
+# Script to add href to anchors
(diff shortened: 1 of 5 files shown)


oddhack:
This is one of those cases where it would help us if you would describe what the functional improvement or change resulting from this PR is - how does it help someone reading the document?

krOoze:
@oddhack edited some info into the OP.

Tobski:
Ok now I know what this is I am very happy to see this change - thanks @krOoze! This will save some lookup headaches 🎉

PR #1164, opened 2020-01-15
Update vkGet*ProcAddr example usage

Jormungardian:
Updating the example usage to align closer to spec defining this as an instance extension.
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_debug_utils.html

### DIFF
--- a/appendices/VK_EXT_debug_utils.txt
+++ b/appendices/VK_EXT_debug_utils.txt
@@ -138,8 +138,8 @@ happens and the third will log warnings to stdout.
     VkDebugUtilsMessengerEXT cb1, cb2, cb3;
 
     // Must call extension functions through a function pointer:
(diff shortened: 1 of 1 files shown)


pdaniell-nv:
This is the correct change, but you have to be a little careful you don't bump into this loader bug:
https://github.com/KhronosGroup/Vulkan-Loader/issues/116

Jormungardian:
@pdaniell-nv Thanks for looking at this and thanks for making me aware of that regression. Through my testing on different vendor configs and drivers I did see different behaviors when getting this function pointers with vkGetDeviceProcAddr (some would return valid pointers and some wouldn't) which I wasn't sure what to attribute it to.

oddhack:
@jekstrand I think we agreed to close this once internal !3567 was published, which has happened - do you want to take care of that?

jekstrand:
Yup.  The new spec language makes the old text valid (and probably preferable for performance).

PR #1168, opened 2020-01-29
Reserve Codeplay vendor ID

EwanC:
Codeplay Software Ltd is a ISV, and as such doesn't have a PCIe ID to use for the `VkPhysicalDeviceProperties::vendorID` in our [ComputeAorta](https://codeplay.com/products/computesuite/computeaorta) product.

Reserve the next free ID from `VkVendorId` in vk.xml


EwanC:
Do I need to make a second matching PR against https://github.com/KhronosGroup/Vulkan-Headers/blob/master/registry/vk.xml ?

oddhack:
@EwanC you do not. Vulkan-Headers gets updated when a new spec update is generated. However, be aware that we probably won't be doing a spec update until February 17th, due to the upcoming F2F meeting. Also, you need to execute the CLA before this can be merged. Since Codeplay is a Khronos member, it's a bit easier if you submit the MR against the internal vulkan gitlab repo, but we can work with this.

PR #1169, opened 2020-01-30
Add hover text with VUID to html

krOoze:
hovertext

### DIFF
--- a/scripts/genanchorlinks.py
+++ b/scripts/genanchorlinks.py
@@ -25,7 +25,7 @@ def genAnchorLinks(in_file, out_file):
 		print('Error: File %s does not exist.' % in_file)
 		sys.exit(2)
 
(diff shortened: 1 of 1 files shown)


oddhack:
We're having some internal discussion about maybe imbedding the VUID names as visible text for searchability. Putting this on hold while we decide a way forward.

krOoze:
@oddhack That seems bit too spammy to me.
With https://stackoverflow.com/questions/25512436/searching-for-text-ctrlf-across-hidden-spans it should be possible to make a hidden full-text search trap. Additionally the VUIDs should be indexed in the chunked search box, and the search box should be inplemented for single-page too.

oddhack:
I have suggested that as well. There are a lot of incompatible opinions on the goal of making it possible for validation layer error reports to land on useful surrounding spec context in reasonable amounts of time.

I discovered the hidden text search thing by accident while searching for something else and eventually realized I was landing on hidden latexmath source markup.

PR #1182, opened 2020-02-17
Refer to attachments instead of attachment references

krychu:
Should the description refer to attachments rather than attachment references? I suspect that two RenderPasses are not compatible if their attachments are different but references inside Subpasses still happen to be the same.

### DIFF
--- a/chapters/renderpass.txt
+++ b/chapters/renderpass.txt
@@ -2602,8 +2602,8 @@ If the arrays are of different lengths, attachment references not present in
 the smaller array are treated as ename:VK_ATTACHMENT_UNUSED.
 
 Two render passes are compatible if their corresponding color, input,
-resolve, and depth/stencil attachment references are compatible and if they
-are otherwise identical except for:
+resolve, and depth/stencil attachments are compatible and if they are
+otherwise identical except for:
 
   * Initial and final image layout in attachment descriptions
   * Load and store operations in attachment descriptions


krOoze:
I am afraid it probably does not work that way. References have to be compatible for the render passes to be compatible. At least the whole chapter revolves around that.

As for attachments themselves, there is:

> Two render passes are compatible [...] if they are otherwise identical

krychu:
> Two render passes are compatible if their corresponding color, input, resolve, and depth/stencil attachment references are compatible and if they are otherwise identical [...]

Reading this was a bit confusing to me for the following reasons:
- It's not clear what a "render pass' corresponding attachment reference" is. Render pass doesn't include references; only via subpasses where to correspondence is between a reference and an attachment.
- Assuming that the above sentence is correct we need to assume it refers to attachment references within subpasses. But then the meaning of compatibility of attachment references between two render passes is not clear, imho. Does it mean two render passes must have the same number of subpasses, all with exactly the same references? Or one render pass can have a subset of subpasses, all with exactly the same references? Or one render pass can have a subset of subpasses with a subset of references, all with exactly the same references?

krOoze:
> It's not clear what "the render pass' corresponding attachment references" are.

Very likely "corresponding" means "at the same index in the array".

> Assuming that the quoted sentence is correct we need to assume it refers to attachment references within subpasses.

Yes, I am assuming just that. Otherwise the whole chapter would need a rewrite as a whole, because it seems to assume that also.

> Does it mean two render passes must have the same number of subpasses, all with exactly the same references?

I would think so, except the special case stated in the chapter.

krychu:
This makes sense yes. I've read the spec section a couple of times now after this convo and I think it's indeed clear. @krOoze thanks for your time, I'll close the ticket now.

PR #1185, opened 2020-02-18
Consolidate pipeline dynamic state VUs

### LINKED ISSUE #1180: Missing attribute "noautovalidity" for pShadingRatePalettes and pExclusiveScissors
Hi,
I see there's no attribute "noautovalidity" for fields **VkPipelineViewportShadingRateImageStateCreateInfoNV::pShadingRatePalettes** and **VkPipelineViewportExclusiveScissorStateCreateInfoNV::pExclusiveScissors**. However other dynamic states like pViewports, pScissors, pViewportWScalings and pDiscardRectangles, they all have **noautovalidity="true"** in the member description in xml. So I think pShadingRatePalettes and pExclusiveScissors shall be in the same way.

krOoze:
- unify the language
- fix #1180 
- add some VUs that feel missing
- remove VUs that are redundant or at wrong place

Some problems remain. Not sure how to deal with `VkRect2D` dynamic states and their non-0 and non-overflow requirements. There's also bunch of `VK_DYNAMIC_STATE_SAMPLE_LOCATIONS_EXT` validity problems remaining that I did not want to get into.

### DIFF
--- a/chapters/fragops.txt
+++ b/chapters/fragops.txt
@@ -359,22 +359,8 @@ to be `0` and the exclusive scissor test is disabled.
 
 .Valid Usage
 ****
-  * [[VUID-VkPipelineViewportExclusiveScissorStateCreateInfoNV-exclusiveScissorCount-02027]]
-    If the <<features-multiViewport,multiple viewports>> feature is not
-    enabled, pname:exclusiveScissorCount must: be `0` or `1`
-  * [[VUID-VkPipelineViewportExclusiveScissorStateCreateInfoNV-exclusiveScissorCount-02028]]
-    pname:exclusiveScissorCount must: be less than or equal to
-    sname:VkPhysicalDeviceLimits::pname:maxViewports
-  * [[VUID-VkPipelineViewportExclusiveScissorStateCreateInfoNV-exclusiveScissorCount-02029]]
-    pname:exclusiveScissorCount must: be `0` or identical to the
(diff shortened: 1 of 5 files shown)


pdaniell-nv:
There is a large unreleased change in the pipeline that will conflict with a lot of this. Let's keep this open until the conflicting changes are released, and then this can be rebased.

oddhack:
@krOoze I believe the "large unreleased change" @pdaniell-nv referred to was VK_EXT_extended_dynamic_state, which is now public. So this can probably be progressed now.

krOoze:
@oddhack It turns out it is actually little bit messy to have to have VU at the place where the whole state is known. Would you be open to some system where the VUs can stay at the struct? I think some way to inherit state from the command VU to the struct VU. Might need markup in the `vk.xml`.

oddhack:
> @oddhack It turns out it is actually little bit messy to have to have VU at the place where the whole state is known. Would you be open to some system where the VUs can stay at the struct? I think some way to inherit state from the command VU to the struct VU. Might need markup in the `vk.xml`.

We would rather have the VU in some form than not have relevant ones, but we are gradually fixing places where VU are at the structure level but require access to information only available at the command level or higher-level structure level. What sort of messiness are you encountering? Usually there is some tedious but simple rewriting to refer to parameters of the command directly and refer to members of the structure as part of the command parameters. Sometimes if multiple commands use the structure, partitioning VU into commonvalidity/ is needed.

PR #1190, opened 2020-02-21
Move `UNDEFINED` layout transition note below appropriate paragraph

krOoze:


stonesthrow:
This just moves note up a paragraph to related paragraph - approved.

PR #1196, opened 2020-02-23
Add missing len attribute in vkQueueSignalReleaseImageANDROID

expipiplus1:


krOoze:
The extension is unreleased, so the contents probably is only a placeholder...

PR #1197, opened 2020-03-02
Clarify vector alignment rules

### LINKED ISSUE #1183: Vector alignment in Standard Buffer Layout issue
Spec version 1.2.132. Section 14.5.4. Offset and Stride Assignment, Standard Buffer Layout section.
It defines struct member alignment according to the **first matching rule**.  The first two rules are:

> 1. If the scalarBlockLayout feature is enabled on the device then every member must be aligned according to its scalar alignment.
> 2. All vectors must be aligned according to their scalar alignment.

Consider struct member of vector type. If scalarBlockLayout is disabled then rule 2 is the first matching rule and it says that alignment of float3 is 4, so it can intersect 16 bytes boundary. This is incorrect if scalarBlockLayout is not enabled. 

It looks like *rule 2* is based on vulkan 1.1's VK_KHR_relaxed_block functionality but that extension only states that we can have { float; float3} without padding but still doesn't allow 16 bytes crossing.

kennyalive:
fix #1183

### DIFF
--- a/chapters/interfaces.txt
+++ b/chapters/interfaces.txt
@@ -989,7 +989,7 @@ ifdef::VK_VERSION_1_2,VK_EXT_scalar_block_layout[]
   member must: be aligned according to its scalar alignment.
 endif::VK_VERSION_1_2,VK_EXT_scalar_block_layout[]
 ifdef::VK_VERSION_1_1,VK_KHR_relaxed_block_layout[]
-. All vectors must: be aligned according to their scalar alignment.
+. All vectors must: be aligned according to their scalar alignment and must: not improperly straddle, as defined above.
 endif::VK_VERSION_1_1,VK_KHR_relaxed_block_layout[]
 ifdef::VK_VERSION_1_2,VK_KHR_uniform_buffer_standard_layout[]
 . If the pname:uniformBufferStandardLayout feature is not enabled on the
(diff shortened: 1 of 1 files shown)


chapters/interfaces.txt
```diff
@@ -989,7 +989,7 @@ ifdef::VK_VERSION_1_2,VK_EXT_scalar_block_layout[]
   member must: be aligned according to its scalar alignment.
 endif::VK_VERSION_1_2,VK_EXT_scalar_block_layout[]
 ifdef::VK_VERSION_1_1,VK_KHR_relaxed_block_layout[]
-. All vectors must: be aligned according to their scalar alignment.
+. All vectors must: be aligned according to their scalar alignment and must: not improperly straddle, as defined above.
```
gnl21: This section is only really defining the alignment of the member, and so this doesn't really fit in here. It's confusing because the alignment rules are nearly all the rules, so the rest are easy to miss. Perhaps moving the other rules ahead of the alignment rules instead of this change would make it less confusing?

kennyalive: Ok, I see. Moving ahead doesn't change a lot because the sentense "All vectors must be aligned according to their scalar alignment." sounds like and absolute rule. I thought about dividing rules into two parts a) simple alignment rule for the case when scalarBlockLayout  is enabled. b) all other cases. This could simplify things a bit because in b) we will have 3 rules instead of 4 and scalarBlockLayout will be mentioned only once. From that point it would be possible to move further but actually I don't see a good way how to fix this completely, so I close this.

PR #1201, opened 2020-03-08
Fix parameters section in vkGetPhysicalDeviceXcbPresentationSupportKHR page

Thurstag:
`visual_id` documentation isn't in a separate item.


oddhack:
This should be fixed in the 1.2.135 spec update - it was so short it was  easier to just do in our internal repo then push out. Thanks for the contribution!

PR #1202, opened 2020-03-10
remove extra 'a'

tgfrerer:
Fixes a typo in 3 locations.


oddhack:
Thanks! We're fixing this for the next public spec update, but there are some other changes that cause this PR to conflict, so I just did the equivalent in our internal repo. Next update will be out in a couple of days.

PR #1204, opened 2020-03-13
Simplify Default Non-Dispatchable Handle Def

awr1:
Making the ND handles a struct containing a 64-bit value should make the 64-bit ISA preprocessor branch/pointer defintion unnecessary and also allow the strong typedef to exist on 32-bit platforms.

Related: https://stackoverflow.com/questions/51743500/why-do-non-dispatchable-handles-use-a-ptr-on-64bit-platofrms

The root `#if !defined(VK_DEFINE_NON_DISPATCHABLE_HANDLE)` is kept so as to conform to the spec:

>The `vulkan_core.h` header allows the `VK_DEFINE_NON_DISPATCHABLE_HANDLE` definition to be overridden by the application. If `VK_DEFINE_NON_DISPATCHABLE_HANDLE` is already defined when `vulkan_core.h` is compiled, the default definition is skipped. This allows the application to define a binary-compatible custom handle which may provide more type-safety or other features needed by the application. Applications must not define handles in a way that is not binary compatible - where binary compatibility is platform dependent.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -165,11 +165,7 @@ server.
 
         <type category="define" name="VK_DEFINE_NON_DISPATCHABLE_HANDLE">
 #if !defined(VK_DEFINE_NON_DISPATCHABLE_HANDLE)
-#if defined(__LP64__) || defined(_WIN64) || (defined(__x86_64__) &amp;&amp; !defined(__ILP32__) ) || defined(_M_X64) || defined(__ia64) || defined (_M_IA64) || defined(__aarch64__) || defined(__powerpc64__)
-        #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) typedef struct object##_T *object;
-#else
-        #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) typedef uint64_t object;
-#endif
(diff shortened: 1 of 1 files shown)


krOoze:
Hm, I think converting struct to an int is undefined behavior though.

awr1:
Casts to arithmetic types directly are forbidden (if one wanted to, a valid way of getting the integer value, apart from grabbing the `opaqueHandle` field, would be `uint64_t a = *(uint64_t*)&ndHandle`), however the point of ND handles is that they are opaque and thus shouldn't be played around with (you already would need to perform that kind of cast for the 64-bit platforms to acquire their underlying integer value).

This should be binary compatible; I think all calling conventions (regardless of platform) will handle it the same as it did prior.

krOoze:
I am pretty sure layers have to do it. As well as e.g. `vkSetDebugUtilsObjectNameEXT` requires even users to convert it to `uint64_t`. Your preassumption is incorrect.

oddhack:
Thanks for the suggestion. We tried something like this early in Vulkan development without success, and given seriously compatibility concerns a change like this raises with some of our members, we're not going to do it.

jekstrand:
> This should be binary compatible; I think all calling conventions (regardless of platform) will handle it the same as it did prior.

Yes, it would be binary compatible (probably).  However, it would break the compile for every Vulkan app/driver ever written.  That's not something we want to do.

PR #1215, opened 2020-03-29
Random editing and markup

krOoze:


appendices/VK_KHR_external_memory.txt
```diff
@@ -137,8 +137,8 @@ Options for defining this transition include:
 A new structure has the advantage that the type of external transition can
 be described in as much detail as necessary.
 However, there is not currently a known need for anything beyond
-differentiating external vs.
-internal accesses, so this is likely an over-engineered solution.
+differentiating external versus internal accesses, so this is likely an
```
oddhack: Better dealt with by adding vs. to endAbbrev in scripts/reflow.py.

appendices/spirvenv.txt
```diff
@@ -717,7 +717,7 @@ endif::VK_VERSION_1_2,VK_KHR_shader_float16_int8[]
   ** *Private*
   ** *Function*
 ifdef::VK_KHR_shader_clock[]
-  * Scope for code:OpReadClockKHR must be limited to:
+  * Scope for code:OpReadClockKHR must: be limited to:
```
oddhack: Any edits to normative terms, while almost certainly correct, will still require increased scrutiny by the WG before it will sign off on them. So I suggest splitting these changes out. Things that are actually just typos / markup errors the WG will usually just approve on my say-so so they're faster.

krOoze: That shouldn't be a problem.

I edited only the ones that I were 110 % sure about and only the tip of the iceberg. There seems to be a large number of colloquial use of the words rather than the RFC meaning. :/

appendices/VK_KHR_external_memory.txt
krOoze: Maybe.
Probably could also be reworded to english instead of using cutesy latin words.

oddhack: "versus" and its abbreviation are inherited from Middle English. We'll continue to use it.

appendices/spirvenv.txt
krOoze: @oddhack I think this PR should be safe now.

appendices/VK_KHR_external_memory.txt
krOoze: @oddhack Reworded one case, added exception to the script too.

Yes, latin is part of many modern languages. What I mean is that in 90 % of the cases it still looks like "look I know latin" rather than being the best way to phrase something.

scripts/reflow.py
```diff
@@ -113,10 +113,10 @@
 # A single letter followed by a period, typically a middle initial.
 endInitial = re.compile(r'^[A-Z]\.$')
 # An abbreviation, which doesn't (usually) end a line.
-endAbbrev = re.compile(r'(e\.g|i\.e|c\.f)\.$', re.IGNORECASE)
+endAbbrev = re.compile(r'(e\.g|i\.e|c\.f|vs)\.$', re.IGNORECASE)
```
krOoze: @oddhack BTW, first time I heard of "cf.". Doesn't see, to be used in the spec. Per internet, "c.f." seems to be incorrect spelling

chapters/VK_EXT_debug_report.txt
```diff
@@ -168,7 +168,7 @@ type of object handle being reported, are:
 include::{generated}/api/enums/VkDebugReportObjectTypeEXT.txt[]
 
 [[debug-report-object-types]]
-.VkDebugReportObjectTypeEXT and Vulkan Handle Relationship
+.etext:VkDebugReportObjectTypeEXT and Vulkan Handle Relationship
```
oddhack, line 171: This and a couple of similar changes make the 'allchecks' target fail, which breaks our internal CI. The problem is that etext: isn't just a "format this like elink, but don't link" macro, it's semantically actually different (probably should be called enum:). The checker script detects that etext: is being used with something which isn't an enumerant name. For the time being I'm reverting these changes despite the minor formatting improvements they would otherwise offer.

krOoze: @oddhack Hm, the styleguide says `etext` is not validated.

So, what do I need to do with this PR? Did you revert the whole thing, or just the bad occurences internally?

PR #1217, opened 2020-03-30
Random RFC terminology markup fixes

krOoze:


chapters/initialization.txt
```diff
@@ -49,8 +49,8 @@ The table below defines the various use cases for
 fname:vkGetInstanceProcAddr and expected return value ("`fp`" is "`function
 pointer`") for each case.
 
-The returned function pointer is of type tlink:PFN_vkVoidFunction, and must
-be cast to the type of the command being queried.
+The returned function pointer is of type tlink:PFN_vkVoidFunction, and must:
+be casted to the type of the command being queried before use.
```
jeffbolznv, line 53: I think "cast" is more natural than "casted"

chapters/devsandqueues.txt
```diff
@@ -1280,7 +1280,7 @@ include:
 include::{generated}/api/enums/VkMemoryOverallocationBehaviorAMD.txt[]
 
   * ename:VK_MEMORY_OVERALLOCATION_BEHAVIOR_DEFAULT_AMD lets the
-    implementation decide if overallocation should be allowed.
+    implementation decide if overallocation is to be allowed.
```
jeffbolznv, line 1283: maybe just "is"

chapters/fundamentals.txt
```diff
@@ -1260,14 +1260,14 @@ application and implementation.
 [NOTE]
 .Note
 ====
-If ename:VK_ERROR_UNKNOWN is received, the application should be checked
+If ename:VK_ERROR_UNKNOWN is received, the application should: be checked
```
jeffbolznv, line 1263: Strictly speaking, notes don't need normative language.

krOoze: Right, in that case Notes should avoid keywords all the same, unless they are referencing normatively established requirement. It is still confusing and I think some Notes still raise real requirements.

chapters/synchronization.txt
```diff
@@ -1231,7 +1231,7 @@ One example of this would be if the pending signal operation is, itself,
 waiting on the same semaphore at a lower value and the call to
 fname:vkSignalSemaphore signals that lower value.
 Furthermore, if there are two or more processes or threads signaling the
-same timeline semaphore from the host, the client must ensure that the
+same timeline semaphore from the host, the application must: ensure that the
```
TomOlson, line 1234: Policy is not to use normative terms in Notes, so this should just be "the application must ensure". There are cases where we've done it, but they are bugs and we should remove them.

chapters/fundamentals.txt
```diff
@@ -1256,19 +1256,22 @@ the application.
...
+Under these conditions, it may: be returned from any command returning a
+elink:VkResult.
 
 [NOTE]
 .Note
 ====
-If ename:VK_ERROR_UNKNOWN is received, the application should be checked
-against the latest validation layers to verify correct behavior as much as
-possible.
+If ename:VK_ERROR_UNKNOWN is received, the application would benefit from
```
TomOlson, line 1265: The phrase "the application would benefit from being checked..." is quite awkward. If you don't like "the application should be checked", perhaps something like "recommended practice is to check the application..."? 

To be honest, this Note was already a mess. An alternative to this fix would be to reorder the statements like this:
> ename:VK_ERROR_UNKNOWN is not expected to be returned from any command if application behavior is valid and the implementation is bug-free. If the error is received, recommended practice is to use the latest validation layers to check for invalid behavior on the part of the application. If no invalid usage is detected, report the error to the implementor as a possible implementation bug.

krOoze: I mostly want to deal with the RFC keywords. Did not really wanted to get suck rewording this Note.
If I just replace "should" with "recommended" I am back to square one. I guess I will just revert this for the time being.

oddhack:
@jeffbolznv I'm going to go ahead and merge based on Tom's feedback. If you need more tweakage let's handle internally.

PR #1228, opened 2020-04-06
Fix optional bitmasks in non-optional array

krOoze:


oddhack:
Usual request, can we see the resulting changes in the generated VUs?

krOoze:
@oddhack Right, it is supposed to be null diff with the `pBindingFlags` fix. Though something changed in newer commits. Hold on.

krOoze:
@oddhack Conflict with 139 resolved. Diff:

diff

oddhack:
@krooze is this MR still relevant? @jeffbolznv made a similar change in the 1.2.139 spec update and, AFAICT, the same validity comes out of your branch either at the 1.2.139 commit or the one following with your change.  Am I missing something?

krOoze:
@oddhack I think so. I still get a diff matching the one posted above.

PR #1232, opened 2020-04-07
Require impl to return sane vulkan version

### LINKED ISSUE #904: What's the deal with vkEnumerateInstanceVersion
Rant from #674:

1) `vkEnumerateInstanceVersion` returns only one code: `VK_SUCCESS`. That's weird; why? Is it planned to be extended with something here in 1.2? Should I now be checking its return value, or treat it like a `void` returning command?

2) Why does it not allow coexistence of `1.N` and potential `2.0`. It can only return one Vulkan version. Should it be restricted to only return a `1.N` value, and later a vkEnumerateInstanceVersion2 will be added for `2.N` versions etc? Or how will that work?

krOoze:
fix #904.2

### DIFF
--- a/appendices/boilerplate.txt
+++ b/appendices/boilerplate.txt
@@ -160,7 +160,7 @@ version of the header was used during their generation.
 Applications should not use this value as their
 slink:VkApplicationInfo::pname:apiVersion.
 Instead applications should explicitly select a specific fixed major/minor
-API version using, for example, one of the dname:VK_API_VERSION_*_* values.
+API version using, for example, one of the `VK_API_VERSION_*_*` values.
 
 include::{generated}/api/defines/VK_HEADER_VERSION_COMPLETE.txt[]
 
--- a/chapters/devsandqueues.txt
+++ b/chapters/devsandqueues.txt
@@ -102,10 +102,31 @@ include::{generated}/api/structs/VkPhysicalDeviceProperties.txt[]
     device.
     See <<sparsememory-physicalprops,Sparse Properties>> for details.
 
-ifdef::VK_VERSION_1_1[]
+An implementation must: return pname:apiVersion
+<<extendingvulkan-coreversions-versionnumbers,version>> equal to
+ifndef::VK_VERSION_1_1[1.0.]
+ifndef::VK_VERSION_1_2[1.1.]
+ifndef::VK_VERSION_1_3[1.2.]
+Patch version returned by the implementation is informational and may: be
+any value.
+An application must: expect pname:apiVersion to be any value greater or
+equal to 1.0.0 and less than 2.0.0.
+
 [NOTE]
 .Note
 ====
+The above restriction is different when addressing implementations than when
+addressing applications.
+An implementation written against this version of the specification reports
+exactly that version.
+An application written against this version of the specification has to
+accept device of any 1.x version.
+
+An implementation is allowed to return a non-zero patch version, so
+applications need to be careful when comparing the returned value to
+`VK_API_VERSION_1_*` macros, which **do** have `0` for a patch version.
+
+ifdef::VK_VERSION_1_1[]
 The value of pname:apiVersion may: be different than the version returned by
 flink:vkEnumerateInstanceVersion; either higher or lower.
 In such cases, the application must: not use functionality that exceeds the
@@ -115,8 +136,8 @@ is the version associated with a slink:VkInstance and its children, except
 for a slink:VkPhysicalDevice and its children.
 sname:VkPhysicalDeviceProperties::pname:apiVersion is the version associated
 with a slink:VkPhysicalDevice and its children.
-====
 endif::VK_VERSION_1_1[]
+====
 
 The pname:vendorID and pname:deviceID fields are provided to allow
 applications to adapt to device characteristics that are not adequately
--- a/chapters/initialization.txt
+++ b/chapters/initialization.txt
@@ -214,6 +214,34 @@ include::{generated}/api/protos/vkEnumerateInstanceVersion.txt[]
     of Vulkan supported by instance-level functionality, encoded as
     described in <<extendingvulkan-coreversions-versionnumbers>>.
 
+An implementation must: return pname:pApiVersion
+<<extendingvulkan-coreversions-versionnumbers,version>> that is equal to
+ifndef::VK_VERSION_1_1[1.0.]
+ifndef::VK_VERSION_1_2[1.1.]
+ifndef::VK_VERSION_1_3[1.2.]
+Patch version returned by the implementation is informational and may: be
+any value.
+An application must: expect pname:pApiVersion to be any value greater or
+equal to 1.1.0 and less than 2.0.0.
+
+[NOTE]
+.Note
+====
+The above restriction is different when addressing implementations than when
+addressing applications.
+An implementation written against this version of the specification reports
+exactly that version.
+An application written against this version of the specification has to
+accept instance of any 1.x version.
+1.0.x will never get returned here, because fname:vkEnumerateInstanceVersion
+simply would not exist, and flink:vkGetInstanceProcAddr returns `NULL` for
+it.
+
+An implementation is allowed to return a non-zero patch version, so
+applications need to be careful when comparing the returned value to
+`VK_API_VERSION_1_*` macros, which **do** have `0` for a patch version.
+====
+
 [NOTE]
 .Note
 ====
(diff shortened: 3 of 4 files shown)


appendices/boilerplate.txt
```diff
@@ -160,7 +160,7 @@ version of the header was used during their generation.
 Applications should not use this value as their
 slink:VkApplicationInfo::pname:apiVersion.
 Instead applications should explicitly select a specific fixed major/minor
-API version using, for example, one of the dname:VK_API_VERSION_*_* values.
```
TomOlson, line 163: What is the purpose of this change? `dname:` seems perfectly appropriate here, doesn't it?

chapters/devsandqueues.txt
```diff
@@ -102,6 +102,17 @@ include::{generated}/api/structs/VkPhysicalDeviceProperties.txt[]
     device.
     See <<sparsememory-physicalprops,Sparse Properties>> for details.
 
+An implementation must: return pname:apiVersion version that is greater or
+equal to 1.0.0, and less than 2.0.0.
```
TomOlson, line 106: Discussed in the WG today. We think the implementation should be required to return a valid version number, so we'd prefer to say (in the 1.2 spec) that the implementation must: return 1.0.x, 1.1.x, or 1.2.x. If we do that, the note below should also be extended to note that future versions of Vulkan may return higher major or minor release version numbers, and applications should be written to allow for that.

chapters/devsandqueues.txt
```diff
@@ -102,6 +102,17 @@ include::{generated}/api/structs/VkPhysicalDeviceProperties.txt[]
...
+
+[NOTE]
+.Note
+====
+An implementation is allowed to return a non-zero patch version, so
+applications need to be careful when comparing the retuned value to
```
TomOlson, line 112: typo: 'retuned' -> 'returned'

chapters/initialization.txt
```diff
@@ -214,6 +214,17 @@ include::{generated}/api/protos/vkEnumerateInstanceVersion.txt[]
     of Vulkan supported by instance-level functionality, encoded as
     described in <<extendingvulkan-coreversions-versionnumbers>>.
 
+An implementation must: return pname:pApiVersion version that is greater or
+equal to 1.1.0, and less than 2.0.0.
+
```
TomOlson, line 219: Again, we'd prefer to require the implementation to return values corresponding to published specs, but note that applications should be written to be forward compatible with future versions.

appendices/boilerplate.txt
krOoze: dname currently does not handle wildcards *

chapters/devsandqueues.txt
krOoze: Right. Strictly speaking a 1.2 spec should say the device must return exactly 1.2. Though I am not extreme to write it like that. But that also means the device is allowed to be any 1.x version (as long as instance is >1.0). E.g. in 1.1 instance the device can report 1.2. So it feels like 1.1 spec should still say it can be 1.2.

chapters/initialization.txt
krOoze: Again, I kinda disagree. An application written against Vulkan 1.1 must be able to tolerate if the function returns e.g. 1.2.

krOoze: Hm, for now I went with the extremist wording just to test how it feels...

Tobski: So I think I kind of agree with... both of you?

Implementations should be returning exactly the version they support - not "any arbitrary version". We should make this clear here - CTS is written to test functionality based on this returned value.

But to @krOoze's point, I agree applications should be able to handle any future value corresponding to a real spec, though there's nothing fundamentally stopping them going "whoa hey this version is more than I expected".

So I'd suggest changing it into two parts. Something like:

> The value returned in pname:pApiVersion must: correspond to the version of the Vulkan API specification supported by the implementation.
> Minor versions of the Vulkan 1.x specification are backwards compatible, and so applications should: be written to handle any version in the range [1.1.x,2.x.x)

TomOlson: I like where you're going with this, but there are some details to work out. Re:

> The value returned in pname:pApiVersion must: correspond...

It's a nit, but a layer (e.g. devsim) can report "1.1" when running on a 1.2 device. That seems like it ought to be allowed; it certainly doesn't make the implementation non-conformant.

Re:

> applications should: be written to handle any version in the range [1.1.x,2.x.x)

Devices are certainly free to restrict themselves to running on 1.2.x or greater. We might say 

>applications targeting 1.x.y should: be written to accept any API version number in the range [1.x.y,2.0.0)

krOoze: @TomOlson Counternit: Not sure layers are restricted to Vulkan spec. I assumed they are generally allowed to do whatever they want, and so have separate API contract with the application that uses them.

If the layer wants to pretend it is conformant 1.1 device running on 1.2. then it must behave as 1.1 device.

@Tobski 
> though there's nothing fundamentally stopping them going "whoa hey this version is more than I expected".

IDK, if an app wants to cripple itself from running on newer versions, who I am to tell it should not. Does not feel noteworthy. Maybe only concern is if someone accidendally writes `if(version != 1.2) abort();` instead of `if(version < 1.2)`. Not sure if this level of hand-holding is necessary. The link to meaning of a version is in the first sentence. And how semver works is generally known.

krOoze: PS: I did not really want to get into what version the application should "handle". I tried to just say the **value** the application gets might be in the whole 1.X range, and I don't care what the app does with the value as long as it acknowledges this fact. I.e. it might crash for 2.0, but it should work gracefully for 1.X (even if that means quiting).

Basically, conformant Vulkan app should **not** have a construct like

```
if (version == VK_VERSION_1_1) doSomething();
else if (version == VK_VERSION_1_0) doSomethingElse();
// But what about all the other versions? This code is probably not valid.
```

But on a second thought I should probably not be telling the app what to do at all. If it is programmed to crash, that is its own business, as long as it does not feed Vulkan API nonsense.

Tobski: @TomOlson 

> It's a nit, but a layer (e.g. devsim) can report "1.1" when running on a 1.2 device. That seems like it ought to be allowed; it certainly doesn't make the implementation non-conformant.

Sure, I meant more "you can't advertise 1.2 if you don't support 1.2 functionality". Layers aren't affected by this language when doing that sort of thing since they are supposed to promise a conformant implementation - it doesn't have to be the same conformant version. A layer advertising 1.1 in this situation just has to make sure that all 1.1 functionality is advertised, and behavior is as expected for a 1.1 implementation.

> applications targeting 1.x.y should: be written to accept any API version number in the range [1.x.y,2.0.0)

That would be nonsense when reading the 1.1 spec. But sure, I take the point - 1.1.x is an example relevant to the current spec only - I had assumed the final language would be more precise.

@krOoze 

> IDK, if an app wants to cripple itself from running on newer versions, who I am to tell it should not.

In this case, the "I" is the Vulkan specification 😄. Should represents a recommendation - applications can feel free to not follow it 🤷‍♀️

So....

...I think we're in agreement on this sentence?

> The value returned in pname:pApiVersion must: correspond to the version of the Vulkan API specification supported by the implementation.

And after thinking about it more, I think I agree the second sentence I suggested is probably redundant and not super useful - so yea maybe just that one?

TomOlson: @tobski, I like it, but under the Khronos definition of "support" (which exists outside the specification), an implementation can support multiple versions - so a conformant 1.1 driver supports both the Vulkan 1.1 and 1.0 specifications, and talking about "***the*** version of the Vulkan API specification supported..." is logically incorrect. An implementation that is conformant to the Vulkan 1.2 specification could return 1.2 *or* the version number of any backward-compatible successor version - notionally 1.3, 1.4, et al. 

We could say:
> `must:` correspond to the highest backward-compatible version of the Vulkan API specification supported by the implementation

That statement tells the app writer that they may get a version number higher than 1.2, and the "Extending Vulkan" chapter tells them what version numbers they can confidently expect to be compatible with. The statement doesn't actually constrain the implementer at all, because the only versions of the API the implementation *can* support in the Khronos sense are the one corresponding to pname:pApiVersion and its predecessors.

krOoze: It is odd that we are getting into a prolonged discussion over something that should be a dead simple versioning scheme. I should probably rethink it, so do not waste too much time reviewing the current version.

> An implementation that is conformant to the Vulkan 1.2 specification could return 1.2 or the version number of any backward-compatible successor version - notionally 1.3, 1.4, et al.

I don't follow. An implementation written against the 1.2 Vulkan specification cannot simply return 1.3+.

Tobski:
Hi @krOoze, as per #1249, I'd like to thank you for trying to get some sane requirements in the api for this - again I think we agree with what you're trying to do, but got lost in the details of what that meant.

As per the other issue, if you do have any specific feedback or thoughts you want to share on the matter, please feel free to do so here or on another issue, and we'll make sure to run things by you before we accept a solution.

PR #1235, opened 2020-04-08
Fix chunked target

krOoze:
fix custom patch inserting stuff into random locations

@oddhack fixes the stuff mentioned in #702


oddhack:
What's the nature of the problem this addresses? The chunked spec we generate doesn't have glaring problems AFAWK, so is this a performance issue with #702 not getting run correctly in the chunked spec, or something like that?

krOoze:
@oddhack The things in the `custom.patch` are injected at the wrong location. E.g. the "0. Preamble" is at the wrong location in the TOC.  The scripts\styles are also at some nonsense location instead of `<head>`.

oddhack:
I pulled this branch locally and brought it up to date with KhronosGroup/Vulkan-Docs master. Building the chunked spec works OK, but when I view it in Chromium, the searchbox says "Loading Search Data" forever. The 1.2.144 spec I just published doesn't do this - the "Loading" message shows for a fraction of a second then the searchbox clears and terms can be entered. Can you rebase this on master and verify that it's doing what you think it's doing?

krOoze:
#1285 will subsume this, so not sure if it is worthwile to continue this PR...
Maybe let's just update the `custom.patch` file until then?

krOoze:
FWIW this is updated.

PR #1238, opened 2020-04-09
Add lunr to package.json

krOoze:
allows to simply `npm install` all deps

### DIFF
--- a/config/chunkindex/README.adoc
+++ b/config/chunkindex/README.adoc
@@ -32,5 +32,5 @@ Portions of the chunked.js script are taken from
 These portions are copyright Scott Jehl, Filament Group, Inc., under an MIT
 license.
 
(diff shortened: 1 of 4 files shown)


oddhack:
TBH I'd like to remove this stuff. I don't know what its utility is and we very much want people to use our Docker image, or be completely responsible for their own build environment, at  this point.

krOoze:
I believe it is something for the searchbox.

oddhack:
The package* files, yes.

krOoze:
Right. BTW I don't see where you are installing the `he` and `escape-string-regexp` in the docker. `package.json` could be still helpful to keep the dependencies organized even for the docker.

oddhack:
> Right. BTW I don't see where you are installing the `he` and `escape-string-regexp` in the docker.

Thanks! I had forgotten to push the updated image / Dockerfile and was just running off the local cache without realizing it.

PR #1246, opened 2020-04-16
Move app-addressing note away from styleguide

krOoze:


chapters/fundamentals.txt
```diff
@@ -885,6 +885,18 @@ knowledge of the application, without allowing return of values not defined
...
 
+[NOTE]
+.Note
+====
+Application developers are encouraged to be careful when using `switch`
+statements with Vulkan API enums.
+This is because new extensions can add new values to existing enums.
+The use of a `default:` statement, within a `switch`, may avoid future
+compilation issues.
+Particularly for enums like e.g. elink:VkDriverId that may change at any
```
oddhack, line 896: Any of them can change any time an extension is added, so this sentence feels redundant.

krOoze: Right, I mean this one can change without enabling an extension.

PR #1249, opened 2020-04-17
Add apiVersion VU

krOoze:


critsec:
(On behalf of the WG) Making a request for something less than 1.0 undefined behavior is okay, but we don't want to do that for 2.0-or-later. The spec already says what happens when `apiVersion` is greater than the supported instance version; making requests for some higher versions result in undefined behavior doesn't seem to add value and limits our backwards-compatibility options if/when we do create version 2.0.

krOoze:
> The spec already says what happens when apiVersion is greater than the supported instance version;

Which is what?

Providing 2.x there is semantic nonsense. 1.7 Vulkan application can work on 1.4 instance, if it refrains from using stuff added in 1.5&ndash;1.7. Application for 2.0 cannot work no matter what it does; it probably linked the wrong loader in the first place and cannot be even sure if `vkCreateInstance` exists or how many parameters it has...

> our backwards-compatibility options if/when we do create version 2.0

Does not really, because 2.0 is, by definition, not backwards-compatible. In 2.0 spec you could say it accepts 2.x values. In Vulkan 1.x providing 2.x seems like undefined behavior.

Besides, your `vkEnumerateInstanceVersion` is not even future-proof. It is unable to return multiple major versions. And it is not clear what minor versions of 1.x are supported if 2.0 is reported.

critsec:
It would be possible for the hypothetical Vulkan 2.0 to retain some elements of 1.0 (like `vkEnumerateInstanceVersion`) and say things like that a Vulkan 2.0 implementation must also provide 1.0, while still making some incompatible changes elsewhere in the API. Possibly we could do that while still allowing an app to gracefully/incrementally fall back to 1.x functionality. I'm not saying we will do that or even that we're likely to -- we just don't know yet. But there doesn't seem to be any good reason to remove that option at this point.

Currently it is not a problem for an app to pass 2.x here, it should just be treated the same as if they passed 1.42 or other doesn't-yet-exist future version. It would be reasonable for validation layers to issue a warning, but there doesn't seem to be any benefit to either applications or implementations of making this undefined behavior.

krOoze:
> Currently it is not a problem for an app to pass 2.x here, it should just be treated the same as if they passed 1.42 or other doesn't-yet-exist future version.

It **is** undefined behavior. It is unknown whether 2.x will support 1.x, therefore it is invalid usage to then use any 1.x functionality, including `vkCreateInstance` itself. That is non-obvious, and the VU just spells it out without mind-twisting.

2.0 spec later can make it defined behavior. Hell, even a minor version can lift a VU restriction. You are saying lot of stuff that addresses spec-makers; it feels like it should be in the Conventions document. But VU addresses application-makers.

krOoze:
Since the discussion is here, moving the non-controversial part of change to #1252

critsec:
> It is undefined behavior. It is unknown whether 2.x will support 1.x, therefore it is invalid usage to then use any 1.x functionality, including calling vkCreateInstance itself.

While 2.x might have some backwards-incompatible aspects, it might not be completely backwards incompatible. We might want to make `vkCreateInstance` work in a backwards-compatible way. If we make this invalid usage now, that will never be possible unless we require the app to "lie" about its api version -- we'll close that door forever. On the other hand, making it invalid usage now doesn't seem to have any benefit that I can see. An app that knows what 2.x is is still constrained to only use 1.x functionality on a 1.x implementation. Starting in 1.1, implementations don't need to pay attention to the app's api version, but if they wanted to they could just clamp the app's value (whether 1.x or 2.x) to the highest version supported by the implementation.

> 2.0 spec later can make it defined behavior. Hell, even a minor version can lift a VU restriction.

They can only list VU restrictions when running against implementations that support the new version of the spec. A 2.0 spec can't lift a VU restriction when running against a 1.x implementation -- there's no way to retroactively make sure that 1.x implementations that already exist handle the previously-invalid input correctly. That's the crux of the issue: once we make this invalid usage in 1.x, we never have the option of making this part of the API backwards-compatible in 2.x.

krOoze:
You seem to have it other way around. You are always allowed to make invalid usage valid in the future. But you are not allowed to make valid usage invalid.

> Starting in 1.1, implementations don't need to pay attention to the app's api version, but if they wanted to they could just clamp the app's value (whether 1.x or 2.x) to the highest version supported by the implementation.

Makes assumptions that are not given in the specification.
The spec says the parameter is ignored, so I could just provide Vulkan 999.0.0 there and still be considered valid app and expect to work forever as a 1.1 app.
Again, you seem to have it the other way around. By **not** making it invalid usage in Vulkan 1 you are preventing yourself from making Vulkan 2.0 implementation compatible in the future.

critsec:
> You are always allowed to make invalid usage valid in the future.

But that only affects implementations of the new version where the usage is valid, and can only be used by applications when they know they're running against that new implementation. If the Vulkan 1.2 spec says that passing `VkApplicationInfo::apiVersion = 2.0` is invalid, then we can't ever make it valid to pass that to a Vulkan 1.2 implementation in the future. A new version of the API (m.n) can only say that it is now valid to pass `apiVersion = 2.0` to implementations that support m.n or newer.

> Makes assumptions that are not given in the specification.

The specification does not require implementations to behave differently based on the value here, except that Vulkan 1.0 implementations must return an error if given anything other than 1.0. So it's not an assumption: there is no behavior required from implementations by the specification based on this value. It's informational for the driver (e.g. they might include it in crash/error reports if they have such a thing), but is mostly there for validation layers and to enable implicit layers to turn themselves off in cases where they would cause compatibility problems.

I feel like we're talking past each other. Do we have different definitions of what compatibility means here?

krOoze:
> I feel like we're talking past each other.

Yes, I think so. You seem to keep talking about implementations like it matters here.

This cannot affect implementations much, because VU addresses only applications.

Requiring 1.x application to say it is 1.x application does not restrict you in any way to make the major version implementation as you see fit in the future.

In fact, not having this requirement would restrict your options. Because as it is now, your hypotetical 2.0 implementation (if you go the way of extensive compatibility) could encounter a 1.1 application pretending to be a 2.0 application using this parameter and you would have no good way to handle that situation, and it restricts your future options. Either you would have to have a compatibility fully equivalent minor version, or ensure such app can never encounter such implementation if correctly linked, or add the restriction anyway 10 years from now when it would be harder to pretend it is only bugfix and not compatibility breaking change.

With the VU, there is proper three-way version hand-shake between app, loader, and ICD. And so you have more options what your future implementation could be.

critsec:
> This cannot affect implementations much, because VU addresses only applications.

That's not true. When something is invalid for applications to do, it means that implementations are free to behave however they want when an application *does* do it. VUs are restrictions on applications that give specific freedoms to implementations -- VUs are a contract between applications and implementations that are important to both sides. Most VU statements exist specifically because implementations need the freedom granted by them (we don't just make things invalid for the fun of it -- in general, well-defined behavior is better for developers). That is the fundamental reason why we can't retroactively make previously-invalid usage valid: we have no way of suddenly demanding that existing implementations behave in a specific way when they see what was previously invalid input.

> could encounter a 1.1 application pretending to be a 2.0 application

If an application claims to be using 2.0 but does things that are not valid in 2.0, it is not the driver's job to handle that. The validation layers will tell the developer what they're doing wrong, to the extent they can. I agree that it would be stupid and invalid (in the generic sense, not in the specific sense used by the Vulkan spec) for an app to do that today, because 2.0 doesn't exist and there's no way to know whether something will be valid when running on a 2.0 implementation. However, if the "1.1 application" claims to support 2.0 but only does things that are valid in 1.1, then everything should (actually, *is currently required to*) work fine. And if Vulkan 2.0 ends up being sufficiently backwards-compatible that those things are valid in 2.0, then they would also work fine there.

The case I'm concerned about is, if/when we *do* have a 2.0 spec, if we define it such that parts of the 2.0 API (like `vkCreateInstance`) are backwards-compatible with 1.x, can a *valid* 2.0 application run against a 1.x implementation? That is already designed to work properly for a 1.z application on top of a 1.x implementation (z > x); the application can't use any functionality beyond 1.x, but it's valid to say that it supports up to 1.z in `VkApplicationInfo`. The 2.0-application case is the same, in the hypothetical future where 2.0 exists and has the necessary (partial) backwards-compatibility.

> there is proper three-way version hand-shake between app, loader, and ICD.

`VkApplicationInfo::apiVersion` is not meant to provide part of such a handshake. It's essentially a hint to the validation layers that allows them to warn the developer if the developer does something that is valid on the current implementation, but is not valid in the API version the developer is primarily targeting. For example if the app is written against 1.2 and does something that is invalid on 1.2, but happens to be running on a 1.3 implementation where that thing is valid, then validation layers could still warn the developer. That helps them find bugs locally before deploying to other users with different configurations, even if the developer is using a shiny brand new GPU and driver.

krOoze:
@critsec As with #1232, something smells here to me. We are getting into prolonged discussion over something that should be dead obvious. It indicates to me that in ways the versioning system is improperly described as a whole, and we both seem to be reverting to some colloquial understanding of it (that inevitably happens to differ in subtle ways).

Anyways, I need to rethink this.

Tobski:
Hi @krOoze firstly I'd like to thank you for trying to get some sane validation in for the api version which is currently lacking - I think we all agree with what you're trying to do in both this and #1232, but got lost in the details of what that meant. It seems we don't have as much agreement as we thought internally, and I apologise for putting you on the receiving end of this confusion. We're going to take some time to try to figure that out before we try to and try to figure this out more internally before we try to resolve this.

If you do have any specific feedback or thoughts you want to share on the matter, please feel free to do so here or on another issue and we'll take that into consideration, and respond once we've had chance to internalise it.

I'll tag you on the internal issue so that we run things by you before accepting a solution.

krOoze:
> I apologise for putting you on the receiving end of this confusion.

No need. If the PRs expose the confusion, then they served their purpose as well.

> If you do have any specific feedback or thoughts you want to share on the matter

Yea, I have some ideas, but they are little half baked themselves.

Basic sanity requirements I would like to have is that X.Y implementers should read (only) X.Y spec. And app written against X.Y, should also read (only) X.Y spec (unless the Physical Device forces it to read X.Z spec, where Z < Y).

I would like to see the versioning scheme and compatibility guarantees and exceptions to those separately and centrally described. In other places the spec should not worry and concern itself what other versions do, and describe only the version it is. E.g 1.1 spec should tell (the 1.1) implementations to report themselves as 1.1. The 1.1 spec should instruct 1.1 apps to report themselves as 1.X (we can guarantee the minor version here is ignored, resp. used only "informationally", by the Loader and ICDs. But I am not sure same guarantee can be given for 2.0. The 2.0 impl might need some way to disambiguate 1.x and 2.x apps.).

It does get tricky when 2.0 is considered. And even more hairy if we contemplate 1.x instance could have 2.y physical devices, or vice-versa. And I am afraid `vkEnumerateInstanceVersion` and the physical device equivalent are not sufficiently future-proof. It feels like they should be able to return multiple versions as to report every major version they support (and also which minor version within that major version is supported).

@critsec 
> can a valid 2.0 application run against a 1.x implementation?

That is an interesting question. I think that is a problem only if the Instance is 1.x and the physical device is 2.y. Not sure though if I like the case where instance version < physical device version. Maybe should be outright forbidden.

PR #1250, opened 2020-04-18
Generate Unix newlines across platforms

krOoze:


oddhack:
What's the benefit of this?

krOoze:
It makes whichever file this generates to have Unix newlines. This script is used across repos, and sometimes it is invoked in Windows, which results in generated files having CRLF. I encountered this in Vulkan-Tools repo. One could set up git to do it, but it feels better to files have the established format in the first place; that is always UTF-8 + Unix newlines.

PR #1257, opened 2020-04-22
Minor text edit for consistency

haxpor:
This will make it consistent with the paragraph preceding it.


oddhack:
Thanks, good typo catch!

Issue #1278, opened 2020-05-16
Fix clang-tidy warnings for [hicpp-signed-bitwise]


MuAlphaOmegaEpsilon:
Running clang-tidy on some basic Vulkan code emits warnings related to the usage of VK_MAKE_VERSION, VK_API_VERSION_1_X and VK_VERSION_MAJOR/MINOR, due to the usage of bitwise shift operations performed on signed integers:
```
warning: use of a signed integer operand with a binary bitwise operator [hicpp-signed-bitwise]
```

oddhack:
Closed by #1279, I believe.

PR #1279, opened 2020-05-16
Use unsigned integers to perform bitwise operations

MuAlphaOmegaEpsilon:
Address issue #1278.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -143,7 +143,7 @@ server.
         <type requires="ggp_c/vulkan_types.h" name="GgpFrameToken"/>
 
         <type category="define">#define <name>VK_MAKE_VERSION</name>(major, minor, patch) \
-    ((((uint32_t)(major)) &lt;&lt; 22) | (((uint32_t)(minor)) &lt;&lt; 12) | ((uint32_t)(patch))))</type>
+    ((((uint32_t)(major)) &lt;&lt; 22) | (((uint32_t)(minor)) &lt;&lt; 12) | ((uint32_t)(patch)))</type>
         <type category="define">#define <name>VK_VERSION_MAJOR</name>(version) ((uint32_t)(version) &gt;&gt; 22)</type>
         <type category="define">#define <name>VK_VERSION_MINOR</name>(version) (((uint32_t)(version) &gt;&gt; 12) &amp; 0x3ff)</type>
         <type category="define">#define <name>VK_VERSION_PATCH</name>(version) ((uint32_t)(version) &amp; 0xfff)</type>
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -143,7 +143,7 @@ server.
         <type requires="ggp_c/vulkan_types.h" name="GgpFrameToken"/>
 
         <type category="define">#define <name>VK_MAKE_VERSION</name>(major, minor, patch) \
-    (((major) &lt;&lt; 22) | ((minor) &lt;&lt; 12) | (patch))</type>
+    ((((uint32_t) major) &lt;&lt; 22) | (((uint32_t) minor) &lt;&lt; 12) | ((uint32_t) patch)))</type>
         <type category="define">#define <name>VK_VERSION_MAJOR</name>(version) ((uint32_t)(version) &gt;&gt; 22)</type>
         <type category="define">#define <name>VK_VERSION_MINOR</name>(version) (((uint32_t)(version) &gt;&gt; 12) &amp; 0x3ff)</type>
         <type category="define">#define <name>VK_VERSION_PATCH</name>(version) ((uint32_t)(version) &amp; 0xfff)</type>
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -143,7 +143,7 @@ server.
         <type requires="ggp_c/vulkan_types.h" name="GgpFrameToken"/>
 
         <type category="define">#define <name>VK_MAKE_VERSION</name>(major, minor, patch) \
-    ((((uint32_t) major) &lt;&lt; 22) | (((uint32_t) minor) &lt;&lt; 12) | ((uint32_t) patch)))</type>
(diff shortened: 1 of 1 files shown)


xml/vk.xml
```diff
@@ -143,23 +143,23 @@ server.
         <type requires="ggp_c/vulkan_types.h" name="GgpFrameToken"/>
 
         <type category="define">#define <name>VK_MAKE_VERSION</name>(major, minor, patch) \
-    (((major) &lt;&lt; 22) | ((minor) &lt;&lt; 12) | (patch))</type>
-        <type category="define">#define <name>VK_VERSION_MAJOR</name>(version) ((uint32_t)(version) &gt;&gt; 22)</type>
-        <type category="define">#define <name>VK_VERSION_MINOR</name>(version) (((uint32_t)(version) &gt;&gt; 12) &amp; 0x3ff)</type>
+    (((major) &lt;&lt; 22U) | ((minor) &lt;&lt; 12U) | (patch))</type>
```
krOoze, line 146: I don't see why the second operand should be unsigned. Positive literal number is perfectly normal and good code. This should probably not be needed.

xml/vk.xml
```diff
@@ -143,23 +143,23 @@ server.
...
+        <type category="define">#define <name>VK_VERSION_MAJOR</name>(version) ((uint32_t)(version) &gt;&gt; 22U)</type>
+        <type category="define">#define <name>VK_VERSION_MINOR</name>(version) (((uint32_t)(version) &gt;&gt; 12U) &amp; 0x3ff)</type>
         <type category="define">#define <name>VK_VERSION_PATCH</name>(version) ((uint32_t)(version) &amp; 0xfff)</type>
 
         <type category="define">// DEPRECATED: This define has been removed. Specific version defines (e.g. VK_API_VERSION_1_0), or the VK_MAKE_VERSION macro, should be used instead.
-//#define <name>VK_API_VERSION</name> <type>VK_MAKE_VERSION</type>(1, 0, 0) // Patch version should always be set to 0</type>
+//#define <name>VK_API_VERSION</name> <type>VK_MAKE_VERSION</type>(1U, 0U, 0U) // Patch version should always be set to 0</type>
         <type category="define">// Vulkan 1.0 version number
-#define <name>VK_API_VERSION_1_0</name> <type>VK_MAKE_VERSION</type>(1, 0, 0)// Patch version should always be set to 0</type>
+#define <name>VK_API_VERSION_1_0</name> <type>VK_MAKE_VERSION</type>(1U, 0U, 0U)// Patch version should always be set to 0</type>
```
krOoze, line 154: Maybe just cast those to `uint32_t`? If it is 64 bit then gets downcasted to 32, some other overzealous tool might complain.

xml/vk.xml
MuAlphaOmegaEpsilon: Nice spot, I missed there was an option to relax the warning constraint.
Thank you!

xml/vk.xml
MuAlphaOmegaEpsilon: Alright, updated it!

oddhack:
The input values used elsewhere in the headers are never going to be negative integers, nor large enough to overflow. But if you want to do this, I'd rather you put the typecasts into the VK_MAKE_VERSION macro instead of all the places that invoke that macro - less intrusive and does not require changing my release process for spec updates, which this would as it stands.

xml/vk.xml
fluppeteer: FWIW, I don't think that should be necessary. IIRC, common 64-bit systems (I can't vouch for a Cray 2...) all have 32-bit int and 64-bit long long, with Microsoft using 32-bit longs and most other platforms using 64-bit longs. Unless we're somehow trying to run on a 16-bit system, 1U, 0U and 0U should be 32-bit, and I'd find it easier to read. I won't object, but my mild preference would be to avoid unnecessary (uint32_t) casting. Unless, I suppose, we've done it everywhere else.

https://en.cppreference.com/w/c/language/integer_constant

krOoze: @fluppeteer Well Vulkan does not use the old `int` system at all. Except in the macros.

`uint32_t` could be `int` or `long` per your own direction of thinking too. Some other overzealous tool might complain sooner or later. Technically, literal for `uint32_t` is `UINT32_C(42)`.

From C++ standpoint, it is important the result to have well defined unambiguous type (for function overloading, and for templating).

krOoze:
Yea, what @oddhack said. Adding the casts to `VK_MAKE_VERSION` macro seems better, as it would allow the user code to use that macro too without getting warnings.

xml/vk.xml
fluppeteer: Hmm. Fair enough, I suppose there might be a tool which complains about casting from long to int even when it knows it's not actually narrowing. (I might call that a _buggy_ tool...) Likewise, I'd hope any C++ wrapper would be happy with the obvious type promotions in this case rather than having overloaded it based on nuance of int. I don't feel strongly about it, though; it makes it a bit less legible, but a lot of the API can be accused of the same crime.

MuAlphaOmegaEpsilon:
Here it is, pushed the one-liner commit!

xml/vk.xml
```diff
@@ -143,7 +143,7 @@ server.
         <type requires="ggp_c/vulkan_types.h" name="GgpFrameToken"/>
 
         <type category="define">#define <name>VK_MAKE_VERSION</name>(major, minor, patch) \
-    (((major) &lt;&lt; 22) | ((minor) &lt;&lt; 12) | (patch))</type>
+    ((((uint32_t) major) &lt;&lt; 22) | (((uint32_t) minor) &lt;&lt; 12) | ((uint32_t) patch)))</type>
```
krOoze, line 146: In a macro, generally you want to be quite liberal with parentheses; it is text, not variables. E.g. `VK_MAKE_VERSION(1, 0, 1 + 1)` would expand to `(uint32_t)1 + 1`, which is not what you intended. Unlikely somebody uses it that way, but still...

MuAlphaOmegaEpsilon: Done! I always get tricked by this scenarios with macros...

PR #1285, opened 2020-05-30
Make chunked html permalinkable

krOoze:
Make chunked html permalinkable. I.e. so it accepts any anchor in `vkspec.html#vkHash` as if it was single-page spec.

Demo: http://vulkan.hys.cz/chunked_smart/vkspec.html#VkInstanceCreateInfo

TODO:
- [x] Do not load page when already on the correct chapter
- [x] Make links in the spec point to `vkspec.html` instead of `chap*.html` (might as well rewrite the chunker, ugh...)
- [x] Check that script is not triggered by external links
- [x] Rename index.html to something less indexy
- [x] Deal with the searchbox
- [x] Handle JavaScript\HTTP errors
- [x] Scroll to appropriate part in the TOC too
  - [x] Broken on mobile where the toc is on top; fix
- [ ] Make refpages link to this version (@oddhack wants to take care of it)
- [x] Gracefully handle disabled JavaScript
- [x] Reintroduce injecting Preamble into the TOC, or make it part of every chapter.
- [x] Properly integrate the "loadable" (#702) script
- [x] Searchbox feels to be loaded prematurely on first page load
- [x] Use the `?` for versioning browser cache
- [x] Make loading message\page less obtrusive (no blink)
- [x] <s>Deal with offline viewing (`file://`)</s>, or fail gracefully with specific error message
  - [ ] Make a fallback for offline viewing
- [ ] Restore link handles to code listings (perhaps best done outside this PR to include single-page)
- Soliciting feedback
- ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,11 +83,12 @@ SCRIPTS  = $(CURDIR)/scripts
 # HTMLDIR - 'html' target
 # PDFDIR - 'pdf' target
 # CHECKDIR - 'allchecks' target
-OUTDIR	  = $(GENERATED)/out
-HTMLDIR   = $(OUTDIR)/html
-VUDIR	  = $(OUTDIR)/validation
-PDFDIR	  = $(OUTDIR)/pdf
-CHECKDIR  = $(OUTDIR)/checks
+OUTDIR	   = $(GENERATED)/out
+HTMLDIR    = $(OUTDIR)/html
+CHUNKEDDIR = $(OUTDIR)/chunked
+VUDIR	   = $(OUTDIR)/validation
+PDFDIR	   = $(OUTDIR)/pdf
+CHECKDIR   = $(OUTDIR)/checks
 
 # PDF Equations are written to SVGs, this dictates the location to store those files (temporary)
 PDFMATHDIR:=$(OUTDIR)/equations_temp
@@ -234,25 +235,18 @@ $(OUTDIR)/$(KATEXDIR): $(KATEXSRCDIR)
 # Spec targets
 # There is some complexity to try and avoid short virtual targets like 'html'
 # causing specs to *always* be regenerated.
-ROSWELL = ros
-ROSWELLOPTS ?= dynamic-space-size=4000
-CHUNKER = $(HOME)/common-lisp/asciidoctor-chunker/roswell/asciidoctor-chunker.ros
+CHUNKER = $(PYTHON) $(SCRIPTS)/motherchunker.py
 CHUNKINDEX = $(CURDIR)/config/chunkindex
-# Only the $(ROSWELL) step is required unless the search index is to be
-# generated and incorporated into the chunked spec.
-#
 # Dropped $(QUIET) for now
 # Should set NODE_PATH=/usr/local/lib/node_modules or wherever, outside Makefile
 # Copying chunked.js into target avoids a warning from the chunker
 chunked: $(HTMLDIR)/vkspec.html $(SPECSRC) $(COMMONDOCS)
-	$(QUIET)$(PATCH) $(HTMLDIR)/vkspec.html -o $(HTMLDIR)/prechunked.html $(CHUNKINDEX)/custom.patch
+	$(QUIET)$(MKDIR) $(CHUNKEDDIR)
+	$(QUIET)$(CHUNKER) $< $(CHUNKEDDIR)/vkspec.html
 	$(QUIET)$(CP) $(CHUNKINDEX)/chunked.css $(CHUNKINDEX)/chunked.js \
-	    $(CHUNKINDEX)/lunr.js $(HTMLDIR)
-	$(QUIET)$(ROSWELL) $(ROSWELLOPTS) $(CHUNKER) \
-	    $(HTMLDIR)/prechunked.html -o $(HTMLDIR)
-	$(QUIET)$(RM) $(HTMLDIR)/prechunked.html
-	$(QUIET)$(RUBY) $(CHUNKINDEX)/generate-index.rb $(HTMLDIR)/chap*html | \
-	    $(NODEJS) $(CHUNKINDEX)/build-index.js > $(HTMLDIR)/search.index.js
+	    $(CHUNKINDEX)/lunr.js $(CHUNKEDDIR)
+	$(QUIET)$(RUBY) $(CHUNKINDEX)/generate-index.rb $(CHUNKEDDIR)/chap*html | \
+	    $(NODEJS) $(CHUNKINDEX)/build-index.js > $(CHUNKEDDIR)/search.index.js
 
 html: $(HTMLDIR)/vkspec.html $(SPECSRC) $(COMMONDOCS)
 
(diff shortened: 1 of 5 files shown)


oddhack:
To be sure I understand what's going on: you're intercepting load events on the client, where the anchor is still available, and rewriting to the appropriate chunked chapter there?

Does the map dictionary only get injected into the base vkspec.html document? I'm guessing that's why you have "Make links in the spec point to vkspec.html" above.

krOoze:
@oddhack Yea, the `vkspec.html` acts as a template to which the other chapters are injected via JavaScript based on the `#vkWhatever` part of the URL.

Currently it uses fragments from the chunker, but I don't think I need the chunker, really. The downside currently is the chunker produces links to `chap*.html` when appropriate, which escapes the smart version when followed. Same problem with the searchbox results. The TOC is part of the template (generated from the original spec), which means it has the correct links already. And so does anything that just linked within the same chapter.

> Does the map dictionary only get injected into the base vkspec.html document?

Yes. Also its `<head>` (scripts and styles) and header and footer is used, and those same parts in `chap*.html` files are therefore ignored.

oddhack:
> The downside currently is the chunker produces links to `chap*.html` when appropriate, which escapes the smart version when followed.

Those aren't hard to adjust by more postprocessing of the chunks. It sounds like a sed script could do for the chap*.html links.

I think of the chunker as a giant black box that isn't going to get updated by the author, and hope to find a better solution in time (preferably one integrated into asciidoctor, given the painful runtime of the chunker).

oddhack:
I'm not sure what you mean by "get rid of the chunker"? Write a replacement chunker?! 

FWIW, the Vulkan WG is pretty enthusiastic about this change once the various issues are worked out. It is a more elegant solution to the chunked-anchor problem than anything we'd thought of, mostly because we were thinking entirely of server-side solutions.

krOoze:
> I'm not sure what you mean by "get rid of the chunker"? Write a replacement chunker?!

Yes. Already did that in this PR.
Resp. I only need the chapter fragments from the original html, which is relatively trivial to do in the Python script.

> Once the various issues are worked out.

Right, what are the clear show-stoppers? I think searchbox needs to be reintegrated, rest is mostly polish.

oddhack:
> > I'm not sure what you mean by "get rid of the chunker"? Write a replacement chunker?!
> 
> Yes. Already did that in this PR.
> Resp. I only need the chapter fragments from the original html, which is relatively trivial to do in the Python script.

What kind of performance do you get relative to asciidoctor-chunker? The several minute runtime has been a sore point on that tool for as long as we've been using it, and being effectively unsupported and in such an obscure language that none of us would be able to fix problems is also an issue.

> Right, what are the clear show-stoppers? I think searchbox needs to be reintegrated, rest is mostly polish.

Most of your checklist look like things you already have planned. The refpage integration is something I'd prefer to take care of once this is ready to merge. Beyond that, our biggest concern is just getting the resulting document enough exposure to different browsers and OSes that we're confident it won't cause problems for the userbase. We try to do that by farming it out to people in Khronos using different environments and it usually takes a while as it's a low priority task for most of them, but we do get reports back in 3-4 weeks.

krOoze:
> What kind of performance do you get relative to asciidoctor-chunker?

40 s or so for me, incl. the single-page build as dependency.

krOoze:
@oddhack The "preamble" seems to contain the copyright. Do you want it in every chapter, or treat it like a separate chapter as it is now?

oddhack:
I think just having the copyright in the preamble should be OK, since that's how it is now AFAIK. The  reason for the proprietary spec copyright statement on the document we publish is that it is the  Khronos-ratified version of the document.

chunked/vkspec.html seems fine.

krOoze:
I am done I think.
By some dark magic even the searchbox seems to work with minimal changes...

oddhack:
I'll ask folks to start trying it out in different browser environments on today's meeting. Hopefully we can get it approved to post soon.

One thing about the refpages that will be odd is that when a refpage includes an xref to something that's not an API entity but, say, a section or table anchor, it links into the single-page document. While it's straightforward for me to modify the 'For more information' links to go to the chunked spec instead of the single-page spec, the non-API xrefs will still go to the single-page document. Short of adding non-API anchors to the rewrite table in your scripts, that seems unavoidable.

krOoze:
> Short of adding non-API anchors to the rewrite table in your scripts, that seems unavoidable.

It does indeed contain all anchors, not just API. It is preprocessed on the HTML level. Everything that has html `id` attribute ends up in the table.

oddhack:
Something several people noticed  in reviewing this is that there seemed to be a significantly longer delay, compared to the current chunked spec, when following a link into the spec. This results in a visible flash / stutter as it lands on vkspec.html and then redirects the anchor. This was somewhat objectionable. However, that's comparing the current version hosted on khronos.org vs. your sandbox copy on vulkan.hys.cz, so possibly relative speed of the servers is involved. Any thoughts on this?

This was mostly visible when following internal links inside the chunked document. If the stuttering is unavoidable in terms of the redirection / JS lookup, it could be alleviated by reverting to the behavior of the old chunker, where internal links go directly to the target chunk instead of indirecting through vkspec.html. That wouldn't improve things when coming from external links like the reference pages, but that's OK as they benefit much more from no longer loading the single-page document.

krOoze:
Dunno, seems about the same to me.
There is a flash of empty page saying that the new page is loading. I could remove that in the case the load is quick, or perhaps show a floating popup message instead.

It is just some freehosting. Can't hurt to try on khronos.org. I am gonna just rebase it:

oddhack:
I wanted to test performance on the same server as the published chunked spec so tried pulling this branch and building a local copy via 'makeSpec -spec all chunked ; chromium file://$cwd/gen/out/chunked/vkspec.html#extension-appendices-list', but I see something like this consistently no matter how I try to navigate around:

chunked

krOoze:
@oddhack It does not really works with `file://` (CORS seems to block the requests even in same folder). Should work on the server though.

krOoze:
So I added the less obtrusive loading message, so the loding shold feel better and perhaps (at least psychologically) bit shorter.

krOoze:
Ok now I added the less obtrusive loading message.
The loding should feel better and perhaps (at least psychologically) bit shorter.

oddhack:
> @oddhack It does not really work locally with `file://` (CORS seems to block the requests even in the same folder). Should work on the server though.

That's a big problem. A fair number of people want to build specs for offline consumption, and I personally would have no way of testing spec builds short of uploading it to our webserver. I poked at CORS a bit and there might be some way of configuring browsers to bypass it, so if that can be done at least for Chromium and Firefox that might do.

krOoze:
Standard practice is to test on localhost server like XAMPP.

Yes, it can be disabled in browsers, but that is development feature not for regular use. The security issue being that if you download malicious html on your desktop, it could upload all your files there to some server. Of course, no one seemed to care up to last year or something like that.

Anyway:
 - Firefox: `privacy.file_unique_origin` to `false` in `about:config`
 - Chrome (chromium?): `start chrome --allow-file-access-from-files` (also needs above commit)

That would work for your testing, but not healthy to ask this of the offline users.
I guess I could make the old school standalone chunks too if offline viewing is a thing.

krOoze:
Is there any new feedback to this?

PR #1289, opened 2020-06-08
Add noscript fallback for html

krOoze:
Add noscript fallback for html, so that the optional script added in #702 is simply skipped.
demo: http://vulkan.hys.cz/noscript/vkspec.html (works with either JavaScript on or off)

### DIFF
--- a/config/loadable_html/extension.rb
+++ b/config/loadable_html/extension.rb
@@ -12,9 +12,7 @@ class MakeHtmlLoadable < Extensions::Postprocessor
 
     if document.attr? 'stem'
 
-      loading_msg = '
-<div id="loading_msg"><p>Loading... please wait.</p></div>
-'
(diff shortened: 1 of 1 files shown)


krOoze:
@oddhack I was working on this yesterday, but it seems to cause serious issues with Firefox. You might want to revert this one real quick.

oddhack:
Erggh. We had tried Firefox but nobody noticed problems.

I can't approve a new PR against the scripts by myself and am planning to do a spec update tomorrow evening. I'm going to push an internal reversion of this change and we can revisit #1300 after the spec update goes out.

krOoze:
@oddhack Hm, I retried master now, and seems to work. Sorry to panic you. If others have tested it, then it is probably ok and I just had a corrupted file.

krOoze:
PS: Last Windows update broke my wi-fi so it drops like every half an hour. The problem seems to happen when the browser tries to read from its cache after it is corrupted by disconnect. Probably would happen on the old version too.

PR #1292, opened 2020-06-13
Add VK_EXT_directfb_surface

caramelli:
DirectFB interfaces was largely widespread in the GNU/Linux embedded world and the VK_EXT_directfb_surface extension allows rendering via DirectFB on such platform.

A screenshot with simple Vulkan examples running with the VK_EXT_directfb_surface extension is available on the HiGFXback project:

https://github.com/caramelli/higfxback/wiki/DirectFB#vulkan-rendering

And the yagears project provides sample code to use this VK_EXT_directfb_surface extension:

https://github.com/caramelli/yagears

Nicolas Caramelli

### DIFF
--- a/include/vulkan/vulkan.h
+++ b/include/vulkan/vulkan.h
@@ -71,6 +71,12 @@
 #endif
 
 
+#ifdef VK_USE_PLATFORM_DIRECTFB_EXT
+#include <directfb.h>
+#include "vulkan_directfb.h"
+#endif
+
+
 #ifdef VK_USE_PLATFORM_XLIB_XRANDR_EXT
 #include <X11/Xlib.h>
 #include <X11/extensions/Xrandr.h>
--- a/scripts/genvk.py
+++ b/scripts/genvk.py
@@ -293,6 +293,7 @@ def makeGenOpts(args):
                                                                      ] ],
         [ 'vulkan_xcb.h',         [ 'VK_KHR_xcb_surface'          ], commonSuppressExtensions ],
         [ 'vulkan_xlib.h',        [ 'VK_KHR_xlib_surface'         ], commonSuppressExtensions ],
+        [ 'vulkan_directfb.h',    [ 'VK_EXT_directfb_surface'     ], commonSuppressExtensions ],
         [ 'vulkan_xlib_xrandr.h', [ 'VK_EXT_acquire_xlib_display' ], commonSuppressExtensions ],
         [ 'vulkan_metal.h',       [ 'VK_EXT_metal_surface'        ], commonSuppressExtensions ],
         [ 'vulkan_beta.h',        betaRequireExtensions,             betaSuppressExtensions ],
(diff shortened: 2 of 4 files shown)


oddhack:
For future reference, the extension numbers in vk.xml are a single-threaded resource, so it's essential to reserve extension numbers in master branch before relying on them. There is a pending internal MR to reserve the same extension number you're using and while I can probably accommodate your PR, please make sure to reserve the number first with any future extensions.

As far as the actual functionality, could you clarify what Vulkan implementation this is for? There may be some additional questions raised by the working group once they've all had a chance to take a look at this.

caramelli:
Graphics rendering on GNU/Linux systems can rely on free software like
X11 or Wayland, each with their pros and cons, and DirectFB is also an
option with its advantages (that mainly targets embedded systems).
We like DirectFB or not, but having the VK_EXT_directfb_surface
extension offers the possibility of running Vulkan with a software
architecture based on DirectFB.
This VK_EXT_directfb_surface extension has been tested with
SwiftShader, so it can be used at least with a CPU-based
implementation of Vulkan.

I'm not familiar with the process used to register extensions. So I
understand that I need to create a new request to reserve a number,
with something like this in vk.xml:

        <extension name="VK_EXT_extension_347" number="347"
author="EXT" contact="Nicolas Caramelli @caramelli"
supported="disabled">
            <require>
                <enum value="0"
     name="VK_EXT_EXTENSION_347_SPEC_VERSION"/>
                <enum value="&quot;VK_EXT_extension_347&quot;"
     name="VK_EXT_EXTENSION_347_EXTENSION_NAME"/>
            </require>
        </extension>

I define the number 347 because the last number used in the master
branch is 346, is it correct ?

Nicolas Caramelli

oddhack:
We discussed yesterday. The actual XML / script changes look OK, but what's missing is the corresponding specification language describing how the new APIs function, and an extension appendix summarizing the functionality. I recommend templating off an existing WSI surface extension since the functionality is very similar. You're probably not familiar with spec authoring, so you may find it useful to do something like

> make clean; find . -name '*txt' | xargs egrep VK_KHR_win32_surface

to find files you would need to add to.

I understand that writing the spec language may provide burdensome for you to do in the near term but we really want an extension to be fully documented before incorporating it in the specs we publish. If you aren't able to write the spec language right now, just reserving the extension number in a separate PR would be OK - we could accept that and leave this PR pending until it includes spec language, too.

Regarding the extension reservation, yes, what you wrote above is exactly what you would do. Our public-facing documentation needs to make a note of this - I think this may actually be the first time we've seen a completely unanticipated extension show up on github, so we're finding some process issues.

1ace:
@caramelli before you start writing the whole thing, have you looked at [`VK_KHR_display`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_display.html) (& [`VK_KHR_display_swapchain`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_display_swapchain.html)) to see if it fits your needs?

caramelli:
Yes, but the idea is to render with DirectFB in the same way as Wayland, Xlib, XCB or other WSI.
In addition, for composition, we work directly with the DirectFB surface object used in the parameters of the vkCreateDirectFBSurfaceEXT() function.

Nicolas Caramelli

PR #1295, opened 2020-06-18
Remove allocation VU for a well defined case

krOoze:
Remove the `maxMemoryAllocationCount` and `maxSamplerAllocationCount` VUs. The cases are well defined:

> If a call to `vkAllocateMemory` would cause the total number of allocations to exceed these limits, such a call will fail and **must** return `VK_ERROR_TOO_MANY_OBJECTS`.

> If `maxSamplerAllocationCount` is exceeded, `vkCreateSampler` will return `VK_ERROR_TOO_MANY_OBJECTS`.


oddhack:
This should be fixed in the 1.2.150 spec update.

PR #1302, opened 2020-06-29
Fix merge for VK_EXT_directfb_surface extension

caramelli:
VK_EXT_directfb_surface is duplicated in vk.xml

Nicolas Caramelli


oddhack:
I have already fixed this internally and will propagate that out momentarily. The problem was that there were two PRs in flight touching the XML for this extension which weren't aware of each other - what normally happens is the reservation is accepted, then the extension spec branch is based on that.

PR #1310, opened 2020-07-04
Fix VK_KHR_shader_draw_parameters missing as a feature alias

krOoze:
plus cleanup of the 1.1 diff chapter.

### DIFF
--- a/appendices/versions.txt
+++ b/appendices/versions.txt
@@ -188,37 +188,72 @@ number of key extensions into the core API:
 
 include::{generated}/meta/promoted_extensions_VK_VERSION_1_1.txt[]
 
-The only changes to the functionality added by these extensions were to
-`VK_KHR_shader_draw_parameters`, which had the
-<<features-shaderDrawParameters, pname:shaderDrawParameters>> feature bit
-added to determine support in the core API, and
-<<features-variablePointersStorageBuffer,
-pname:variablePointersStorageBuffer>> from `VK_KHR_variable_pointers` was
-made optional.
-
-[[versions-1.1-new-features]]
-Additionally, Vulkan 1.1 added support for <<shaders-group-operations, group
-operations>> with <<shaders-scope-subgroup, subgroup scope>>,
-<<memory-protected-memory, protected memory>>, and a new command to
-<<vkEnumerateInstanceVersion, enumerate the instance version>>.
-
-=== Differences relative to `VK_KHR_sampler_ycbcr_conversion`
-
-If the `<<VK_KHR_sampler_ycbcr_conversion>>` extension is not supported,
-support for the <<features-samplerYcbcrConversion,samplerYcbcrConversion>>
-feature is optional.
-Support for this feature is defined by
-slink:VkPhysicalDeviceVulkan11Features::pname:samplerYcbcrConversion when
-queried via flink:vkGetPhysicalDeviceFeatures2.
+All differences in behavior between these extensions and the corresponding
+Vulkan 1.1 functionality are summarized below.
 
 === Differences relative to `VK_KHR_16bit_storage`
 
 If the `<<VK_KHR_16bit_storage>>` extension is not supported, support for
-the <<features-storageBuffer16BitAccess,storageBuffer16BitAccess>> feature
-is optional.
+the <<features-storageBuffer16BitAccess, pname:storageBuffer16BitAccess>>
+feature is optional.
 Support for this feature is defined by
-slink:VkPhysicalDeviceVulkan11Features::pname:storageBuffer16BitAccess when
-queried via flink:vkGetPhysicalDeviceFeatures2.
+slink:VkPhysicalDevice16BitStorageFeatures::pname:storageBuffer16BitAccess
+ifdef::VK_VERSION_1_2[]
+or slink:VkPhysicalDeviceVulkan11Features::pname:storageBuffer16BitAccess
+endif::VK_VERSION_1_2[]
+when queried via flink:vkGetPhysicalDeviceFeatures2.
+
+=== Differences relative to `VK_KHR_sampler_ycbcr_conversion`
+
+If the `<<VK_KHR_sampler_ycbcr_conversion>>` extension is not supported,
+support for the <<features-samplerYcbcrConversion, pname:samplerYcbcrConversion>>
+feature is optional.
+Support for this feature is defined by
+slink:VkPhysicalDeviceSamplerYcbcrConversionFeatures::pname:samplerYcbcrConversion
+ifdef::VK_VERSION_1_2[]
+or slink:VkPhysicalDeviceVulkan11Features::pname:samplerYcbcrConversion
+endif::VK_VERSION_1_2[]
+when queried via flink:vkGetPhysicalDeviceFeatures2.
+
+=== Differences relative to `VK_KHR_shader_draw_parameters`
+
+If the `<<VK_KHR_shader_draw_parameters>>` extension is not supported,
+support for the
+{spirv}/KHR/SPV_KHR_shader_draw_parameters.html[`SPV_KHR_shader_draw_parameters`]
+SPIR-V extension is optional.
+Support for this feature is defined by
+slink:VkPhysicalDeviceShaderDrawParametersFeatures::pname:shaderDrawParameters
+ifdef::VK_VERSION_1_2[]
+or slink:VkPhysicalDeviceVulkan11Features::pname:shaderDrawParameters
+endif::VK_VERSION_1_2[]
+when queried via flink:vkGetPhysicalDeviceFeatures2.
+
+=== Differences relative to `VK_KHR_variable_pointers`
+
+If the `<<VK_KHR_variable_pointers>>` extension is not supported,
+support for the <<features-variablePointersStorageBuffer,
+pname:variablePointersStorageBuffer>> feature is optional.
+Support for this feature is defined by
+slink:VkPhysicalDeviceVariablePointersFeatures::pname:variablePointersStorageBuffer
+ifdef::VK_VERSION_1_2[]
+or slink:VkPhysicalDeviceVulkan11Features::pname:variablePointersStorageBuffer
+endif::VK_VERSION_1_2[]
+when queried via flink:vkGetPhysicalDeviceFeatures2.
+
+=== Additional Vulkan 1.1 Feature Support
+
+[[versions-1.1-new-features]]
+In addition to the promoted extensions described above, Vulkan 1.1 added
+support for:
+
+  * The <<shaders-group-operations, group operations>> and
+    <<shaders-scope-subgroup, subgroup scope>>.
+  * The <<memory-protected-memory, protected memory>> feature.
+  * A new command to enumerate the instance version:
+    flink:vkEnumerateInstanceVersion.
+  * The slink:VkPhysicalDeviceShaderDrawParametersFeatures feature query
+    struct (where the `<<VK_KHR_shader_draw_parameters>>` extension did not
+    have one).
 
 include::{generated}/interfaces/VK_VERSION_1_1.txt[]
 
--- a/chapters/features.txt
+++ b/chapters/features.txt
@@ -50,20 +50,23 @@ must: be supported.
 |====
 | Extension                                 | Feature(s)
 ifdef::VK_VERSION_1_2[]
+ifdef::VK_KHR_shader_draw_parameters[]
+| `<<VK_KHR_shader_draw_parameters>>`       | <<features-shaderDrawParameters,pname:shaderDrawParameters>>
+endif::VK_KHR_shader_draw_parameters[]
 ifdef::VK_KHR_draw_indirect_count[]
-| `<<VK_KHR_draw_indirect_count>>`          | <<features-drawIndirectCount,drawIndirectCount>>
+| `<<VK_KHR_draw_indirect_count>>`          | <<features-drawIndirectCount,pname:drawIndirectCount>>
 endif::VK_KHR_draw_indirect_count[]
 ifdef::VK_KHR_sampler_mirror_clamp_to_edge[]
-| `<<VK_KHR_sampler_mirror_clamp_to_edge>>` | <<features-samplerMirrorClampToEdge,samplerMirrorClampToEdge>>
+| `<<VK_KHR_sampler_mirror_clamp_to_edge>>` | <<features-samplerMirrorClampToEdge,pname:samplerMirrorClampToEdge>>
 endif::VK_KHR_sampler_mirror_clamp_to_edge[]
 ifdef::VK_EXT_descriptor_indexing[]
-| `<<VK_EXT_descriptor_indexing>>`          | <<features-descriptorIndexing,descriptorIndexing>>
+| `<<VK_EXT_descriptor_indexing>>`          | <<features-descriptorIndexing,pname:descriptorIndexing>>
 endif::VK_EXT_descriptor_indexing[]
 ifdef::VK_EXT_sampler_filter_minmax[]
-| `<<VK_EXT_sampler_filter_minmax>>`        | <<features-samplerFilterMinmax,samplerFilterMinmax>>
+| `<<VK_EXT_sampler_filter_minmax>>`        | <<features-samplerFilterMinmax,pname:samplerFilterMinmax>>
 endif::VK_EXT_sampler_filter_minmax[]
 ifdef::VK_EXT_shader_viewport_index_layer[]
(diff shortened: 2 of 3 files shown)


appendices/versions.txt
```diff
@@ -188,37 +188,70 @@ number of key extensions into the core API:
...
 === Differences relative to `VK_KHR_sampler_ycbcr_conversion`
 
 If the `<<VK_KHR_sampler_ycbcr_conversion>>` extension is not supported,
 support for the <<features-samplerYcbcrConversion,samplerYcbcrConversion>>
 feature is optional.
 Support for this feature is defined by
-slink:VkPhysicalDeviceVulkan11Features::pname:samplerYcbcrConversion when
-queried via flink:vkGetPhysicalDeviceFeatures2.
+slink:VkPhysicalDeviceSamplerYcbcrConversionFeatures::pname:samplerYcbcrConversion
+ifdef::VK_VERSION_1_2[]
```
sfricke-samsung, line 213: This ifdef added is wrong as `or slink:VkPhysicalDeviceVulkan11Features::pname:samplerYcbcrConversion` is known with a 1.1 version of the spec

Also the reason for not having the original feature is this section is describing the "Differences relative to VK_KHR_sampler_ycbcr_conversion" so I don't see the reason to add both features

appendices/versions.txt
```diff
@@ -188,37 +188,70 @@ number of key extensions into the core API:
...
-<<memory-protected-memory, protected memory>>, and a new command to
-<<vkEnumerateInstanceVersion, enumerate the instance version>>.
+=== Differences relative to `VK_KHR_16bit_storage`
+
+If the `<<VK_KHR_16bit_storage>>` extension is not supported, support for
+the <<features-storageBuffer16BitAccess,storageBuffer16BitAccess>> feature
+is optional.
+Support for this feature is defined by
+slink:VkPhysicalDevice16BitStorageFeatures::pname:storageBuffer16BitAccess
+ifdef::VK_VERSION_1_2[]
```
sfricke-samsung, line 201: same, not needed to access `VkPhysicalDeviceVulkan11Features`

appendices/versions.txt
```diff
@@ -188,37 +188,70 @@ number of key extensions into the core API:
...
+[[versions-1.1-new-features]]
+In addition to the promoted extensions described above, Vulkan 1.1 added
+support for:
+
+  * The <<shaders-group-operations, group operations>> and
+    <<shaders-scope-subgroup, subgroup scope>>.
+  * The <<memory-protected-memory, protected memory>> feature.
+  * A new command to enumerate the instance version:
+    flink:vkEnumerateInstanceVersion.
+  * The <<features-shaderDrawParameters,shaderDrawParameters>> feature query
```
sfricke-samsung, line 253: either we should remove this line or add all the new added query features... I prefer the first as this seems like duplicate info

appendices/versions.txt
krOoze: Is it? `VkPhysicalDeviceVulkan11Features` was added in Vulkan 1.2, so I should not be referencing it in 1.1.

appendices/versions.txt
krOoze: There are no other added feature bools. All the others are promoted, and protected memory is mentioned above.

appendices/versions.txt
sfricke-samsung: ok, so it was added to the spec in the 1.2 timeframe, but if I load 
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html
it is there

Think of it as:

it was not in 1.1.`N`, but it is in 1.1.`N+1`

sfricke-samsung: nvm, that is totally wrong and breaks backward compatibility... sorry, you are totally correct on this!

appendices/versions.txt
```diff
@@ -188,37 +188,70 @@ number of key extensions into the core API:
...
+endif::VK_VERSION_1_2[]
+when queried via flink:vkGetPhysicalDeviceFeatures2.
 
-=== Differences relative to `VK_KHR_16bit_storage`
+=== Differences relative to `VK_KHR_shader_draw_parameters`
 
-If the `<<VK_KHR_16bit_storage>>` extension is not supported, support for
-the <<features-storageBuffer16BitAccess,storageBuffer16BitAccess>> feature
-is optional.
+If the `<VK_KHR_shader_draw_parameters>>` extension is not supported,
```
sfricke-samsung, line 220:
```suggestion
If the `<VK_KHR_shader_draw_parameters>` extension is not supported,
```

appendices/versions.txt
```diff
@@ -188,37 +188,70 @@ number of key extensions into the core API:
...
-slink:VkPhysicalDeviceVulkan11Features::pname:samplerYcbcrConversion when
-queried via flink:vkGetPhysicalDeviceFeatures2.
+slink:VkPhysicalDeviceSamplerYcbcrConversionFeatures::pname:samplerYcbcrConversion
+ifdef::VK_VERSION_1_2[]
+or slink:VkPhysicalDeviceVulkan11Features::pname:samplerYcbcrConversion
+endif::VK_VERSION_1_2[]
+when queried via flink:vkGetPhysicalDeviceFeatures2.
 
-=== Differences relative to `VK_KHR_16bit_storage`
+=== Differences relative to `VK_KHR_shader_draw_parameters`
```
sfricke-samsung, line 218: So I actually don't think there should be a `Differences relative to VK_KHR_shader_draw_parameters` (and why we didn't add it to start with) because this suggest there WAS a `shaderDrawParameters` prior to 1.1

The reason for things like VK_KHR_16bit_storage is because 
- in 1.0 if you exposed the `storageBuffer16BitAccess` is guaranteed
- in 1.1 the extension is not core, but a 1.1 implementation doesn't guarantee `storageBuffer16BitAccess`

This logic is not an issue with `VK_KHR_shader_draw_parameters` as it was all additive

sfricke-samsung: Same thing with`VK_KHR_variable_pointers`/`VkPhysicalDeviceVariablePointersFeatures`

chapters/features.txt
```diff
@@ -50,20 +50,23 @@ must: be supported.
 |====
 | Extension                                 | Feature(s)
 ifdef::VK_VERSION_1_2[]
+ifdef::VK_KHR_shader_draw_parameters[]
+| `<<VK_KHR_shader_draw_parameters>>`       | <<features-shaderDrawParameters,pname:shaderDrawParameters>>
```
sfricke-samsung, line 54: I agree this still should be here as this seems correct

chapters/features.txt
```diff
@@ -3489,6 +3492,10 @@ All Vulkan graphics implementations must: support the following features:
   * <<features-robustBufferAccess, pname:robustBufferAccess>>
 ifdef::VK_VERSION_1_1[]
   * <<features-multiview, pname:multiview>>, if Vulkan 1.1 is supported.
+ifdef::VK_KHR_shader_draw_parameters[]
```
sfricke-samsung, line 3495: if you have a 1.0 implementation there is no `shaderDrawParameters` to expose

appendices/versions.txt
krOoze: I used the 1.2 chapter as a guide. See e.g. [Differences relative to VK_KHR_sampler_mirror_clamp_to_edge](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#_differences_relative_to_vk_khr_sampler_mirror_clamp_to_edge). Basically anything that is not trivial promotion has this difference section.

chapters/features.txt
krOoze: ergo the `ifdef::VK_VERSION_1_1[]`.

appendices/versions.txt
sfricke-samsung: That is not the criteria for this section

Anything non-trival (or any changes) belongs in the Extension appendix (ex `./chapters/VK_KHR_sampler_mirror_clamp_to_edge.txt`)

 This section is ONLY for things that were non-optional that became optional in order to fold the extension into core (

Example: `VK_KHR_timeline_semaphore` the `timelineSemaphore` was added in `VkPhysicalDeviceVulkan12Features` but there is not `Difference relative to` section because a 1.2 implementation has to support `timelineSemaphore`

The sole purpose of these sections was to help point out "These extensions are going into core, but the original guarantees of features still need to be queried as the 'guaranteed feature support' wasn't folded into core"

krOoze: Timeline semaphores were trivial promotion.

For 1:1 comparison see `VK_KHR_draw_indirect_count` or the `VK_KHR_sampler_mirror_clamp_to_edge`.

Previously the feature was non-optional (as determined via the feature alias of the extension support). Now it is optional in core.

appendices/versions.txt
```diff
@@ -188,37 +188,71 @@ number of key extensions into the core API:
...
+support for:
+
+  * The <<shaders-group-operations, group operations>> and
+    <<shaders-scope-subgroup, subgroup scope>>.
+  * The <<memory-protected-memory, protected memory>> feature.
+  * A new command to enumerate the instance version:
+    flink:vkEnumerateInstanceVersion.
+  * The <<features-shaderDrawParameters, pname:shaderDrawParameters>>
+    feature query struct, where `<<VK_KHR_shader_draw_parameters>>` did have
+    none.
```
krOoze, line 255: There are some massive Swapchain changes that I should also note here.

appendices/versions.txt
sfricke-samsung: I am getting 2nd opinions inside Khronos, wanted to leave comment to acknowledge I haven't forgotten about this :)

sfricke-samsung: So after thinking about it more, I agree with these changes as whole

appendices/versions.txt
```diff
@@ -188,37 +188,71 @@ number of key extensions into the core API:
...
+when queried via flink:vkGetPhysicalDeviceFeatures2.
 
-=== Differences relative to `VK_KHR_16bit_storage`
+=== Differences relative to `VK_KHR_shader_draw_parameters`
 
-If the `<<VK_KHR_16bit_storage>>` extension is not supported, support for
-the <<features-storageBuffer16BitAccess,storageBuffer16BitAccess>> feature
-is optional.
+If the `<<VK_KHR_shader_draw_parameters>>` extension is not supported,
+support for the <<features-shaderDrawParameters, pname:shaderDrawParameters>>
```
sfricke-samsung, line 221: `VK_KHR_shader_draw_parameters` should match more of the wording of `VK_KHR_draw_indirect_count` as both didn't have the feature bit prior to being promoted

I suggest

> If the VK_KHR_shader_draw_parameters extension is not supported, shader draw parameters are optional. Support for this feature is defined by  ...

krOoze: I believe it already is. It is copied from `VK_EXT_descriptor_indexing`, I think.
What is the exact diff here that you see?

sfricke-samsung: There was no `shaderDrawParameters` before 1.1, Prior if you had the extension enabled, it then allowed you to have shader draw parameters, therefore, like `VK_KHR_draw_indirect_count` which also didn't have a feature bit, falls in the same group

The difference is "there was no feature bit in the extension and it was added with the promotion" it is a subtle detail, but there isn't much more to it than that

krOoze: Yea I get that. I mean, where do you see the difference between my wording and the `VK_KHR_draw_indirect_count`?

sfricke-samsung: We have

> If the VK_KHR_draw_indirect_count extension is not supported, support for the entry points vkCmdDrawIndirectCount and vkCmdDrawIndexedIndirectCount is optional.

not 

> If the VK_KHR_draw_indirect_count extension is not supported, support for the <<features-drawIndirectCount, pname:drawIndirectCount>> is optional.

because `drawIndirectCount` was not there prior to promotion, same with `shaderDrawParameters`

krOoze: The `VK_EXT_descriptor_indexing` (which also didn't have a feature flag) has it the way I do. Probably needs some consistency changes later. Though what you suggest might be unsustainable for some of the extensions, as it would require naming ~20 entrypoints, enums, and pnext structs that it added.

For now I am just gonna do it then, as it seems it only adds `SPV_KHR_shader_draw_parameters`

krOoze: done

sfricke-samsung: I agree `VK_EXT_descriptor_indexing` should be fixed (I can take care of that)

appendices/versions.txt
```diff
@@ -188,37 +188,72 @@ number of key extensions into the core API:
...
+
+=== Additional Vulkan 1.1 Feature Support
+
+[[versions-1.1-new-features]]
+In addition to the promoted extensions described above, Vulkan 1.1 added
+support for:
+
+  * The <<shaders-group-operations, group operations>> and
+    <<shaders-scope-subgroup, subgroup scope>>.
+  * The <<memory-protected-memory, protected memory>> feature.
```
dgkoch, line 251:
```suggestion
  * The <<memory-protected-memory, pname:protectedMemory>> feature.
```

appendices/versions.txt
```diff
@@ -188,37 +188,72 @@ number of key extensions into the core API:
...
+  * A new command to enumerate the instance version:
+    flink:vkEnumerateInstanceVersion.
+  * The <<features-shaderDrawParameters, pname:shaderDrawParameters>>
+    feature query struct, where `<<VK_KHR_shader_draw_parameters>>` did have
```
dgkoch, line 255:
```suggestion
    feature query struct, flink:VkPhysicalDeviceShaderDrawParametersFeatures, where `<<VK_KHR_shader_draw_parameters>>` had
```

appendices/versions.txt
dgkoch: Swapchains are in core so that doesn't seem like it should be added here.

appendices/versions.txt
```diff
@@ -110,8 +110,8 @@ queried via flink:vkGetPhysicalDeviceFeatures2.
 === Differences relative to `VK_KHR_shader_float16_int8`
 
 If the `<<VK_KHR_shader_float16_int8>>` extension is not supported, support
-for at least one of <<features-shaderFloat16,shaderFloat16>> and
-<<features-shaderInt8,shaderInt8>> feature is optional.
+for at least one of <<features-shaderFloat16, pname:shaderFloat16>> and
+<<features-shaderInt8, pname:shaderInt8>> feature is optional.
```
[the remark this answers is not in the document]
krOoze: Well I just fixed the markup as collateral damage of the PR. I have not really otherwisely touched the 1.2 section.
The PR already little bit lacks focus. Not sure I should further be expanding it.

appendices/versions.txt
krOoze: IDK, this links to the Protected Memory chapter. Unmangled human-readable name seems appropriate here.

appendices/versions.txt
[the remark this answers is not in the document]
dgkoch: I guess I can fix it in an internal MR, but it'll generate a conflict here..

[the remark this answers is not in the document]
dgkoch: I'll do the pname changes as well, so you can just drop this hunk.

appendices/versions.txt
krOoze: Well, the 1.1 modifies the pre-existing extension. I am on the fence. If one is being pedantic, then maybe it does not belong here. On the other hand for the sake of completeness, it feels like it should.

appendices/versions.txt
dgkoch: I don't care too much about this one.

appendices/versions.txt
dgkoch: please do it in a separate MR, regardless. This one has already grown way beyond it's scope.

appendices/versions.txt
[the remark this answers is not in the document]
krOoze: Alright,

appendices/versions.txt
krOoze: Yup

appendices/versions.txt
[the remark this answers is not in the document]
krOoze: Transfered all the markup fixes in 1.2 elsewhere.

appendices/versions.txt
krOoze: Changed to the struct, and reworded a bit to have a less cute sentence structure overal.

TomOlson:
@krOoze, it looks like you are withdrawing from participating in Vulkan. On behalf of the Working Group, thanks for your contributions over the past several years - they are much appreciated. If you ever feel like working on the spec again, you'll be welcome.

Issue #1284, opened 2020-05-28
Bugs with proposed fixes for the core+KHR specification version 1.2.141


naiveforce:
[vkspec.pdf](https://github.com/KhronosGroup/Vulkan-Docs/files/4697549/vkspec.pdf)

Suggested fixes as comments placed throughout PDF build of KHR extended version 141 of the specification.

Comment primitives used and their intended meaning:
- red strike-trough: designates parts of the text that should be removed (may include single character sequences, punctuation marks as well as white spaces);
- green underline: designates parts of the text that should be replaced (see comment attached to each item for details)
- orange highlight: designates misc. editorial suggestions (see comment attached to each item for details)
- sticky note: designates parts of the text that might be missing (see comment attached to each item for details)

There are numerous VUs missing pNext description. I had being marking them, up to a point I came to a conclusion that it surely must be by design.

Most of the text in drawing-command-related VUs is duplicated across multiple variants of basic drawing/dispatching commands. Personally I found repeatedly reading same content over and over again distracting and tedious. Not to mention that it surely must be hard to maintain.

Consider using "comments view/list" or some sort of "go to next/previous comment" functionality of your PDF-viewer to not get overwhelmed or miss anything.
Verified to view correctly in Adobe Acrobat Reader and built-in Firefox PDF-viewer.

Hope you will find it useful.

krOoze:
Ok, 653 items; sounds like fun.
I keep accumulator branch for these kind of little things myself; I am gonna see if I can merge some of your stuff in there (unless you intend to take this on yourself).

> There are numerous VUs missing `pNext` description. I had being marking them, up to a point I came to a conclusion that it surely must be by design.

Yes, that is so (https://github.com/KhronosGroup/Vulkan-Docs/issues/1239).

naiveforce:
By all means use it as you see fit. I've spent all of my mental stamina on repeatedly going through the spec and trying to catch as much as possible. Believe me adding these as comments to PDF was no fun either. By the time I've reached the final page it started to take up to several seconds just to create a new, empty comment.

krOoze:
Hey, gotta enjoy all those AI-grade tasks while you still can. :p

oddhack:
The repeated VUs are intentional as each of them turns into a separate validation path in the validation layers. They are not hard to maintain as they're only written once in the spec markup.

oddhack:
[...]

PR #1339, opened 2020-07-30
Remove noautovalidity from pNexts in vk.xml

krOoze:
ex #1284

diff:
diff1
diff2
diff3
diff4

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -3163,7 +3163,7 @@ typedef void <name>CAMetalLayer</name>;
         </type>
         <type category="struct" name="VkDebugUtilsMessengerCallbackDataEXT">
             <member values="VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CALLBACK_DATA_EXT"><type>VkStructureType</type> <name>sType</name></member>
-            <member optional="true">const <type>void</type>*                                                        <name>pNext</name></member>
+            <member>const <type>void</type>*                                                                        <name>pNext</name></member>
             <member optional="true"><type>VkDebugUtilsMessengerCallbackDataFlagsEXT</type>                          <name>flags</name></member>
             <member optional="true" len="null-terminated">const <type>char</type>*                                  <name>pMessageIdName</name></member>
             <member optional="true"><type>int32_t</type>                                                            <name>messageIdNumber</name></member>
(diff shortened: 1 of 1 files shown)


xml/vk.xml
```diff
@@ -3163,7 +3163,7 @@ typedef void <name>CAMetalLayer</name>;
         </type>
         <type category="struct" name="VkDebugUtilsMessengerCallbackDataEXT">
             <member values="VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CALLBACK_DATA_EXT"><type>VkStructureType</type> <name>sType</name></member>
-            <member optional="true">const <type>void</type>*                                                        <name>pNext</name></member>
+            <member>const <type>void</type>*                                                                        <name>pNext</name></member>
```
krOoze, line 3166: Plus remove this `optional`, which likely has no purpose.

jeffbolznv:
Many of these look wrong. The purpose of `noautovalidity` is to make it so only the base structure in a chain gets the auto-generated VU, and I think this matches how the layers implement it. The styleguide says (about `structextends`):

```
  * attr:structextends ...
    This should usually only list the top-level structure that is
    extended, for all possible chained structures - the chained structures
    themselves should have their `pNext` tagged with attr:noautovalidity.
```

Some of them, like the renderpass2 stuff, look like correct changes, but I think all the Features/Properties structs should be reverted.

krOoze:
In that case, it is actually mising in many places. E.g. `VkPhysicalDeviceVariablePointersFeatures`. If layers relied on this, it would be broken.

Perphaps ping me what is the chosen way forward before I instead start adding `noautovalidity`ies at bulk into the XML.

oddhack:
If everyone's now agreed that structures with the structextends attribute should also have noautovalidity on pNext, which will suppress the "pNext must be NULL" VUs, I'll take care of that. It's easier for us to handle those sorts of bulk XML edits internally. If there's somewhere else that you think noautovalidity does or does not belong, or something else like the 'optional' you refer to above, please bring that out as a separate issue.

oddhack:
Hm. Adding or removing noautovalidity from pNext actually makes no difference at all in the generated VU for extending structures. @tobski would probably have to explain the guts of the validity generator but this makes sense to me. Actually the most useful thing to do might probably be to remove the comment from the style guide about setting noautovalidity on pNext in extending structures, since it's irrelevant.

krOoze:
@oddhack Yes, the generator does not care. It cares only if it sees `structextends`, and so it knows it is extension struct. And as such the `novalidity` does nothing. Though that logic is slightly broken, as a `struct` could be both extension and top-level, as is the case of e.g. `VkDebugUtilsMessengerCreateInfoEXT`.

https://github.com/KhronosGroup/Vulkan-Docs/blob/68bedacc6014941d5c04d168ef62df82031e0ddc/scripts/validitygenerator.py#L1012-L1017

PR #1342, opened 2020-08-03
Remove redundant transient image VU

krOoze:
Redundant to `VUID-VkImageCreateInfo-usage-00963`, which says:

> If `usage` includes `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT`, then bits other than `VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT`, `VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT`, and `VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT` **must** not be set


TomOlson:
Unless I'm missing something, these are not actually redundant. `00963` constrains other `usage` flags based on `USAGE_TRANSIENT_ATTACHMENT_BIT`. `01925` constrains `USAGE_TRANSIENT_ATTACHMENT_BIT` based on the `create` flags. The first constraint does not imply the second. That fact that they are expressed in the opposite order is a little bit annoying, but makes sense given that they are in two different VU blocks.

krOoze:
Right, I think so. I must have thought those were usage flags. Apologies.

PR #1347, opened 2020-08-13
Fix null type for VkQueue

krOoze:


chapters/devsandqueues.txt
```diff
@@ -1894,7 +1894,7 @@ The queue returned by fname:vkGetDeviceQueue2 must: have the same
 pname:flags value from this structure as that used at device creation time
 in a slink:VkDeviceQueueCreateInfo instance.
 If no matching pname:flags were specified at device creation time then
-pname:pQueue will return dlink:VK_NULL_HANDLE.
+pname:pQueue will return `NULL` handle.
```
oddhack, line 1897: Just NULL is fine.

fluppeteer: Standing, I think, by a query I made in private a minute ago: surely pQueue will *be* NULL. What *returns* it is vkGetDeviceQueue2().

krOoze: @oddhack I am little bit unconfortable with that, as `pQueue` is a pointer to handle. I could wordsmith around it as "the handle returned in `pQueue` must: be `NULL`".
@fluppeteer Above should also fix your nit, I think.

fluppeteer: @krOoze - thanks, it certainly fixes the nit, and I believe you on the rest (without having had much of a look at the context myself).

PR #1350, opened 2020-08-17
Fix language around device extensions implemented by layers

pmuetschard:
Device extensions implemented by layers require the *layer* to be specified at `vkCreateInstance` time, while the *extension* needs to be specified at `vkCreateDevice` time.

### DIFF
--- a/chapters/devsandqueues.txt
+++ b/chapters/devsandqueues.txt
@@ -1044,9 +1044,9 @@ flink:vkGetPhysicalDeviceFeatures.
 
 After verifying and enabling the extensions the sname:VkDevice object is
 created and returned to the application.
-If a requested extension is only supported by a layer, both the layer and
-the extension need to be specified at fname:vkCreateInstance time for the
-creation to succeed.
+If a requested extension is only supported by a layer, the layer needs to
+be specified at fname:vkCreateInstance time and the extension at
+fname:vkCreateDevice time for the creation to succeed.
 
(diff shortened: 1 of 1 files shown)


chapters/devsandqueues.txt
```diff
@@ -1044,9 +1044,9 @@ flink:vkGetPhysicalDeviceFeatures.
 
 After verifying and enabling the extensions the sname:VkDevice object is
 created and returned to the application.
-If a requested extension is only supported by a layer, both the layer and
-the extension need to be specified at fname:vkCreateInstance time for the
-creation to succeed.
+If a requested extension is only supported by a layer, the layer needs to
+be specified at fname:vkCreateInstance time and the extension at
+fname:vkCreateDevice time for the creation to succeed.
```
krOoze, line 1049: Right, seems like there was a copy-paste error.

Actually, I would think a device extension supported only by a layer would simply not be enumerated by `vkEnumerateDeviceExtensionProperties`, and as such this whole paragraph is unnecessary?

Tobski: I think I agree that the sentence at least is unnecessary; this made more sense when device layers were a thing, which they no longer are. We should probably just delete this. @pmuetschard would you be able to update the PR to remove this sentence altogether?

oddhack: @pmuetschard we went ahead and fixed this internally in the 1.2.157 spec update as @tobski suggested, so closing this.

Issue #1354, opened 2020-08-23
Add a way to identify the DRM node of a physical device 


emersion:
One of the few missing specs to write a Vulkan Wayland compositor is a way to map Vulkan physical devices and DRM nodes. The best that can be done right now is using PCI IDs ([example](https://github.com/nyorain/kms-vulkan/blob/e66655b05ebc6281c1ebffec12681240caf305be/vulkan.c#L255)). However some GPUs aren't PCI (e.g. platform devices).

It would be nice to be able to get a reliable way to do the mapping. See https://gitlab.freedesktop.org/wayland/wayland-protocols/issues/10 for a list of ways to refer to DRM devices.

cubanismo:
Agreed, this would be useful, and frankly I think we just hadn't got around to it yet.  Would you mind writing up a patch to add the necessary structures?

We discussed a bit at Khronos and it sounds like we're in agreement that exposing a minor number for both the primary and render node would be a good solution.  We also noted the API may need to accommodate Vulkan devices which do not have one or both types of DRM device nodes, as there may be, for example, SW rendering engines exposed as Vulkan devices, and devices without display subsystems that do not wish to expose control or primary nodes.

emersion:
Submitted a draft: https://github.com/KhronosGroup/Vulkan-Docs/pull/1356

PR #1356, opened 2020-08-26
Add specification for VK_EXT_physical_device_drm

emersion:
Closes: https://github.com/KhronosGroup/Vulkan-Docs/issues/1354

### DIFF
--- a/appendices/VK_EXT_physical_device_drm.txt
+++ b/appendices/VK_EXT_physical_device_drm.txt
@@ -1,0 +1,46 @@
+// Copyright (c) 2020 Simon Ser
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_EXT_physical_device_drm.txt[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2021-06-09
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+    - Simon Ser
+
+=== Description
+
+This extension provides new facilities to query DRM properties for physical
+devices, enabling users to match Vulkan physical devices with DRM nodes on
+Linux.
+
+Its functionality closely overlaps with
+`EGL_EXT_device_drm`^<<VK_EXT_physical_device_drm-fn1,1>>^.
+Unlike the EGL extension, this extension does not expose a string containing
+the name of the device file and instead exposes device minor numbers.
+
+DRM defines multiple device node types. Each physical device may have one
+primary node and one render node associated. Physical devices may have no
+primary node (e.g. if the device doesn't have a display subsystem), may have
+no render node (e.g. if it's a software rendering engine), or may have neither
+(e.g. if it's a software rendering engine without a display subsystem).
+
+To query DRM properties for a physical device, chain
+slink:VkPhysicalDeviceDrmPropertiesEXT to slink:VkPhysicalDeviceProperties2.
+
+include::{generated}/interfaces/VK_EXT_physical_device_drm.txt[]
+
+=== References
+
+  . [[VK_EXT_physical_device_drm-fn1]]
+    https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_device_drm.txt[`EGL_EXT_device_drm`]
+
+=== Version History
+
+  * Revision 1, 2021-06-09
+    - First stable revision
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -5894,6 +5894,16 @@ typedef void <name>CAMetalLayer</name>;
             <member><type>size_t</type>                 <name>extraCount</name></member>
             <member>const <type>void</type>* const *    <name>pExtras</name></member>
         </type>
+        <type category="struct" name="VkPhysicalDeviceDrmPropertiesEXT" returnedonly="true" structextends="VkPhysicalDeviceProperties2">
+            <member values="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRM_PROPERTIES_EXT"><type>VkStructureType</type> <name>sType</name></member>
+            <member><type>void</type>* <name>pNext</name></member>
+            <member><type>VkBool32</type> <name>hasPrimary</name></member>
+            <member><type>VkBool32</type> <name>hasRender</name></member>
+            <member><type>int64_t</type> <name>primaryMajor</name></member>
+            <member><type>int64_t</type> <name>primaryMinor</name></member>
+            <member><type>int64_t</type> <name>renderMajor</name></member>
+            <member><type>int64_t</type> <name>renderMinor</name></member>
+        </type>
     </types>
     <comment>Vulkan enumerant (token) definitions</comment>
 
@@ -15915,10 +15925,14 @@ typedef void <name>CAMetalLayer</name>;
                 <command name="vkCmdSetVertexInputEXT"/>
             </require>
         </extension>
-        <extension name="VK_EXT_extension_354" number="354" author="EXT" contact="Simon Ser @emersion" supported="disabled">
+        <extension name="VK_EXT_physical_device_drm" number="354" author="EXT" type="device" contact="Simon Ser @emersion" supported="vulkan" requires="VK_KHR_get_physical_device_properties2">
             <require>
-                <enum value="0"                                             name="VK_EXT_EXTENSION_354_SPEC_VERSION"/>
-                <enum value="&quot;VK_EXT_extension_354&quot;"              name="VK_EXT_EXTENSION_354_EXTENSION_NAME"/>
+                <enum value="1"                                             name="VK_EXT_PHYSICAL_DEVICE_DRM_SPEC_VERSION"/>
+                <enum value="&quot;VK_EXT_physical_device_drm&quot;"        name="VK_EXT_PHYSICAL_DEVICE_DRM_EXTENSION_NAME"/>
+
+                <enum offset="0" extends="VkStructureType" name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRM_PROPERTIES_EXT"/>
+
+                <type name="VkPhysicalDeviceDrmPropertiesEXT"/>
             </require>
         </extension>
         <extension name="VK_EXT_extension_355" number="355" author="EXT" contact="Ralph Potter gitlab:@r_potter" supported="disabled">
--- a/chapters/devsandqueues.txt
+++ b/chapters/devsandqueues.txt
@@ -693,6 +693,34 @@ These are properties of the PCI bus information of a physical device.
 include::{generated}/validity/structs/VkPhysicalDevicePCIBusInfoPropertiesEXT.txt[]
 --
 endif::VK_EXT_pci_bus_info[]
+
+ifdef::VK_EXT_physical_device_drm[]
+[open,refpage='VkPhysicalDeviceDrmPropertiesEXT',desc='Structure containing DRM information of a physical device',type='structs']
+--
+
+To query the DRM information of a physical device, add a
+slink:VkPhysicalDeviceDrmPropertiesEXT structure to the pname:pNext chain of
+the slink:VkPhysicalDeviceProperties2 structure.
+The sname:VkPhysicalDeviceDrmPropertiesExt structure is defined as:
+
+include::{generated}/api/structs/VkPhysicalDeviceDrmPropertiesEXT.txt[]
+
+  * pname:sType is the type of this structure.
+  * pname:pNext is `NULL` or a pointer to a structure extending this structure.
+  * pname:hasPrimary is a boolean indicating whether the physical device has a
+    DRM primary node.
+  * pname:hasRender is a boolean indicating whether the physical device has a
+    DRM render node.
+  * pname:primaryMajor is the DRM primary node major number, if any.
+  * pname:primaryMinor is the DRM primary node minor number, if any.
+  * pname:renderMajor is the DRM render node major number, if any.
+  * pname:renderMinor is the DRM render node minor number, if any.
+
+include::{generated}/validity/structs/VkPhysicalDeviceDrmPropertiesEXT.txt[]
+--
+
+endif::VK_EXT_physical_device_drm[]
+
 endif::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
 
 [open,refpage='vkGetPhysicalDeviceQueueFamilyProperties',desc='Reports properties of the queues of the specified physical device',type='protos']


oddhack:
You will need to make a separate PR which just reserves the extension number as a "disabled" extension (there are examples in the XML). I can accept that immediately while the actual extension develops. But extension 353 has already been reserved, so reserve 354 instead.

emersion:
Sure! Done in https://github.com/KhronosGroup/Vulkan-Docs/pull/1360.

oddhack:
> Sure! Done in #1360.

Thanks. Now just rebase off current main branch and you can take your time finishing the extension.

xml/vk.xml
```diff
@@ -4861,6 +4861,12 @@ typedef void <name>CAMetalLayer</name>;
             <member><type>VkBool32</type>                           <name>formatA4R4G4B4</name></member>
             <member><type>VkBool32</type>                           <name>formatA4B4G4R4</name></member>
         </type>
+        <type category="struct" name="VkPhysicalDeviceDrmPropertiesEXT" returnedonly="true" structextends="VkPhysicalDeviceProperties2">
+            <member values="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRM_PROPERTIES_EXT"><type>VkStructureType</type> <name>sType</name></member>
+            <member><type>void</type>*                              <name>pNext</name></member>
+            <member><type>int32_t</type>                           <name>primary_minor</name></member>
+            <member><type>int32_t</type>                           <name>render_minor</name></member>
```
djdeath, line 4868: major/minor numbers are defined as unsigned :+1: 
```
unsigned int major(dev_t dev);
unsigned int minor(dev_t dev);
```

Are you planning on using `int32_t` as a way to report error/unknown?

emersion: Yeah, my plan was to make -1 mean "there is no node of this type".

cubanismo: In practice, aren't major/minor <= 255?  I had thought using int32_t and -1 for unknown would be a good idea personally, but there could always be a separate explicit "not a DRM device" (or separate "no render node" and "no primary node" if appropriate) field as well if others prefer.

The MR looks like a good start to me.  Will review again when the spec language is filled in.

emersion:
I've added some spec language. This is my first time writing a Vulkan spec, please let me know if anything's wrong or missing.

emersion:
It sounds like only sending the minor number wouldn't be enough on FreeBSD, because the major number isn't fixed: https://www.freebsd.org/cgi/man.cgi?query=major

So we'd need to add major numbers here.

cubanismo:
Seems reasonable.  Made me curious how one finds a device under /dev: https://www.freebsd.org/cgi/man.cgi?query=devname is a pretty cool family of functions.  The language looks fine so far to me.  Will review again when the major number support is there.  Was the unsigned/signed issue raised above resolved one way or the other?

xml/vk.xml
emersion: >In practice, aren't major/minor <= 255?

On [Linux](https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/typesizes.h;h=8f16903a216aae5d0fb58b2aae4f00d6972e9b00;hb=HEAD#l29) and [FreeBSD](https://github.com/freebsd/freebsd/blob/4bfbbd74a8206fc222ca8b3ee874a963ee6900e0/sys/sys/_types.h#L113) user-space, `dev_t` is a 64-bit type. However [in the Linux kernel](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/kdev_t.h) major/minor can only be <= 255 as you said. On FreeBSD [the full 64-bit value may be used](https://github.com/freebsd/freebsd/blob/b3866f4bf2a28d20aa9a91a6f8737394e3546837/sys/sys/types.h#L381). OpenBSD seems to have a minor that can [go over 255](https://github.com/openbsd/src/blob/13001500586b3e292b0572e147c78146228ce70c/sys/sys/types.h#L211).

On Linux and OpenBSD, major/minor are defined to return an `unsigned int`. On FreeBSD, they are defined to return an `int`.

Ultimately, major/minor are system-specific and may be (1) bigger than `int32_t` (2) undefined (libdrm doesn't (yet) require that `dev_t` can be split into major/minor). One idea would be to embed `dev_t` fields in the struct directly, but I'm not sure it's a good idea (and this wouldn't make it any easier to transfer over the wire, e.g. for Wayland).

Overall, I'm starting to re-consider the use of major/minor, because finding a way to use them that doesn't put us into a corner is tricky and it sounds like more pain with little benefit. Maybe we should do like the EGL extension and provide a path as a string. The benefits of major/minor over a device node path I can think of are:

- Not tied to a specific file-system layout, so works when exchanging them between namespaces that may have `/dev/dri` set up differently. But we don't really care about this for Vulkan, since the Vulkan driver and the app run in the same process. Also a lot of clients expect `/dev/dri` to be set up correctly anyways…
- Can easily compare two devices for equality without having to deal with symlinks/etc. Side note, major/minor are still different for the primary and render nodes. If the app gets a path, it needs to open it, use `drmGetDevice` and then `drmDevicesEqual` to compare them. Or it could just use `realpath`.

I'm also not 100% clear of how we would "return" a string from `vkGetPhysicalDeviceProperties2`. Use a `const char *` field?

r-potter: > I'm also not 100% clear of how we would "return" a string from vkGetPhysicalDeviceProperties2. Use a const char * field?

We typically do this by populating a fixed size array. For example, VkPhysicalDeviceDriverProperties in VK_KHR_driver_properties does this:

```
// Provided by VK_VERSION_1_2
typedef struct VkPhysicalDeviceDriverProperties {
    VkStructureType         sType;
    void*                   pNext;
    VkDriverId              driverID;
    char                    driverName[VK_MAX_DRIVER_NAME_SIZE];
    char                    driverInfo[VK_MAX_DRIVER_INFO_SIZE];
    VkConformanceVersion    conformanceVersion;
} VkPhysicalDeviceDriverProperties;
```

cubanismo: POSIX seems to have failed us here.

I don't have quite as bleak an outlook though. On both Linux and the BSDs, major and minor are values that fit in 32-bits, and likely always will be unless those OS's want to burn down the world. I'd suggest we simply return int64_t major/minor numbers, assert their range is limited such that the value can always be safely cast to the type used by the major() and minor() functions on the OS on which the Vulkan instance is running, and use a separate boolean field to indicate cases where there is no corresponding device node.

I think returning a string would be mildly problematic because:

* It may be hard to pick a fixed length for a path string. I think POSIX requires things be under /dev, and I'd guess most OS's limit themselves to PATH_MAX as a practical matter, but there's no real guarantee AFAIK.
* I'm assuming there may be cases where sandboxes disallow access to /dev or even present a very lightly populated copy of /dev to apps, but apps may still want to match some piped-in FDs against the Vulkan device using fstat().
* In practice, our Linux driver (and likely others on Linux) has to make assumptions at build time about the layout of /dev or add a dependency on udev and/or libdrm to provide such a string, and this sort of tradeoff/choice is better left to apps IMHO.

emersion: OK, all of that makes sense to me. major/minor isn't the best solution, but yeah the alternatives aren't great either.

Just to make sure we haven't missed an opportunity: can you confirm it'd be cumbersome to embed some `dev_t` fields in the Vulkan struct? This would require to `#include <sys/types.h>` on Unix, but would avoid using any system-specific API ([`dev_t` is POSIX](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/types.h.html)).

emersion:
Updated to include bool and major fields. This is ready for another review round.

emersion:
Gentle ping

emersion:
Ping

r-potter:
@stonesthrow This could probably use a look in the SI TSG in case anyone other than cubanismo has reviews to contribute (unless @oddhack prefers to route it through the maintenance call)

stonesthrow:
cubanismo was waiting for response from emersion. We can review and discuss this week.

r-potter:
This should probably also have a feature structure:

```
typedef struct VkPhysicalDeviceDrmFeaturesEXT {
    VkStructureType         sType;
    void*                           pNext;
    VkBool32                    physicalDeviceDrmProperties;
} VkPhysicalDeviceDrmFeaturesEXT;
```

I realise it doesn't really do anything in this case, but we now require them as a matter of course for consistency/future-proofing. You would need to add the relevant description for the struct, and to add something to the `Feature Requirements` section to force anyone implementing the extension to also support the feature:

```
ifdef::VK_EXT_physical_device_drm[]
  * <<features-physicalDeviceDrmProperties, pname:physicalDeviceDrmProperties>>
    if the `<<VK_EXT_physical_device_drm>>` extension is supported.
endif::VK_EXT_physical_device_drm[]
```

appendices/VK_EXT_physical_device_drm.txt
```diff
@@ -0,0 +1,43 @@
...
+
+Its functionality closely overlaps with
+`EGL_EXT_device_drm`^<<VK_EXT_physical_device_drm-fn1,1>>^.
+Unlike the EGL extension, this extension does not expose a string containing
+the name of the device file and instead exposes device minor numbers.
+
+DRM defines multiple device node types. Each physical device may have one
+primary node and one render node associated. Physical devices may have no
+primary node (e.g. if the device doesn't have a display subsystem) or may have
+no render node (e.g. if it's a software rendering engine).
```
cubanismo, line 30: Might be worth making it more explicit that devices may have neither (E.g., a SW rendering engine with no display system, or the NVIDIA driver if the user opts not to load our optional DRM support module).

cubanismo:
We discussed process in today's system integration meeting.

The spec content appears largely complete, pending resolution of the minor issues noted above. The Vulkan working group currently asks that EXT (multivendor) extensions ensure reviewed CTS coverage and validation layer support is available as well, so I think the next step is adding that. This doesn't need to be anything terribly complicated, and a simple touch test of the APIs is probably sufficient here. @emersion, is that something you're willing to take on?

emersion:
Sorry, I've tried to look into this twice, but I just can't figure out how to do it. I'd rather make this extension a MESA-vendored one.

emersion:
>This should probably also have a feature structure

Do we really need that? Can't we introduce it later on if we need to?

emersion:
Renamed as a MESA extension to sidestep the paperwork.

Please review.

emersion:
Ping

stonesthrow:
My apologies emersion, We discussed this PR a few weeks ago, and I had an action to contact you. As a vulkan extension/feature we need a CTS test for this before merging and closing. Are you familiar with the CTS suite of test and are you interested in creating a test - similar to what is done for other platforms?

emersion:
Yes I know what the CTS is, but I don't want to touch it. I renamed the extension to MESA to avoid the CTS. I don't have time to work on validation layers or the CTS.

cubanismo:
It would be a shame to drop multi-vendor specifications down to vendor-specific over the "paperwork," and also to miss out on the benefits of said paperwork at the same time. Would you be willing to restore the EXT prefix if I wrote the CTS tests? I'm thinking of a very light test that verifies the following:

-If the extension is reported for a given device, the including the extension struct in a physical device properties query succeeds.

-for [hasPrimary,hasRender]; if the property is true, scan /dev/dri/* for a matching device node, fail if one is not found.

This takes a few liberties, in that it assumes the dri device node path (to avoid a broader search) and population of that path with device nodes for all valid DRM devices in use by Vulkan, but that seems reasonable to me and seems like sufficient coverage to verify the implementation is sane.

emersion:
> miss out on the benefits of said paperwork

Kind of off-topic, but:

- I'm only interested in community-friendly open-source drivers. That means Mesa.
- The CTS is a complicated C++ codebase.
- A CTS test for this extension has questionable value. This extension is trivial, there's little room for error on the driver side. If there's a bug, the CTS won't catch the most important ones, e.g. missing major/minors hints or swapped hints.

I think implementation requirements would make much more sense than CTS requirements. That said...

> Would you be willing to restore the EXT prefix if I wrote the CTS tests?

Thanks for stepping up. I'll restore the EXT prefix and I'll start working on a Mesa implementation.

emersion:
>scan /dev/dri/*

BTW, this won't work on non-Linux platforms such as FreeBSD. The right way to do it is by using libdrm's `drmGetDevices`.

emersion:
RADV implementation: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/8390

cubanismo:
I understand the reasoning behind your disinterest in pursuing a CTS test, so thanks for giving me the chance to write one. I haven't accounted for your note about FreeBSD needing drmGetDevices() to properly enumerate the devices yet, but here's a first pass at the test as described:

https://github.com/cubanismo/VK-GL-CTS/tree/phys-device-drm

I don't have an implementation yet, but I "tested" this by sort of pretending there was an implementation in our driver (if that makes sense) and verifying I exercised the main code paths at least. I'll try to figure out the best way to integrate libdrm calls in CTS (I'm guessing dlopen/dlsym) and write up a real implementation next.

Joshua-Ashton:
> It would be a shame to drop multi-vendor specifications down to vendor-specific over the "paperwork,"

There is no reason that any vendor couldn't implement this.

emersion:
> here's a first pass at the test as described

Nice!

>I'm guessing dlopen/dlsym

Ugh. Why can't we just link normally to libdrm?

Might want to consider to skip the test on non-Linux platforms.

evelikov:
Hey everyone, have you seen [this upstream commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=cfb9b89f116a076ff333d4d0dc4aa862079481ba)?

If I understand it correctly, it basically says that the (DRM) device major/minor is non-deterministic and should not be relied upon. So having such information as part of the spec sounds moot. On the other hand a unique identifier like `ID_PATH_TAG` is fairly common and consistent.

Fwiw I'm aiming to find some time (in the next couple of days) and add `ID_PATH_TAG` to libdrm's `drmDevice` API - we had an implementation in Mesa for ages ;-)

emersion:
>If I understand it correctly, it basically says that the (DRM) device major/minor is non-deterministic and should not be relied upon

The former is correct: between two reboots or unplug/replug, the major/minor numbers can change. The latter is false: the major/minor can be relied upon to reliably identify a DRM device.

For this use-case, the major/minor is completely fine. Major/minor will also be used by the Wayland protocol. There's no need for all of the extra complexity of `ID_PATH_TAG`.

evelikov:
The document says "... chardev minor numbers are not recycled immediately ... " which does not sound like they are reliable. Perhaps they are seemingly reliable today due to reasons, but the goal is to not rely upon them.

Not sure what you mean with complexity - the implementation already uses either libudev, libdrm or both. So the extra code is essentially a function call.

emersion:
> The document says "... chardev minor numbers are not recycled immediately ... " which does not sound like they are reliable. 

It's the contrary. Chardev minor numbers are not recycled immediately, so unplugging card0 and re-plugging it doesn't re-use the same minor (card0), and instead uses another one (card1).

> Not sure what you mean with complexity - the implementation already uses either libudev, libdrm or both. So the extra code is essentially a function call.

Generating `ID_PATH_TAG` without libudev is non-trivial.

cubanismo:
I think it's fine if the major/minor numbers reported by a Vulkan device are only valid while that Vulkan device is. This is effectively true of all properties of Vulkan physical devices. If you hot-unplug them, their capabilities are suddenly invalid.

I've updated the proposed CTS test to use libdrm to look up the DRM device nodes. Haven't had a chance to test this yet, but wanted to get it out there to ensure I got the general form correct:

https://github.com/cubanismo/VK-GL-CTS/tree/phys-device-drm

emersion:
>I've updated the proposed CTS test to use libdrm to look up the DRM device nodes.

This looks fine to me. A few comments:

- Why does primary checks for `st_rdev` while render checks for `st_dev`?
- The test fails when libdrm is unavailable, we should probably skip instead.
- The loop on `DRM_NODE_MAX` needs to check `drmDevices[i]->available_nodes & (1 << j)`.

evelikov:
@cubanismo thanks. Wasn't aware of that (seemingly) device constant properties get invalidated.

cubanismo:
> This looks fine to me. A few comments:
> 
>     * Why does primary checks for `st_rdev` while render checks for `st_dev`? (both should check for `st_rdev`, right?)

Bug/typo. Will fix.

>     * The test fails when libdrm is unavailable, we should probably skip instead.

I'm unsure on this. It would make it easy to pass CTS on this extension by simply deleting libdrm from your test system. I thought maybe it was better to require anyone exposing the extension to have a sane DRM environment as well instead. Sure you can use DRM without libdrm.so, but it seems pretty corner-case. Users can always submit with patches if they find it too onerous for their environment.

>     * The loop on `DRM_NODE_MAX` needs to check `drmDevices[i]->available_nodes & (1 << j)`.

Good catch.

emersion:
> It would make it easy to pass CTS on this extension by simply deleting libdrm from your test system.

Ah, I see, makes sense.

cubanismo:
All suggestions should be addressed in the latest CTS, and we have a real implementation that passes the test now as well, so I've created a pull request for the test code:

https://github.com/KhronosGroup/VK-GL-CTS/pull/239

appendices/VK_EXT_physical_device_drm.txt
emersion: Done!

r-potter:
> > This should probably also have a feature structure
> Do we really need that? Can't we introduce it later on if we need to?

We just discussed this in the working group and would like to see it added. We now require this for all new extensions, and chose not to introduce a specific exemption for system integration extensions.

emersion:
What's the motivation for this "feature structure"?

r-potter:
There are a few overlapping points:

* Consistency with other extensions going forwards
* Due to code structure, the validation layers find them easier to work with
* Future proofing against an interaction with another proposal under discussion

The usual pattern for most (logical device) extensions is that enabling the extension itself does nothing, and enabling the feature is what turns on functionality. This extension interacts with physical devices rather than logical devices and needs to be usable prior to vkCreateDevice. That makes this a bit of an edge case, but we are trying to define as consistent an approach here as we reasonably can.

cubanismo:
Thanks for raising this with the working group Ralph. As it looks like the related discussion didn't directly address these types of extensions (Those only touching physical device properties), I've thrown it back to the working group asking for further clarification. Let's see what comes back from that. It's not that hard to add a struct, but I'm not yet clear how it would be used in practice.

r-potter:
The working group revisited this, and settled on an update to our requirements for authoring extensions that will exempt this extension from requiring a feature struct.

Essentially we switched from:

> every [physical-]device extension needs a feature enable struct

to:

> every [physical-]device extension needs a feature enable struct, unless (a) it doesn't add any device-level commands, or any structs or enum values used in device-level commands.

emersion:
Thanks for the udpate!

emersion:
Any news about this?

cubanismo:
Just waiting on CTS. I haven't seen any feedback there, and final review needs to happen internally apparently, so if there's no more feedback from non-Khronos members, I'll move the CTS over to internal review shortly.

emersion:
Ping

DadSchoorse:
Are there any news on when this might get merged?

stonesthrow:
@cubanismo , if the internal CTS is done, there is interest to move this forward this week.

Everyone make sure everything is done, all feedback has been addressed so we can close this out.

emersion:
Updated to add an explicit date in "Version History", which was previously reading "TBD".

stonesthrow:
@oddhack - approved.

appendices/VK_EXT_physical_device_drm.txt
```diff
@@ -0,0 +1,44 @@
...
+(e.g. if it's a software rendering engine without a display subsystem).
+
+To query DRM properties for a physical device, chain
+slink:VkPhysicalDeviceDrmPropertiesEXT to slink:VkPhysicalDeviceProperties2.
+
+=== References
+
+  . [[VK_EXT_physical_device_drm-fn1]]
+    https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_device_drm.txt[`EGL_EXT_device_drm`]
+
```
TomOlson, line 40: The extension overview should include the standard list of new types, entry points, enums, etc added by the extension. The information is autogenerated from the XML by the build chain, so all you need to do is insert this line here:

```suggestion
include::{generated}/interfaces/VK_EXT_physical_device_drm.txt[]
```

TomOlson (commented):
otherwise LGTM

appendices/VK_EXT_physical_device_drm.txt
emersion: Done!

TomOlson:
@oddhack WG has signed off, this is good to go in the next release assuming @emersion has no more changes

marty-johnson59:
@oddhack, @cubanismo, @emersion,  are we all good to go for releasing this MR on Monday (updated to TOT and approved to merge)?

cubanismo:
Yes, looks like it was merged, so it should be ready for release.

emersion:
Yes, +1 from me.

oddhack:
It is going out Monday.

PR #1379, opened 2020-10-12
Add missing handle comment for some VkObjectType values

DadSchoorse:
For most VkObjectType values there's a comment with the matching handle name. Since they are the most elegant way to get the object type for a handle, I plan to use these comments to generate parts of Wine's vulkan code.

This PR adds comments to VkObjectType values that previously didn't have any.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -9758,7 +9758,7 @@ typedef void <name>CAMetalLayer</name>;
             <enum extends="VkStructureType" extnumber="157" offset="3"          name="VK_STRUCTURE_TYPE_IMAGE_PLANE_MEMORY_REQUIREMENTS_INFO"/>
             <enum extends="VkStructureType" extnumber="157" offset="4"          name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SAMPLER_YCBCR_CONVERSION_FEATURES"/>
             <enum extends="VkStructureType" extnumber="157" offset="5"          name="VK_STRUCTURE_TYPE_SAMPLER_YCBCR_CONVERSION_IMAGE_FORMAT_PROPERTIES"/>
-            <enum extends="VkObjectType"    extnumber="157" offset="0"          name="VK_OBJECT_TYPE_SAMPLER_YCBCR_CONVERSION"/>
+            <enum extends="VkObjectType"    extnumber="157" offset="0"          name="VK_OBJECT_TYPE_SAMPLER_YCBCR_CONVERSION" comment="VkSamplerYcbcrConversion"/>
             <enum extends="VkFormat"        extnumber="157" offset="0"          name="VK_FORMAT_G8B8G8R8_422_UNORM"/>
             <enum extends="VkFormat"        extnumber="157" offset="1"          name="VK_FORMAT_B8G8R8G8_422_UNORM"/>
             <enum extends="VkFormat"        extnumber="157" offset="2"          name="VK_FORMAT_G8_B8_R8_3PLANE_420_UNORM"/>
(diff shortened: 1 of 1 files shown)


TomOlson:
@DadSchoorse, this seems very fragile to us - naming a field "comment" implies that it is ignorable (not consumed by anything but human eyes) and optional (nothing breaks if you leave it out). If code is going to depend on it we will need to add CI to make sure we don't forget it when we add new types, et cetera, and we'd rather not.

The [style guide](https://www.khronos.org/registry/vulkan/specs/1.2/styleguide.html) specifies a fixed mapping between object type names and enum values, and we guarantee to enforce it in new types. Probably obvious by inspection, but the rule is "delete OBJECT_TYPE_ and camel-case what's left, except capitalize the extension suffix if there is one". We do that in at least one internal script. The only slightly crufty bit is recognizing extension suffixes, but vk.xml contains a complete list. We recommend getting the type names that way.

oddhack:
As the XML maintainer / schema creator, agreed with @tom.olson. Comments can contain anything. While it has happened that we make an error in the naming scheme Tom describes, when we realize that we fix it by making the old, incorrect name an alias of the new, correct name.

DadSchoorse:
> The [style guide](https://www.khronos.org/registry/vulkan/specs/1.2/styleguide.html) specifies a fixed mapping between object type names and enum values, and we guarantee to enforce it in new types. Probably obvious by inspection, but the rule is "delete OBJECT_TYPE_ and camel-case what's left, except capitalize the extension suffix if there is one". We do that in at least one internal script. The only slightly crufty bit is recognizing extension suffixes, but vk.xml contains a complete list. We recommend getting the type names that way.

I considered this option, but it also seems fragile to me:
It needs the guarantee that no handle name contains any capitalized abbreviation.
Since struct names don't strictly follow this (e.g. `VkGeometryAABBNV`, `VkPhysicalDeviceTextureCompressionASTCHDRFeaturesEXT`) I concluded that future handle names might also contain completely capitalized parts.

TomOlson:
@DadSchoorse that's a fair point - there are structure type names that violate the style guide, and we should by our rules create aliases for them. I'll raise an internal issue.

It's still the case that aside from 'comment="Optional"' in some of the query types, there are no other places in vk.xml where there is information in comment attribute values that looks like it is meant to have machine-readable semantics, and I expect the WG will want to keep it that way. For object types, string conversion does work today. I'll ping the WG and see what they want to do.

expipiplus1:
There is an incomplete table generated in `Vulkan-Docs/gen/refpage/VkObjectType.txt`

oddhack:
After some internal discussion we're intending to add an explicit new attribute on the handle \<type> tags showing the corresponding OBJECT_TYPE token, and check the spelling consistency in CI. Also the table omission @expipiplus1  noted above will be fixed. Generating that table will be possible but we'll probably defer that part from the initial fixes. So this PR can probably be closed when we publish the XML updates in a couple of weeks.

oddhack:
This should be fixed in the 1.2.163 spec update. There is a new 'objtypeenum' attribute on handle types specifying the corresponding OBJECT_TYPE enum.

oddhack:
@DadSchoorse @SveSop @expipiplus1 please take a look at #1412. We don't want to just yank the comment= attributes out from under people's feet before they can switch over, but as there is a meaningful tagging scheme for this information now, we do want to remove them pretty soon - this is a chance for feedback.

expipiplus1:
Thanks, @oddhack, that's very considerate. I'm not using these comments myself so of course this is OK with me :)

DadSchoorse:
@oddhack I removed the code from Wine that use the comments.

PR #1396, opened 2020-11-16
pNext pointers not marked as optional in vk.xml

antlilja:
While generating language bindings from vk.xml I noticed that the pNext member in structs is not marked as optional, but from what i understand pNext should be allowed to be null.

Example from khronos.org: "pNext is NULL or a pointer to the next structure in a structure chain."
(https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBaseInStructure.html)

Maybe I'm misinterpreting what optional means in this context.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -652,11 +652,11 @@ typedef void <name>CAMetalLayer</name>;
             <comment>Struct types</comment>
         <type category="struct" name="VkBaseOutStructure">
             <member><type>VkStructureType</type> <name>sType</name></member>
-            <member>struct <type>VkBaseOutStructure</type>* <name>pNext</name></member>
+            <member optional="true">struct <type>VkBaseOutStructure</type>* <name>pNext</name></member>
         </type>
         <type category="struct" name="VkBaseInStructure">
             <member><type>VkStructureType</type> <name>sType</name></member>
-            <member>const struct <type>VkBaseInStructure</type>* <name>pNext</name></member>
+            <member optional="true">const struct <type>VkBaseInStructure</type>* <name>pNext</name></member>
         </type>
         <type category="struct" name="VkOffset2D">
             <member><type>int32_t</type>        <name>x</name></member>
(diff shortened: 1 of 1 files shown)


oddhack:
pNext is handled specially in the valid usage generator - it collects the structextends attriibutes of extension structures and determines what (if any) structures are allowed in the pNext chain based on that. How are you using the optional attribute in your bindings?

antlilja:
I was using optional to handle whether a pointer could be null or not.

oddhack:
For pNext, extension structures by definition may or may not be present. On the flip side, structures often start off being non-extensible, such that pNext must always be NULL, but extension structures introduced in the future can change that. , So bindings should allow for the pointer being null, or being valid, as a general matter.

antlilja:
But why is pNext not marked as optional then if it's allowed to be null? Maybe I've totally understood be use of optional?

oddhack:
@Tobski may have a different recollection but basically, what we use these annotations for is generating valid usage statements and code. Since the validation for pNext is inferred from other parts of the XML, the attribute is irrelevant to those uses, so we haven't had to think about it. Off the cuff it sounds sensible to set the option for use cases like your, but we'll need to think about it.

antlilja:
Ok thank you.

sfricke-samsung:
Going to quickly make sure this doesn't break any downstream Khronos tools!

sfricke-samsung (approved):
@oddhack confirmed this has no issues when used with the Validation Layers. I am assuming most upstream repos will no-op this since its just adding an attribute no one should be explicitly looking for currently

oddhack:
> @oddhack confirmed this has no issues when used with the Validation Layers. I am assuming most upstream repos will no-op this since its just adding an attribute no one should be explicitly looking for currently

Thanks! We should likely add something to the style guide about setting this attribute on pNext members but given Monday's discussion, we're good to merge this now (I will probably fix up some added white space in the XML later, too).

PR #1397, opened 2020-11-19
Reserve extension 364

crohdewm:


oddhack:
@crohdewm assuming you work for NV? If so, can you propose this as an internal Khronos gitlab MR? It's already out of sync with the internal repository's extension reservations as we've had a couple come in this week.

PR #1409, opened 2020-11-28
Remove extraneous newline from texel block size table

martty:
Small formatting typo I stumbled upon.


oddhack:
Thanks! Good catch.

PR #1412, opened 2020-12-03
Remove redundant / incomplete handle comments from VkObjectType enums

oddhack:
Followup to #1379. There is now a semantically meaningful schema
attribute mapping handles to corresponding VK_OBJECT_TYPES enum names.
This removes the comments which were previously being used by some
downstream consumers of the XML to infer this information on the basis
that (a) they are not defined to be usable for this purpose (b) they are
incomplete and (c) we have no intention of maintaining them in the
future when new handle types are defined.


oddhack:
We will leave this soak for about a week to get any feedback people may have, then merge if there are no issues.

oddhack:
No negative feedback, so proceeding.

PR #1422, opened 2020-12-31
Clarify docs for VK_FORMAT_D32_SFLOAT_S8_UINT

ROTARTSI82:
Make the description follow the other descriptions, as I was almost tricked into thinking it was a 32-bit format or even a 40-bit format.

Just a tiny change.

### DIFF
--- a/chapters/formats.txt
+++ b/chapters/formats.txt
@@ -483,7 +483,7 @@ endif::VK_EXT_4444_formats[]
   * ename:VK_FORMAT_D24_UNORM_S8_UINT specifies a two-component, 32-bit
     packed format that has 8 unsigned integer bits in the stencil component,
     and 24 unsigned normalized bits in the depth component.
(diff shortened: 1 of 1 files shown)


gnl21:
I don't think this is quite correct. The description says that there are 40 bits that are used and *optionally* 24 unused bits. The format compatibility section describes this format as "D32S8 (40 bit) Block size 5 bytes 1 texel/block". It also says that depth and stencil formats are opaque and so the actual storage may not match what you expect anyway, so implementations could use any size >= 40 bits. I would expect that the two most likely formats are 40 bit (as 2 planes, one of D32F followed by one of S8) and 64 bits (each texel as D32F_S8_X24).

With that in mind, is there something that you think could still be clarified?

ROTARTSI82:
Ah. Thanks for the explanation. Sorry about that.

PR #1425, opened 2021-01-06
Claim the extension number and add the new vendor.

mgorchak-blackberry:
Hi,

A couple of months ago I asked about possibility to add the new platform extension to add support for QNX. I went through the legal jungle and now we ready to sign CLA. I will submit all further changes using this corporate account.

Here is a small initial PR to claim the extension number and add the new vendor.

Thanks,


oddhack:
I'm going to merge this, but I might have to pull the actual interface descriptions (in the \<require> block) out later, depending on what downstream components do. The "disabled" attribute is supposed to prevent processing of these, but there are a lot of different projects consuming the XML at this point. Generally what we do is just reserve the extension number in the original MR and then have an MR which actually introduces and specifies all the APIs, but this will probably work OK.

PR #1439, opened 2021-01-27
Fix typo in ifdef/endif directives

cmarcelo:
Found when Vulkan-ValidationLayers code generation scripts complained about duplicated VUID in `validusage.json`.

CC: @oddhack @alan-baker


oddhack:
We already have an internal MR queued up that should fix this in the next spec update, so I suggest closing this. Thanks for catching it.

cmarcelo:
OK! Closing.

PR #1450, opened 2021-02-11
WIP: Add VK_EXT_acquire_wl_display

lubosz:
This is a continuation of the original PR #1001 

I rebased the patch on the current master branch and am willing to integrate changes from reviews, so this can land.

An overview of all related branches and merge requests to the stack can be found in the [README here](https://gitlab.freedesktop.org/lubosz/wlxr-pkgbuilds).

The only part which the Vulkan extension depends on is the Wayland protocol, currently proposed in a [new MR by Xaver Hugl](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/67).

Not sure how the procedure is with `Vulkan-Headers` and `Vulkan-Loader`, but I have branches for both enabling this extension:

https://github.com/lubosz/Vulkan-Loader/tree/acquire_wl_display
https://github.com/lubosz/Vulkan-Headers/tree/acquire_wl_display

An updated implementation of this extension in mesa can be found here:
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/8981

### DIFF
--- a/chapters/VK_EXT_acquire_wl_display/acquire_wl_display.txt
+++ b/chapters/VK_EXT_acquire_wl_display/acquire_wl_display.txt
@@ -1,0 +1,51 @@
+[open,refpage='vkAcquireWaylandDisplayEXT',desc='Acquire access to a VkDisplayKHR using Wayland',type='protos']
+--
+
+To acquire permission to directly access a display in Vulkan from a Wayland
+compositor, call:
+
+include::{generated}/api/protos/vkAcquireWaylandDisplayEXT.txt[]
+
+  * pname:physicalDevice is the handle to the physical device the display is on.
+  * pname:display A code:wl_display associated with the Wayland compositor that
+	the application wishes to acquire the display(s) from.
+  * pname:manager A zwp_drm_lease_manager_v1 connected to the Wayland compositor
+	that associated with the connectors.
+  * pname:pConnectorCount The number of valid pointers in the pname:pConnectors
+	array.
+  * pname:pConnectors Array of slink:VkWaylandLeaseConnectorEXT to be acquired.
+
+All permissions necessary to control the display are granted to the Vulkan
+instance associated with pname:physicalDevice until the display is released, the
+Wayland connection specified by pname:display is terminated, or the compositor
+revokes the lease.
+If the Wayland server declines to grant the lease or the granted lease does not
+include the requested connectors, the call must: return the error code
+ename:VK_ERROR_INITIALIZATION_FAILED.
+If the Wayland server later revokes the lease, requests using the
+acquired slink:VkDisplayKHR resources must: fail with
+ename:VK_ERROR_SURFACE_LOST_KHR.
+
+include::{generated}/validity/protos/vkAcquireWaylandDisplayEXT.txt[]
+--
+
+[open,refpage='VkWaylandLeaseConnectorEXT',desc='Resource for relating Wayland connectors to VkDisplayKHRs',type='structs']
+--
+
+The sname:VkWaylandLeaseConnectorEXT structure is defined as:
+
+include::{generated}/api/structs/VkWaylandLeaseConnectorEXT.txt[]
+
+  * pname:pConnector A Wayland zwp_drm_lease_connector_v1 to include in the
+	lease.
+  * pname:pDisplay The corresponding slink:VkDisplayKHR handle will be
+    returned here.
+
+The implementation will make a best-effort attempt to match each
+slink:VkDisplayKHR acquired from the Wayland compositor to the appropriate
+pname:pConnectorIn. If no match is found or the result would be ambiguous, the
+lease is terminated, and flink:vkAcquireWaylandDisplayEXT must: return
+ename:VK_ERROR_INITIALIZATION_FAILED.
+
+include::{generated}/validity/structs/VkWaylandLeaseConnectorEXT.txt[]
+--
--- a/chapters/VK_EXT_acquire_wl_display/acquire_wl_display.txt
+++ b/chapters/VK_EXT_acquire_wl_display/acquire_wl_display.txt
@@ -9,7 +9,7 @@ include::{generated}/api/protos/vkAcquireWaylandDisplayEXT.txt[]
   * pname:physicalDevice is the handle to the physical device the display is on.
   * pname:display A code:wl_display associated with the Wayland compositor that
 	the application wishes to acquire the display(s) from.
-  * pname:manager A zwp_drm_lease_manager_v1 connected to the Wayland compositor
+  * pname:leaseDevice A zwp_drm_lease_device_v1 connected to the Wayland compositor
 	that associated with the connectors.
   * pname:pConnectorCount The number of valid pointers in the pname:pConnectors
 	array.
--- a/appendices/VK_EXT_acquire_wl_display.txt
+++ b/appendices/VK_EXT_acquire_wl_display.txt
@@ -5,11 +5,12 @@
 include::{generated}/meta/{refprefix}VK_EXT_acquire_wl_display.txt[]
 
 *Last Modified Date*::
-    2019-10-17
+    2021-02-11
 *IP Status*::
     No known IP claims.
 *Contributors*::
   - Drew DeVault, Status Holdings, Ltd.
+  - Lubosz Sarnecki, Collabora, Ltd.
 
 This extension allows an application to take exclusive control of a display
 currently associated with a Wayland compositor.
@@ -43,3 +44,5 @@ None.
 
  * Revision 1, 2019-10-17 (Drew DeVault)
    - Initial draft
+ * Revision 2, 2021-02-11 (Lubosz Sarnecki)
+   - Update to Wayland drm-lease-unstable-v1 protocol changes
--- a/appendices/VK_EXT_acquire_wl_display.txt
+++ b/appendices/VK_EXT_acquire_wl_display.txt
@@ -1,0 +1,45 @@
+// Copyright 2018-2021 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_EXT_acquire_wl_display.txt[]
+
+*Last Modified Date*::
+    2019-10-17
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Drew DeVault, Status Holdings, Ltd.
+
+This extension allows an application to take exclusive control of a display
+currently associated with a Wayland compositor.
+When control is acquired, the display will be unassociated from the Wayland
+compositor until the application releases the display or the specified display
+connection is closed.
+Essentially, the Wayland compositor will behave as if the monitor has been
+unplugged until the application releases it.
+
+=== New Enum Constants
+
+None.
+
+=== New Enums
+
+None.
+
+=== New Structures
+
+  * slink:VkWaylandLeaseConnectorEXT
+
+=== New Functions
+
+  * flink:vkAcquireWaylandDisplayEXT
+
+=== Issues
+
+None.
+
+=== Version History
+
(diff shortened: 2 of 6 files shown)


emersion:
cc @cubanismo @rpavlik @jekstrand

swick:
After refreshing my mind I still don't understand why we don't want an instance extension (not a WSI extension) which just imports a drm (lease) fd. Can someone explain?

chapters/VK_EXT_acquire_wl_display/acquire_wl_display.txt
```diff
@@ -0,0 +1,51 @@
...
+
+[open,refpage='VkWaylandLeaseConnectorEXT',desc='Resource for relating Wayland connectors to VkDisplayKHRs',type='structs']
+--
+
+The sname:VkWaylandLeaseConnectorEXT structure is defined as:
+
+include::{generated}/api/structs/VkWaylandLeaseConnectorEXT.txt[]
+
+  * pname:pConnector A Wayland zwp_drm_lease_connector_v1 to include in the
+	lease.
```
rpavlik, line 40: looks like this needs reflow.py

rpavlik:
I'm not very qualified to review the meat of this proposal, but I can wordsmith and spec-edit it. From that POV it looks pretty good aside from needing a reflow.

Can someone refresh my memory about the zwp_ prefix?

Zamundaaa:
@rpavlik zwp means that it's a unstable wayland protocol.

@swick looking at the implementation in Mesa it enables the driver to listen for the lease to be revoked - however I don't know if that brings any actual benefits over just having the drm calls fail.
@ddevault could you comment on that? If the client already has to implement the wayland protocol itself then I don't think it would be too much to ask that they do it all and only pass the fd.
While I think in practice once wl-drm-lease-unstable has landed there won't be a new protocol for that anytime soon, it would IMO make sense to decouple this extension from the protocol. One other upside would be that we won't have to think about fast-tracking of the wayland protocol to stable or anything like that.

emersion:
Unstable Wayland protocols aren't allowed to make any breaking changes. They can get superseded by another protocol, but TBH stable protocols can get superseded too.

> fast-tracking of the wayland protocol to stable

I don't think this is a good idea. I don't want to make an exception for a stable protocol just because Khronos needs it.

Zamundaaa:
> Unstable Wayland protocols aren't allowed to make any breaking changes. They can get superseded by another protocol, but TBH stable protocols can get superseded too.

Thanks for clarifying that. Either way, if we only pass the fd (assuming there are no downsides other than clients having to do a little more work) then the extension is completely independent of any wayland protocol and if it ever gets replaced that won't matter to Vulkan.

>>fast-tracking of the wayland protocol to stable
>
>I don't think this is a good idea. I don't want to make an exception for a stable protocol just because Khronos needs it.

I agree that it's not good, that's why I mentioned it.

emersion:
FWIW, an extension taking a lease FD looks like a good idea to me. Can anyone from Khronos confirm this is an acceptable path forward?

rpavlik:
I can't speak for Vulkan, but it seems like a good idea to me: there's plenty of other extensions already that take an FD

cubanismo:
There's no such thing as a single Khronos view on a given extension design, especially multi-vendor/non-KHR extensions, but speaking as another driver vendor interested in supporting whatever comes of this effort:

Can someone pseudo-code up what the usage of a lease-FD-based extension would look like? I'm assuming this means:

1) App uses native wayland stuff to acquire the lease
2) App gives Vulkan the FD somehow

But I'm unclear where (2) would happen. I'm also unclear what happens if releasing of that lease is decoupled from the Vulkan API (I.e., do you still us vkReleaseDisplay() when using this mechanism, or just some native equivalent? If the latter, how does Vulkan interact with that?).

Given that's all probably resolveable though, this seems reasonable to me. My objections before were largely interlocking assumptions about the wayland mechanism and the Vulkan extension, which seemed unfortunate and brittle. It would be nice to have a more general "acquire unix display" thing that took an FD, and maybe an FD type (Like the other FD-handling Vulkan APIs) just for future extensibility. NV would in all likelihood implement the DRM leasing mechanism, as that's what wayland compositors would be using, but it's nice to keep options open at the Vulkan API level.

The only other concern I always keep in mind, and haven't thought through here, is supporting tunneling the existing Vulkan X11 extension through Xwayland using this mechanism. That still seems possible with the proposed rework, but as I say above, I'm not seeing the full picture yet and my Xwayland knowledge is limited.

swick:
My idea would be something like this

```
req = zwp_drm_lease_device_v1.create_lease_request
while connector = receive zwp_drm_lease_device_v1.connector:
    name = receive connector.name
    // or any other heuristic
    if name == CONNECTOR_NAME:
        req.request_connector(connector)
lease = req.submit()
fd = receive lease.lease_fd

vklease = {.fd = fd, .type = VK_UNIX_DISPLAY_FD_TYPE_DRM_LEASE }
num_displays = vkAcquireUnixDisplay(dev, vklease, NULL, 0)
displays[num_displays]
vkAcquireUnixDisplay(dev, vklease, displays, num_displays)
close(fd)

// use displays

for display in displays
    vkReleaseDisplay(dev, display)
```

> But I'm unclear where (2) would happen.

Is it any clearer now?

> I'm also unclear what happens if releasing of that lease is decoupled from the Vulkan API (I.e., do you still us vkReleaseDisplay() when using this mechanism, or just some native equivalent?

A lease can end any time already either by the compositor revoking access, by a vt switch or by a hot-unplug so even the client revoking the lease shouldn't be a problem. One way the vk implementation could work is by dup'ing the lease fd and closing it when all displays got released with vkReleaseDisplay.

cubanismo:
Interesting, I didn't realize a lease FD could include multiple displays. That does complicate the driver release logic a bit. Another alternative would just be to disallow use of vkReleaseDisplay with these leases. And yes, the X11 extension also had some language to deal with implicit lease revocation outside of Vulkan, so it's not a fundamentally new thing. It's just a little different to have multiple ways to explicitly revoke it.

Overall though, yes, the general workflow looks sane to me.

oddhack:
Marked 'WIP' in the title until ready to go.

emersion:
I'd suggest opening a separate PR for `VK_UNIX_DISPLAY_FD_TYPE_DRM_LEASE`, and closing this one when the new idea gets proven to be a good path forward. @lubosz, would you be willing to work on this?

Wallbraker:
@swick The extension works just like the X11 extension, if somebody cares strongly and has the time they can add the FD extension.

@emersion I'm afraid @lubosz doesn't have time to work on a completely new extension.

If nobody steps up and can promise to do the new extension within a resonable time frame we would very much like to move forward with this extension. This extension is proven to work in a full stack test, Monado running on the sway compositor.

emersion:
Yeah, the extensions aren't mutually exclusive. Bringing up a new extension could be a lot of work. This one seems to be in pretty good shape already.

swick:
There were issues raised in the original PR which are still not addressed.

Wallbraker:
> There were issues raised in the original PR which are still not addressed.

@swick Which are? The last comment from Drew in #1001 is that issues are resolved, so I don't know what issues you are referring to.

swick:
Of course he has. The whole reason why the X11 extension even exists is to make it independent of DRM. This extension is dependent on both wayland and DRM. So no, this extension doesn't work just like the X11 extension. Either get rid of the DRM dependency somehow or get rid of the wayland dependency (which would be straight forward and what I proposed).

emersion:
I don't understand how that's an issue. Yes, the extension depends on Wayland, because it's about the Wayland WSI. Yes, it depends on DRM, because that's about DRM leases. What's the problem with that?

swick:
The fact that it doesn't have to depend on both of them. This is strictly worse than a system with a dependency on just one of them. If there will be another non-DRM leasing type you'd have to add another extension. If there will be another mechanism to obtain a DRM lease you'd have to add another extension.

emersion:
Yes.

The people in this PR are only interested in DRM leases and Wayland. Thus this extension sounds reasonable to me.

The Wayland extension explicitly depends on DRM. It's in the name. It's only natural for a Vulkan extension relying on said Wayland extension to depend on both Wayland and DRM.

If you are interested in something else, nothing stops you from submitting another extension. But that doesn't mean the extension in this PR needs not to exist.

swick:
I shouldn't have to justify why less dependencies are better. You should have to just why the dependency is required or reasonable. This has not happened yet.

> If you are interested in something else, nothing stops you from submitting another extension. But that doesn't mean the extension in this PR needs not to exist.

If the other extension existed there would be literally no reason for this extension to exist. I get the feeling that people here are trying to push this just because it already exists and not because it's a good idea.

emersion:
> I shouldn't have to justify why less dependencies are better.

I'm not saying your proposal is a bad idea. 

> You should have to just why the dependency is required or reasonable. This has not happened yet.

I've done that: this Vulkan extension is about the Wayland protocol named drm-lease. Thus it depends on Wayland and DRM. Is this somehow not clear enough?

> If the other extension existed there would be literally no reason for this extension to exist. I get the feeling that people here are trying to push this just because it already exists and not because it's a good idea.

Then work on your extension. Open a PR, I'd welcome it.

Doesn't mean you need to shut down this one.

Now can we get back to a more constructive discussion please? :)

Wallbraker:
Sounds more that the protocol should instead take a `wl_output`, and leave the leash protocol to be a WSI detail?

emersion:
The compositor won't create `wl_output` globals for non-desktop connectors, and in general for connectors that aren't actively used in the desktop layout.

Wallbraker:
@swick I will admit that some of the push back to your suggestion is because we have a solution that works and no time at all to really work on something completely new.

But something that is bothering me with just doing a DRM fd based one. Is that why do we even need any new Vulkan extension at all? Since we now have a DRM fd why shouldn't we just use the DRM interface and the dmabuf external memory extension and do all of the display code ourselves? Basically why not go full DRM? My biggest argument to that is that now every single client has to rewrite that again and again.



> The compositor won't create `wl_output` globals for non-desktop connectors, and in general for connectors that aren't actively used in the desktop layout.

Ah hmm, looking at the protocol I guess it can be split up into two protocol. One wl_direct_mode one that only lists connectors (and maybe planes). And wl_drm_leash. This extension would only depend on wl_direct_mode and the WSI would be tasked with completely implementing wl_drm_leash.

emersion:
> Basically why not go full DRM?

It would work, on the drivers that implement the right extensions and support DRM format modifiers. That means only radv right now. More to come.

>I guess it can be split up into two protocol.

This sounds like introducing a lot of complexity in the protocol for no good reason other than "Vulkan wants it". I'd rather not do that.

Zamundaaa:
Going full DRM is an option but it should be considered that it's also a bunch of work for clients that Mesa already implements. With a DRM fd extension it's "get lease from compositor, X server or whatever, pass it to Vulkan, done".

A good reason for not having a DRM independent `wl_lease_output` protocol, aside from complexity, is that directly giving clients a non-master DRM fd enables them to gather all the information they would want to use, which could for example include DRM properties that don't yet exist or that we don't deem relevant.

cubanismo:
This issues I raised with the original extension were never addressed other than Drew claiming they weren't issues, which wasn't something I agreed with obviously.

emersion:
Can you elaborate on these issues? It's not immediately clear to me what these are given the lengthy discussion.

SidBlack (approved):
ok

cubanismo:
> Can you elaborate on these issues? It's not immediately clear to me what these are given the lengthy discussion. 

From above:

> My objections before were largely interlocking assumptions about the wayland mechanism and the Vulkan extension, which seemed unfortunate and brittle.

stonesthrow:
This PR has conflicts, its is still marked WIP, and its seems there are open comments to resolve.
Does https://github.com/KhronosGroup/Vulkan-Docs/pull/1001 need to remain open?

lubosz:
Superseded by !1529

PR #1470, opened 2021-03-11
Update to WSI Swapchain chapter regarding "release" and "present" terminology

### LINKED ISSUE #1419: WSI 33.9 needs more rigorous / distinct terms and definitions
I apologize that this is a long, detailed issue. However, I believe it has been a source of confusion for many people besides myself for several years, and it all revolves around a central issue, being that WSI 33.9 conflates distinct terms by using them interchangeably. I may very well just be misunderstanding something, so please correct me if you believe this to be the case.

I started by posting [this](https://www.reddit.com/r/vulkan/comments/kilgcs/ambiguities_in_vulkan_spec_on_the_definition_of/) post on Reddit yesterday. Feel free to read it in case I leave out any information.

I'll start with my understanding, mostly affirmed by the responses to [this](https://github.com/KhronosGroup/Vulkan-Docs/issues/117) previous issue. Please correct me if any of this is wrong:

I believe vkAcquireNextImageKHR does not immediately, meaning prior to its return, "acquire" an image from the swapchain, in the sense that it does not provide the application immediate ownership of it. Rather, when it returns, it is only to say that the application can be certain that, _eventually_, the image with the given index will be released by the PE from the swapchain. As a result, once vkAcquireNextImageKHR returns, it is safe to begin recording and submitting e.g. a render pass for it, so long as those rendering commands are synchronized with the _actual_ acquisition of the image via the signal semaphore provided to vkAcquireNextImageKHR.

Secondly, I believe that it is fully safe and valid to repeatedly (an arbitrarily large number of times) alternate between a vkAcquireNextImageKHR call and a corresponding vkQueuePresentKHR call (assuming there is something valid in the middle, like a render pass), regardless of the number of swapchain images that are available. In other words, once a swapchain image's _eventual_ (future) acquisition has been ascertained by the application via vkAcquireNextImageKHR's return, it is safe to enqueue render commands and a presentation, so long as those commands and the presentation are synchronized via the necessary semaphores, and then _immediately_ make another call to vkAcquireNextImageKHR, regardless of how many swapchain images are available. To explain it one more way (and I _do_ believe these are all equivalent statements), I believe it is fully possible and valid to have more frames in-flight than available swapchain images. After all, so long as the previous images have been enqueued for presentation, one can be certain that they will _eventually_ be presented, and when they are, other images will be released (to make room for the new presentations). Therefore, in such a case, vkAcquireNextImageKHR should be guaranteed to succeed, as it should _know_ that the image corresponding to the returned index will eventually be made available for rendering.

Lastly, I believe that it is _not_ valid to repeatedly (an arbitrarily large number of times) call vkAcquireNextImageKHR without making corresponding calls to vkQueuePresentKHR. Otherwise, you can enqueue yourself into a situation where you're trying to acquire an image from the PE when none will _ever_ be available, since none of them were _ever_ enqueued for presentation.

Assuming my understanding is correct, I'll move on to my confusion.

The documentation on vkAcquireNextImageKHR states the following:

> An image will eventually be acquired if the number of images that the application has currently acquired (but not yet presented) is less than or equal to the difference between the number of images in swapchain and the value of VkSurfaceCapabilitiesKHR::minImageCount. If the number of currently acquired images is greater than this, vkAcquireNextImageKHR should not be called; if it is, timeout must not be UINT64_MAX.

I interpret this as saying that each swapchain image must be "presented" before its index can ever be returned again by vkAcquireNextImageKHR, so if you acquire too many images before they can be "presented", then vkAcquireNextImageKHR can fail since you will have run out of images to acquire. The real problem, though, is that I interpret "presented" to mean, in some way, "presented by the presentation engine." I don't even believe there's a simple way to _know_ when an image has been presented by the presentation engine (vkQueuePresentKHR does not accept a signal fence or semaphore), so how can you possibly know when a call to vkAcquireNextImageKHR is valid or not, assuming that its validity depends on how many images have been acquired but not yet presented?

I believe my confusion is caused by a miscommunication in the docs. Particularly, I believe that when the docs say, "acquired (but not yet presented)", it _actually_ means "acquired their index via a call to vkAcquireNextImageKHR (but not yet enqueued for presentation via a call to vkQueuePresentKHR)". In fact, the sentence clearly states that the application is the one doing the "presenting," yet the application _can't_ present an image; that is not the application's role. All it can do is request that the PE eventually present the image.

Assuming I'm not missing anything, I think a general solution would be to introduce more rigorous / distinct terms and definitions in WSI 33.9.

For instance, it's probably confusing to say vkAcquireNextImageKHR "acquires" the image, or returns the index of the "acquired" image. Acquisition typically implies ownership, yet the docs clearly explain that the application doesn't actually have ownership of the image until the semaphore is signaled. In the earlier issue I mentioned, Ian Elliott stated that vkAcquireNextImageKHR waits for the image to become "available," and its semaphore is signaled when the image is "owned by the app." Perhaps this is better terminology, though surely there is an even better alternative. Is the image really "available" if the application can't actually start drawing to it yet without waiting on a semaphore? Probably not.

Another example: WSI 33.9 clearly states at one point, "Use of a presentable image must occur only after the image is returned by vkAcquireNextImageKHR, and before it is presented by vkQueuePresentKHR". But vkQueuePresentKHR _doesn't_ present an image. As mentioned, the application _can't_ present an image in any way; it can only request that the PE eventually present the image. Again, perhaps there should be a term for "enqueue for presentation" which is clearly distinct from "present", because, unless I'm gravely misunderstanding something, WSI 33.9 seems to use the two terms interchangeably despite them meaning very different things. Perhaps I should be inferring that when it uses "present" from the application's perspective, it means "enqueue for presentation," and when it says "present" from the PE's perspective, it actually means "present". But such an inference is not obvious, and it's entirely impossible when the term "presented" is used in passive voice without any indication of the subject.

Again, if my understanding is wrong, please feel free to correct me and close the issue :)

alexanderguyer:
Migration from https://github.com/alexanderguyer/Vulkan-Docs/pull/1

This pull request aims to resolve #1419

### DIFF
--- a/chapters/VK_KHR_swapchain/wsi.txt
+++ b/chapters/VK_KHR_swapchain/wsi.txt
@@ -45,7 +45,7 @@ The presentable images of a swapchain are owned by the presentation engine.
 An application can: acquire use of a presentable image from the presentation
 engine.
 Use of a presentable image must: occur only after the image is returned by
-fname:vkAcquireNextImageKHR, and before it is presented by
+fname:vkAcquireNextImageKHR, and before it is released by
 fname:vkQueuePresentKHR.
 This includes transitioning the image layout and rendering commands.
 
@@ -870,13 +870,7 @@ If pname:timeout is code:UINT64_MAX, the timeout period is treated as
 infinite, and fname:vkAcquireNextImageKHR will block until an image is
 acquired or an error occurs.
 
-An image will eventually be acquired if the number of images that the
-application has currently acquired (but not yet presented) is less than or
-equal to the difference between the number of images in pname:swapchain and
-the value of slink:VkSurfaceCapabilitiesKHR::pname:minImageCount.
-If the number of currently acquired images is greater than this,
-fname:vkAcquireNextImageKHR should: not be called; if it is, pname:timeout
-must: not be code:UINT64_MAX.
+fname:vkAcquireNextImageKHR should: not be called if the number of images that the application has currently acquired is greater than the difference between the number of images in pname:swapchain and the value of slink:VkSurfaceCapabilitiesKHR::pname:minImageCount.
 
 If an image is acquired successfully, fname:vkAcquireNextImageKHR must:
 either return ename:VK_SUCCESS, or ename:VK_SUBOPTIMAL_KHR if the swapchain
--- a/chapters/VK_KHR_swapchain/wsi.txt
+++ b/chapters/VK_KHR_swapchain/wsi.txt
@@ -870,7 +870,10 @@ If pname:timeout is code:UINT64_MAX, the timeout period is treated as
 infinite, and fname:vkAcquireNextImageKHR will block until an image is
 acquired or an error occurs.
 
-fname:vkAcquireNextImageKHR should: not be called if the number of images that the application has currently acquired is greater than the difference between the number of images in pname:swapchain and the value of slink:VkSurfaceCapabilitiesKHR::pname:minImageCount.
+fname:vkAcquireNextImageKHR should: not be called if the number of images
+that the application has currently acquired is greater than the difference
+between the number of images in pname:swapchain and the value of
+slink:VkSurfaceCapabilitiesKHR::pname:minImageCount.
 
 If an image is acquired successfully, fname:vkAcquireNextImageKHR must:
 either return ename:VK_SUCCESS, or ename:VK_SUBOPTIMAL_KHR if the swapchain


stonesthrow (approved):
Reviewed by Jeff and Keith of Khronos WSI TSG as a clarification improvement

stonesthrow:
Approved by Vulkan WSI TSG

Issue #1461, opened 2021-02-19
Clarification request on tessellation PatchVertices documentation


Venemo:
I'm reading the [spec about built-in variables](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap16.html#interfaces-builtin-variables) and specifically about `PatchVertices`.  
As far as I understand this is the equivalent of [gl_PatchVerticesIn in OpenGL](https://www.khronos.org/opengl/wiki/Built-in_Variable_(GLSL)#Tessellation_evaluation_shader_inputs).

## Is `PatchVertices` different or the same between TCS and TES?

The spec says:

> Decorating a variable with the `PatchVertices` built-in decoration will make that variable contain the number of vertices in the input patch being processed by the shader.

Can you please clarify what exactly *the number of vertices in the input patch being processed by the shader* means specifically for TCS and TES? Does it work like `gl_PatchVerticesIn` which in TES, is output patch size of the TCS?

I took a look at the CTS tests `dEQP-VK.tessellation.*` and I'm afraid there isn't a single one which uses `PatchVertices` in TES.

## How can a shader read patches of differing sizes?

I'd also appreciate some clarification about this:

> A single tessellation control or tessellation evaluation shader can read patches of differing sizes, so the value of the PatchVertices variable may differ between patches.

Does this mean that the same pipeline can contain patches of differing sizes? How is that possible? This would contradict what [the TCS spec says](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap10.html#shaders-tessellation-control). I think every input patch in the pipeline has the same size (according to the input patch size is set in `VkPipelineTessellationStateCreateInfo`), and also every output patch in the pipeline has the same size (set by the `OpExecutionMode OutputVertices` in either TCS or TES).

gnl21:
> Is PatchVertices different or the same between TCS and TES?

My understanding is that yes, the `PatchVertices` variable is the equivalent of the GLSL builtin `gl_PatchVerticesIn`. I think "input patch" is referring to the input of that shader, rather than the whole pipeline, so in the TES the size of the input patch is the same as that output by the TCS. I agree that the spec text could be a bit clearer about that.

> I took a look at the CTS tests dEQP-VK.tessellation.* and I'm afraid there isn't a single one which uses PatchVertices in TES.

I'll create a CTS issue to look into whether there is any coverage of this and to create some if there is not.

> How can a shader read patches of differing sizes?

[...]

PR #1475, opened 2021-03-16
Clarify PatchVertices documentation for tessellation shaders.

Venemo:
I attempted to make the `PatchVertices` documentation clearer. See #1461 for the discussion that motivated this change.

### DIFF
--- a/chapters/interfaces.txt
+++ b/chapters/interfaces.txt
@@ -2752,9 +2752,12 @@ code:PatchVertices::
 Decorating a variable with the code:PatchVertices built-in decoration will
 make that variable contain the number of vertices in the input patch being
 processed by the shader.
-A single tessellation control or tessellation evaluation shader can: read
-patches of differing sizes, so the value of the code:PatchVertices variable
-may: differ between patches.
+In a Tessellation Control Shader, this is the same as the name:patchControlPoints
+member of slink:VkPipelineTessellationStateCreateInfo.
+In a Tessellation Evaluation Shader, code:PatchVertices is equal to the
+tessellation control output patch size.
+When the same shader is used in different pipelines where the patch sizes
+are configured differently, the value of the code:PatchVertices variable will also differ.
 
 .Valid Usage
 ****


Venemo:
I already signed the CLA but it seems that it's still "Pending" here...

oddhack:
@outofcontrol could you take a look at the CLAbot for this PR? It's saying "Timur Kristóf seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please add the email address used for this commit to your account."

@Venemo in any event, it usually takes us a week or two to get the Vulkan WG to sign off on PRs, so this is unlikely to hold things up any further.

Khronoswebmaster:
@Venemo Is it possible you committed this PR with a different email address from the one used on your GitHub account? This is the most common issue we see for this error. More [here](https://docs.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user).

chapters/interfaces.txt
```diff
@@ -2752,9 +2752,16 @@ code:PatchVertices::
 Decorating a variable with the code:PatchVertices built-in decoration will
 make that variable contain the number of vertices in the input patch being
 processed by the shader.
-A single tessellation control or tessellation evaluation shader can: read
+In a Tessellation Control Shader, this is the same as the name:patchControlPoints
+member of slink:VkPipelineTessellationStateCreateInfo.
+In a Tessellation Evaluation Shader, code:PatchVertices is equivalent to the
```
gnl21, line 2757:
```suggestion
In a Tessellation Evaluation Shader, code:PatchVertices is equal to the
```

Just a minor quibble, but I'd prefer to say "equal" here to emphasize that the value is the same but they're not "equivalent" because they have different uses. You can't substitute one for the other.

chapters/interfaces.txt
```diff
@@ -2752,9 +2752,16 @@ code:PatchVertices::
...
+tessellation control output patch size, which is controlled by the code:OpExecutionMode
+code:OutputVertices (specified in either tessellation shader).
```
gnl21, line 2759:
```suggestion
tessellation control output patch size.
```

I don't think it really helps with understanding here to say how these things are specified, and the more places we say things like that, the more places we have to keep consistent, etc, so I'd rather leave this last clause out.

chapters/interfaces.txt
```diff
@@ -2752,9 +2752,16 @@ code:PatchVertices::
...
+When the same tessellation control or tessellation evaluation shader is used
+in different pipelines, the input and/or output patch sizes may: be configured
+differently. In this case, the same tessellation control or evaluation shader can: read
 patches of differing sizes, so the value of the code:PatchVertices variable
-may: differ between patches.
+may: differ between pipelines.
```
gnl21, line 2764: This seems a little bit repetitive now. Maybe it would be sufficient to say:
> The same tessellation control or tessellation evaluation shader may be used in different pipelines where the patches are of differing sizes, so the value of the code:PatchVertices variable may differ.

(Github won't let me include that as a suggestion because it spans a deleted line.)

gnl21 (commented):
Everything that's here looks correct to me, and I'd be OK taking this as it is. I've made a few editorial suggestions that I think would improve it. What do you think, @Venemo?

chapters/interfaces.txt
Venemo: Shouldn't we then just say `code::PatchVertices is the same as the tessellation control output patch size`?

chapters/interfaces.txt
Venemo: Ok, let's leave it out.

chapters/interfaces.txt
Venemo: Agreed, let's get rid of the repetitiveness. I think this might be even more concise:

> When the same shader is used in different pipelines where the patch sizes are configured differently, the value of the code:PatchVertices variable will also differ.

How does this sound?

chapters/interfaces.txt
Venemo: On a second thought, I think "is equal to" sounds better.

gnl21: They're not the same thing, but they have the same value, so I'm not sure saying that they are "the same" is clearer. I think saying that they're equal probably conveys that best.

Venemo: Agreed.

chapters/interfaces.txt
gnl21: Sounds good to me. Thanks.

Venemo: Done.

Venemo:
@Khronoswebmaster Thanks, that was exactly the problem. I forgot that I use a different email address for my GitHub login.

gnl21 (approved):

Khronoswebmaster:
> @Khronoswebmaster Thanks, that was exactly the problem. I forgot that I use a different email address for my GitHub login.

Happy to see it was a simple fix.

Venemo:
@gnl21 What is the process of merging this commit? I see that you approved it but I don't have a merge button here.

gnl21:
@oddhack will merge this when he's ready to make a spec update with it in. (Because a lot of the content of this repo gets brought across from Khronos's internal repo, it's easier to handle it this way). If I reviewed it in time then that will probably be sometime around this weekend.

Venemo:
Awesome, thank you!

SidBlack (commented):
ok

PR #1503, opened 2021-04-16
Split errors chapter

### LINKED ISSUE #1499: [enhancement] Errors chapter should be split
The Errors chapter should be split into valid usage, and return codes. I always look for error codes and then have to scroll over bunch of structure type definitions. Plus the chapter name is not really descriptive.

krOoze:
close #1499

### DIFF
--- a/chapters/fundamentals.txt
+++ b/chapters/fundamentals.txt
@@ -718,18 +718,15 @@ The implicit parameters and their associated object are listed below.
 include::{generated}/hostsynctable/implicit.txt[]
 
 
+[[fundamentals-validusage]]
+== Valid Usage
+
+// preserving old id for permalinking
 [[fundamentals-errors]]
-== Errors
-
-Vulkan is a layered API.
-The lowest layer is the core Vulkan layer, as defined by this Specification.
-The application can: use additional layers above the core for debugging,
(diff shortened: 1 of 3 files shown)


chapters/fundamentals.txt
```diff
@@ -1649,3 +1629,30 @@ include::{generated}/api/structs/VkRect2D.txt[]
...
 --
+
+
+// VkStructureType is pretty long,
+// so keep it as a last chapter section for readability.
+=== Structure Types
+
+[open,refpage='VkStructureType',desc='Vulkan structure types (pname:sType)',type='enums']
+--
+Each enumerant of elink:VkStructureType corresponds to a particular
```
oddhack, line 1640: Please change to "Each enumerant corresponds to a specific structure type, whose pname:sType member contains the enumerant value." People had a hard time understanding this sentence.

With that exception I think people are generally OK with this restructuring, though they want a little longer for review. Thanks.

krOoze: IDK, there's a significant risk that would lead to more wordsmithing. I think I am just gonna revert it to the original.

Though it somewhat bothers me it says "value" instead of "enumerant".

PR #1517, opened 2021-04-28
Add a new VK_EXT_calibrated_timestamps extension

jekstrand:
Intel hardware doesn't necessarily guarantee a single time domain across the entire device for `vkCmdWriteTimestamp()`.  This hasn't been a problem for us because we ran everything on the 3D queue.  However, with things like Vulkan video up-and-coming as well as additional compute queues, this is turning into a real issue.  This tiny little extension allows the client to get calibrated timestamps on a per-queue basis rather than just per-device.

### DIFF
--- a/appendices/VK_EXT_calibrated_queue_timestamps.txt
+++ b/appendices/VK_EXT_calibrated_queue_timestamps.txt
@@ -1,0 +1,29 @@
+// Copyright 2018-2021 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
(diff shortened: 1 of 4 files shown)


polarina:
If I may ask, have you considered calibrated timestamps on a per queue family basis rather than per queue?

I don't know much about hardware specifics myself. It seems more logical at first glance as I look over the proposed specification changes.

jekstrand:
I did consider that.  The problem is that you may have multiple hardware queues in any given queue family and they may not share a clock.  Of course, in such a scenario, this implies that each `VkQueue` gets pinned to one particular HW queue and can't be migrated without breaking the timestamp.  But, depending on the application use-case, that may be better than not having calibrated timestamps at all.

PR #1525, opened 2021-05-07
Add VK_EXT_wl_drm_lease_display

bl4ckb0ne:
Hi, this pull request is the continuation of the work started by Drew Devault and by Lubosz Sarnecki and superseded both #1001 and #1450

I rebased the patch against `main` and upgraded the protocol [`drm_lease_v1`][1] by Xaver Hugl to the latest version (staging). I will go through the issues that were raised on #1001 and #1450 to get the merge request in a good shape.

I will also take care of the Mesa [2] [3] and Monado [4] [5] implementation, I'll update the status of the merge request in time.

This is my first contribution to a Vulkan extensions, your patience will be greatly appreciated.

[1]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/67 
[2]: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/1509
[3]: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/8981
[4]: https://gitlab.freedesktop.org/monado/monado/-/merge_requests/141
[5]: https://gitlab.freedesktop.org/monado/monado/-/merge_requests/683

### DIFF
--- a/appendices/VK_EXT_wl_drm_lease_display.txt
+++ b/appendices/VK_EXT_wl_drm_lease_display.txt
@@ -1,0 +1,53 @@
+// Copyright 2018-2021 The Khronos Group, Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_EXT_wl_drm_lease_display.txt[]
+
+*Last Modified Date*::
+    2021-05-07
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Drew DeVault, Status Holdings, Ltd.
+  - Lubosz Sarnecki, Collabora, Ltd.
+  - Simon Zeni, Status Holdings, Ltd.
+
+This extension allows an application to take exclusive control of a display
+currently leased by a Wayland compositor.
+When control is acquired, the display will be unassociated from the Wayland
+compositor until the application releases the display or the specified display
+connection is closed.
+Essentially, the Wayland compositor will behave as if the monitor has been
+unplugged until the application releases it.
+
+=== New Enum Constants
+
+None.
+
+=== New Enums
+
+None.
+
+=== New Structures
+
+  * slink:VkWaylandDrmLeaseConnectorEXT
+
+=== New Functions
+
+  * flink:vkWaylandDrmLeaseDisplayEXT
+
+=== Issues
+
+None.
+
+=== Version History
+
+ * Revision 1, 2019-10-17 (Drew DeVault)
+   - Initial draft
+ * Revision 2, 2021-02-11 (Lubosz Sarnecki)
+   - Update to Wayland drm-lease-unstable-v1 protocol changes
+ * Revision 3, 2021-05-06 (Simon Zeni)
+   - Update to Wayland drm-lease-v1 protocol changes
+ * Revision 4, 2021-05-07 (Simon Zeni)
+   - Change name from VK_EXT_acquire_wl_display to VK_EXT_wl_drm_lease_display
--- a/chapters/VK_EXT_wl_drm_lease_display/wl_drm_lease_display.txt
+++ b/chapters/VK_EXT_wl_drm_lease_display/wl_drm_lease_display.txt
@@ -1,0 +1,55 @@
+// Copyright 2018-2021 The Khronos Group, Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+[open,refpage='vkWaylandDrmLeaseDisplayEXT',desc='Gives access to a VkDisplayKHR using Wayland DRM lease',type='protos']
+--
+
+To acquire permission to directly lease a display in Vulkan from a Wayland
+compositor, call:
+
+include::{generated}/api/protos/vkWaylandDrmLeaseDisplayEXT.txt[]
+
+  * pname:physicalDevice is the handle to the physical device the display is on.
+  * pname:display A code:wl_display associated with the Wayland compositor that
+    the application wishes to acquire the display(s) from.
+  * pname:leaseDevice A wp_drm_lease_device_v1 connected to the Wayland compositor
+    that associated with the connectors.
+  * pname:pConnectorCount The number of valid pointers in the pname:pConnectors
+    array.
+  * pname:pConnectors Array of slink:VkWaylandDrmLeaseConnectorEXT to be acquired.
+
+All permissions necessary to control the display are granted to the Vulkan
+instance associated with pname:physicalDevice until the display is released, the
+Wayland connection specified by pname:display is terminated, or the compositor
+revokes the lease.
+If the Wayland server declines to grant the lease or the granted lease does not
+include the requested connectors, the call must: return the error code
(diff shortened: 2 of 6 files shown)


appendices/VK_EXT_wl_drm_lease_display.txt
```diff
@@ -0,0 +1,53 @@
+// Copyright 2018-2021 The Khronos Group Inc.
```
oddhack, line 1: So you're welcome to assign copyright to Khronos if you want, but that's not required. All we need is that it be under CC-BY 4.0 and that the CLA is executed on the PR.

chapters/VK_EXT_wl_drm_lease_display/wl_drm_lease_display.txt
```diff
@@ -0,0 +1,51 @@
...
+The sname:VkWaylandDrmLeaseConnectorEXT structure is defined as:
+
+include::{generated}/api/structs/VkWaylandDrmLeaseConnectorEXT.txt[]
+
+  * pname:pConnector A Wayland wp_drm_lease_connector_v1 to include in the
+    lease.
+  * pname:pDisplay The corresponding slink:VkDisplayKHR handle will be
+    returned here.
+
+The implementation will make a best-effort attempt to match each
```
oddhack, line 44: This language should probably be moved up to the vkWaylandDrmLeaseDisplayEXT section.

oddhack:
It would be great if you can get the CI script working, though by manual inspection this looks fine. I *think* all that should be required is to enable Github Actions on the repo you're staging this PR from, and push to the branch or perhaps manually trigger CI on it, though I'm not all that knowledgeable about Actions.

chapters/VK_EXT_wl_drm_lease_display/wl_drm_lease_display.txt
bl4ckb0ne: Moved above

bl4ckb0ne:
@oddhack CI is passing and the commit is rebased on top of `main`, I'm will update the status to ready for review.

cubanismo:
This doesn't seem to address my comments on either of the two prior instances of this series. I still don't see a good reason to tie this to wayland specifically. It seems like at this point the API would be better served by a simple lease FD import, perhaps supplemented by a correlation API in a separate extension at some point, leaving the protocol out of Vulkan to the extent possible.

The only reasonable rebuttal to this I've seen thus far seems to boil down to "We've already gone too far down this road, we need to ship," which given the rate at which this has been actually pushed, not to mention the small amount of code required to implement either direction, doesn't seem like a real concern. The only other reason I could see to pursue a solution that puts the protocol inside the driver is to abstract which protocol and permissions transfer mechanism would be used. That's why we went that direction for the X11 version, and when the wayland protocol was still experimental, that would have made sense *if the extension didn't directly re-use the wayland protocol structs anyway*, which is why I originally pushed to divorce the two APIs a bit more in the other direction (Hide more in the driver, not less) but I see no valid reason for this with the present design and solidified Wayland protocol. What we have here is seemingly the worst of both worlds.

Is anyone actually opposed to an FD-based import direction from a design point of view? The design benefits seem very clear to me.

bl4ckb0ne:
The FD-based extension would miss the point, the specific use case here is VR. Not any kind of display. I feel it would be too complex to make a generic FD extension that satisfies everybody (compositors, VR, etc).

The wayland compositor is the DRM master, it needs to agree to lease some of its display, it needs to be involved.

cubanismo:
Thanks for chiming in Drew. While you're engaged, just want to ask: Do you still want to be listed as the primary contact for this extension? Fine by me either way, just wanted to check since I noticed it while reviewing and understood you hadn't been actively involved for a while.

> This mirrors the approach used for X11 with the Xrandr leasing extension,. so there is prior art. There are other parts of VkDisplay which incorporate the Wayland display directly into Vulkan, so this design is not unusual.

I explained why the design considerations for that extension were different above: Basically NVIDIA wanted to be able to use our internal KMS API with X11, so we needed more abstraction. This extension likely ends up with "DRM" in the name either way, so that's not going to be a consideration here.

> An fd-based approach would also work, and I understand your comments in favor of it. I'm not opposed to this, but I think that it's not any more or less effective than this design, and this design more closely mirrors the code that a Vk client may already have in place (or will add alongside) for X11 support.

I don't think it's going to be a huge delta in client code either way (not that Vulkan has historically raised minimal lines of client code over other design principles) and the FD-based approach makes the extension more flexible. We've already heard from a few customers who want to do their own custom thing with DRM leases. We'll just end up defining the more flexible version soon if we can't agree on it now, making the wayland-specific one redundant functionality we'll have to support forever.

ddevault:
>Do you still want to be listed as the primary contact for this extension?

Yeah, I should still be listed here. I'm not participating directly but I am still overseeing the work (@bl4ckb0ne is working on sourcehut's behalf).

>I explained why the design considerations for that extension were different above: Basically NVIDIA wanted to be able to use our internal KMS API with X11, so we needed more abstraction. This extension likely ends up with "DRM" in the name either way, so that's not going to be a consideration here.

I understand that this is not going to be an issue here, but just to be clear, I have very little patience for listening to Nvidia's concerns regarding their private driver implementation. It's just that: private. If you want a spot on the Linux graphics table, then please join us on our terms, not yours. This extension is designed with the open source graphics stack in mind.

>I don't think it's going to be a huge delta in client code either way (not that Vulkan has historically raised minimal lines of client code over other design principles) and the FD-based approach makes the extension more flexible. We've already heard from a few customers who want to do their own custom thing with DRM leases. We'll just end up defining the more flexible version soon if we can't agree on it now, making the wayland-specific one redundant functionality we'll have to support forever.

Yeah, I understand. I think that a FD based approach would necessitate the simultaneous deprecation of the Xrandr vk extension, which increases the scope of the proposal, discussion, and bikeshedding potential considerably. If you think we should push for that, then I'm not entirely opposed to it, but I'm also not opposed to moving forward with this proposal now and deprecating it alongside the Xrandr extension if and when a more general proposal comes to the table.

Also, can you speak a bit more to the specific scenario your clients are facing when they will need more control over DRM leases?

cubanismo:
> I think that a FD based approach would necessitate the simultaneous deprecation of the Xrandr vk extension

Why is that? Are you saying it wouldn't be technically feasible to implement the Xrandr VK extension on top of the FD-based import mechanism + wayland protocol? If so, why?

> Also, can you speak a bit more to the specific scenario your clients are facing when they will need more control over DRM leases?

Consider a bundle of Vulkan renderer processes that each want to control individual displays on a single GPU directly, much like the VR use case but with no "desktop." A master process would hold the DRM primary status and hand out leases to the child processes. There's no need for composition, so there's no need for a wayland compositor or X11 instance in this scenario.

cubanismo:
> No - I'm saying more that the X-based system becomes obsolete once an FD-based mechanism is available.

OK, understood. I still think the FD-based approach is the right design direction. I'm all for generalization.

PR #1529, opened 2021-05-11
Add VK_EXT_acquire_drm_display

bl4ckb0ne:
This pull request is a continuation of the work done in https://github.com/KhronosGroup/Vulkan-Docs/pull/1525, where it was discussed that instead of having a Wayland-centered approach to DRM leasing to handle VR headset (see  #1001 and #1450), we would change the design to have a more generic FD-based approach.

[Mesa implementation](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/11014)
[Monado usage](https://gitlab.freedesktop.org/monado/monado/-/merge_requests/832)

### DIFF
--- a/appendices/VK_EXT_acquire_drm_display.txt
+++ b/appendices/VK_EXT_acquire_drm_display.txt
@@ -1,0 +1,32 @@
+// Copyright 2018-2021 The Khronos Group, Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_EXT_acquire_drm_display.txt[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2021-06-09
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Simon Zeni, Status Holdings, Ltd.
+
+=== Description
+
+This extension allows an application to take exclusive control of a display
+using the Direct Rendering Manager (DRM) interface. When acquired, the display
+will be under full control of the application until the display is either
+released or the connector is unplugged.
+
+include::{generated}/interfaces/VK_EXT_acquire_drm_display.txt[]
+
+=== Issues
+
+None.
+
+=== Version History
+
+* Revision 1, 2021-05-11 (Simon Zeni)
+    - Initial draft
--- a/chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
+++ b/chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
@@ -1,0 +1,60 @@
+// Copyright 2018-2021 The Khronos Group, Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+[open,refpage='vkAcquireDrmDisplayEXT',desc='Acquire access to a VkDisplayKHR using DRM',type='protos']
+--
+
+To acquire permission to directly a display in Vulkan from the Direct Rendering
+Manager (DRM) interface, call:
+
+include::{generated}/api/protos/vkAcquireDrmDisplayEXT.txt[]
+
+  * pname:physicalDevice The physical device the display is on.
+  * pname:drmFd DRM primary file descriptor.
+  * pname:display The display the caller wishes Vulkan to control.
+
+All permissions necessary to control the display are granted to the Vulkan
+instance associated with the provided pname:physicalDevice until the display is
+either released or the connector is unplugged.
+The provided pname:drmFd must correspond to the one owned by the
+pname:physicalDevice. If not, the error code ename:VK_ERROR_UNKNOWN must be
+returned. The DRM FD must have DRM master permissions.
+If any error is encountered during the acquisition of the display, the call
+must return the error code ename:VK_ERROR_INITIALIZATION_FAILED.
+
+The provided DRM fd should not be closed before the display is released,
+attempting to do it may result in undefined: behaviour.
+
+include::{generated}/validity/protos/vkAcquireDrmDisplayEXT.txt[]
+--
+
+[open,refpage='vkGetDrmDisplayEXT',desc='Query the VkDisplayKHR corresponding to a DRM connector ID',type='protos']
+--
+
+Before acquiring a display from the DRM interface, the caller may want to
+select a specific sname:VkDisplayKHR handle by identifying it using a
+pname:connectorId. To do so, call:
+
+include::{generated}/api/protos/vkGetDrmDisplayEXT.txt[]
+
+  * pname:physicalDevice The physical device to query the display from.
+  * pname:drmFd DRM primary file descriptor.
+  * pname:connectorId Identifier of the specified DRM connector.
+  * pname:display The corresponding slink:VkDisplayKHR handle will be returned
+    here.
+
+If there is no slink:VkDisplayKHR corresponding to the pname:connectorId on
+the pname:physicalDevice, the returning pname:display must be set to
+dlink:VK_NULL_HANDLE.
+The provided pname:drmFd must correspond to the one owned by the
+pname:physicalDevice. If not, the error code ename:VK_ERROR_UNKNOWN must be
+returned. Master permissions are not required, because the file descriptor is
+just used for information gathering purposes.
+The given pname:connectorId must be a resource owned by the provided
+pname:drmFd. If not, the error code ename:VK_ERROR_UNKNOWN must be returned.
+If any error is encountered during the identification of the display, the call
+must return the error code ename:VK_ERROR_INITIALIZATION_FAILED.
+
+include::{generated}/validity/protos/vkGetDrmDisplayEXT.txt[]
+--
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -10450,6 +10450,19 @@ typedef void <name>CAMetalLayer</name>;
             <param><type>VkCommandBuffer</type> <name>commandBuffer</name></param>
             <param>const <type>VkCuLaunchInfoNVX</type>* <name>pLaunchInfo</name></param>
         </command>
+        <command successcodes="VK_SUCCESS" errorcodes="VK_ERROR_INITIALIZATION_FAILED">
+            <proto><type>VkResult</type> <name>vkAcquireDrmDisplayEXT</name></proto>
+            <param><type>VkPhysicalDevice</type> <name>physicalDevice</name></param>
+            <param><type>int32_t</type> <name>drmFd</name></param>
+            <param><type>VkDisplayKHR</type> <name>display</name></param>
+        </command>
+        <command successcodes="VK_SUCCESS" errorcodes="VK_ERROR_INITIALIZATION_FAILED,VK_ERROR_OUT_OF_HOST_MEMORY">
+            <proto><type>VkResult</type> <name>vkGetDrmDisplayEXT</name></proto>
+            <param><type>VkPhysicalDevice</type> <name>physicalDevice</name></param>
+            <param><type>int32_t</type> <name>drmFd</name></param>
+            <param><type>uint32_t</type> <name>connectorId</name></param>
+            <param><type>VkDisplayKHR</type>* <name>display</name></param>
+        </command>
     </commands>
 
     <feature api="vulkan" name="VK_VERSION_1_0" number="1.0" comment="Vulkan core API interface definitions">
@@ -15219,10 +15232,12 @@ typedef void <name>CAMetalLayer</name>;
                 <type name="PFN_vkDeviceMemoryReportCallbackEXT"/>
             </require>
         </extension>
-        <extension name="VK_EXT_extension_286" number="286" type="instance" author="EXT" contact="Drew DeVault sir@cmpwn.com" supported="disabled">
+        <extension name="VK_EXT_acquire_drm_display" number="286" type="instance" requires="VK_EXT_direct_mode_display" author="EXT" contact="Drew DeVault sir@cmpwn.com" supported="vulkan">
             <require>
-                <enum value="0"                                             name="VK_EXT_EXTENSION_286_SPEC_VERSION"/>
-                <enum value="&quot;VK_EXT_extension_286&quot;"              name="VK_EXT_extension_286"/>
+                <enum value="1"                                             name="VK_EXT_ACQUIRE_DRM_DISPLAY_SPEC_VERSION"/>
+                <enum value="&quot;VK_EXT_acquire_drm_display&quot;"        name="VK_EXT_ACQUIRE_DRM_DISPLAY_EXTENSION_NAME"/>
+                <command name="vkAcquireDrmDisplayEXT"/>
+                <command name="vkGetDrmDisplayEXT"/>
             </require>
         </extension>
         <extension name="VK_EXT_robustness2" number="287"  type="device" author="EXT" contact="Liam Middlebrook @liam-middlebrook" supported="vulkan">
--- a/chapters/VK_EXT_direct_mode_display/acquire_release_displays.txt
+++ b/chapters/VK_EXT_direct_mode_display/acquire_release_displays.txt
@@ -18,6 +18,10 @@ ifdef::VK_NV_acquire_winrt_display[]
 include::{chapters}/VK_NV_acquire_winrt_display/acquire_winrt_display.txt[]
 endif::VK_NV_acquire_winrt_display[]
 
+ifdef::VK_EXT_acquire_drm_display[]
+include::{chapters}/VK_EXT_acquire_drm_display/acquire_drm_display.txt[]
+endif::VK_EXT_acquire_drm_display[]
+
 [open,refpage='vkReleaseDisplayEXT',desc='Release access to an acquired VkDisplayKHR',type='protos']
 --
 To release a previously acquired display, call:
--- a/include/vulkan/vulkan.h
+++ b/include/vulkan/vulkan.h
@@ -85,7 +85,6 @@
 #include "vulkan_screen.h"
 #endif
 
-
 #ifdef VK_ENABLE_BETA_EXTENSIONS
 #include "vulkan_beta.h"
 #endif


xml/vk.xml
```diff
@@ -34,6 +34,7 @@ branch of the member gitlab server.
         <platform name="ggp" protect="VK_USE_PLATFORM_GGP" comment="Google Games Platform"/>
         <platform name="provisional" protect="VK_ENABLE_BETA_EXTENSIONS" comment="Enable declarations for beta/provisional extensions"/>
         <platform name="screen" protect="VK_USE_PLATFORM_SCREEN_QNX" comment="QNX Screen Graphics Subsystem"/>
+        <platform name="drm" protect="VK_USE_PLATFORM_DRM_KHR" comment="Direct Rendering Manager interface"/>
```
emersion, line 37: Do we need a separate `vulkan_drm.h` header here? The extension doesn't rely on any platform-specific header.

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
```diff
@@ -0,0 +1,48 @@
...
+
+include::{generated}/api/protos/vkGetDrmDisplayEXT.txt[]
+
+  * pname:physicalDevice The physical device to query the display from.
+  * pname:drmFd Connection to the DRM resource.
+  * pname:connectorId Identifier of the specified DRM connector.
+  * pname:display The corresponding slink:VkDisplayKHR handle will be returned
+    here.
+
+If there is not slink:VkDisplayKHR corresponding to the pname:connectorID on
```
emersion, line 41: Typo: "no"

emersion, line 41: Typo: "connectorId"

xml/vk.xml
bl4ckb0ne: Shoud I put it in the main header, inside linux specific defines?

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
```diff
@@ -0,0 +1,48 @@
...
+--
+
+Before acquiring a display from the DRM interface, the caller may want to
+select a specific sname:VkDisplayKHR handle by identifying it using a
+pname:connectorId. To do so, call:
+
+include::{generated}/api/protos/vkGetDrmDisplayEXT.txt[]
+
+  * pname:physicalDevice The physical device to query the display from.
+  * pname:drmFd Connection to the DRM resource.
```
emersion, line 36: "DRM resource" isn't well-defined. It may be confused with "DRM object", which is a connector/CRTC/blob/etc, and is not what we want to refer to here.

Something like "DRM node file descriptor" would be less ambiguous here.

xml/vk.xml
```diff
@@ -10388,6 +10389,19 @@ typedef void <name>CAMetalLayer</name>;
             <param><type>VkCommandBuffer</type> <name>commandBuffer</name></param>
             <param>const <type>VkCuLaunchInfoNVX</type>* <name>pLaunchInfo</name></param>
         </command>
+        <command successcodes="VK_SUCCESS" errorcodes="VK_ERROR_INITIALIZATION_FAILED">
+            <proto><type>VkResult</type> <name>vkAcquireDrmDisplayEXT</name></proto>
+            <param><type>VkPhysicalDevice</type> <name>physicalDevice</name></param>
+            <param><type>int32_t</type> <name>drmFd</name></param>
+            <param><type>VkDisplayKHR</type> <name>display</name></param>
+        </command>
+        <command successcodes="VK_SUCCESS" errorcodes="VK_ERROR_INITIALIZATION_FAILED">
```
emersion, line 10398: We may want to allow `VK_ERROR_OUT_OF_HOST_MEMORY` as an error code here.

xml/vk.xml
emersion: Hm, I'm not sure I understand. I don't think we need to have `VK_USE_PLATFORM_DRM_KHR` nor `vulkan_drm.h`.

See https://github.com/KhronosGroup/Vulkan-Docs/pull/1356 for an example of an ext that is DRM-specific but doesn't need to pull any platform-specific header.

xml/vk.xml
```diff
@@ -15149,10 +15163,12 @@ typedef void <name>CAMetalLayer</name>;
                 <type name="PFN_vkDeviceMemoryReportCallbackEXT"/>
             </require>
         </extension>
-        <extension name="VK_EXT_extension_286" number="286" type="instance" author="EXT" contact="Drew DeVault sir@cmpwn.com" supported="disabled">
+        <extension name="VK_EXT_acquire_drm_display" number="286" type="instance" author="EXT" contact="Drew DeVault sir@cmpwn.com" platform="drm" supported="vulkan">
```
emersion, line 15166: We probably want to add `requires="VK_EXT_direct_mode_display"` here

emersion:
cc @cubanismo

xml/vk.xml
bl4ckb0ne: `VK_EXT_acquire_xlib_display` has its own header, `VK_KHR_xlib_surface` and `VK_KHR_wayland_surface` too. This extension being Linux only, but not tied to a specific "platform", I was hesitating between adding a `linux` or a `drm` specific header.

emersion: Why do we need to specify a `platform` at all?

bl4ckb0ne: Should the extension be Linux specific?

emersion: Probably not, it should work on BSDs too.

bl4ckb0ne: So I can leave it in the main header, and it should be unimplemented on non-unix platforms?

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
bl4ckb0ne: Thanks! I was struggling to find a good definition and hoped somebody would propose something better.

xml/vk.xml
bl4ckb0ne: I ended up following what `VK_EXT_image_drm_format_modifier` did.

bl4ckb0ne:
CI is now [green](https://github.com/bl4ckb0ne/Vulkan-Docs/actions/runs/854299796)

cc @cubanismo

appendices/VK_EXT_acquire_drm_display.txt
```diff
@@ -0,0 +1,28 @@
...
+    No known IP claims.
+*Contributors*::
+  - Simon Zeni, Status Holdings, Ltd.
+
+=== Description
+
+This extension allows an application to take exclusive control of a display
+using the Direct Rendering Manager (DRM) interface. When acquired, the display
+will be under full control of the application until it is released or the
+connection is closed.
```
cubanismo, line 17: What connection?

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
```diff
@@ -0,0 +1,48 @@
...
+
+include::{generated}/api/protos/vkAcquireDrmDisplayEXT.txt[]
+
+  * pname:physicalDevice The physical device the display is on.
+  * pname:drmFd Connection to the DRM resource.
+  * pname:display The display the caller wishes Vulkan to control.
+
+All permissions necessary to control the display are granted to the Vulkan
+instance associated with the provided pname:physicalDevice until the display is
+either released or the connection is closed.
```
cubanismo, line 19: What connection?

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
```diff
@@ -0,0 +1,48 @@
...
+--
+
+Before acquiring a display from the DRM interface, the caller may want to
+select a specific sname:VkDisplayKHR handle by identifying it using a
+pname:connectorId. To do so, call:
+
+include::{generated}/api/protos/vkGetDrmDisplayEXT.txt[]
+
+  * pname:physicalDevice The physical device to query the display from.
+  * pname:drmFd Connection to the DRM node file descriptor.
```
cubanismo, line 36: Is it required that drmFd correspond to physicalDevice somehow? I assume there is, and some explicit valid usage should likely be added to indicate this.

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
```diff
@@ -0,0 +1,48 @@
...
+  * pname:connectorId Identifier of the specified DRM connector.
```
cubanismo, line 37: Somewhat related to above, suggest adding "on the DRM device associated with drmFd" or something to clarify this a bit. I forget what the namespace of connector IDs is, but I assume you can't correlate connectors on other DRM devices using any random DRM fd + connector ID.

Alternatively, explicit valid usage could be added to clear this up.

cubanismo (changes requested):

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
bl4ckb0ne: It is, I realised that earlier today while implementing it in Mesa. Should we define a new VkResult value like `VK_INVALID_DISPLAY_EXT`, or should we use `VK_ERROR_UNKNOWN` or another existing value?

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
bl4ckb0ne: It should be clarified yes. Same as above, I'm hesitating between adding a `VK_INVALID_CONNECTOR_EXT` or `VK_ERROR_UNKNOWN`.

appendices/VK_EXT_acquire_drm_display.txt
bl4ckb0ne:
```diff
 using the Direct Rendering Manager (DRM) interface. When acquired, the display
-will be under full control of the application until it is released or the
-connection is closed.
+will be under full control of the application until the display is either
+released, or the display connection is closed or lost.
```

Should it be more specific about the connection being the physical cable?

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
bl4ckb0ne:
```diff
 instance associated with the provided pname:physicalDevice until the display is
-either released or the connection is closed.
+either released or the display connection is closed or lost.
```

Same comment as above.

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
cubanismo: Vulkan doesn't usually check things like this in driver implementations. It just asserts the parameters must be valid and compatible in some way in valid usage language in the spec, and then validation layers would do any possible checking/reporting of errors. Drivers are free to crash and burn or return any random error they please (undefined behavior) if valid usage is violated.

bl4ckb0ne: So I could just return `VK_ERROR_UNKNOWN` and set errno to EINVAL ?

appendices/VK_EXT_acquire_drm_display.txt
cubanismo: If this is trying to deal with hot-unplug, it might be better to write some text detailing interactions between leases and unplug events and then reference that here. I think the display_control extension has some hotplug functionality.

I was concerned this language was left over from spec language dealing with wl_display connections.

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
cubanismo: That'd be one valid implementation, I believe, but the spec shouldn't require any particular behavior, nor should portable apps rely on any particular behavior in this case. E.g., you can't conformance test it because the driver could conceivably segfault in this case without being non-conformant.

appendices/VK_EXT_acquire_drm_display.txt
bl4ckb0ne: Closing the connection refers mainly to closing the DRM fd (either master or leased). Lost would be more about pulling the cable from the display connector or a power loss.

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
bl4ckb0ne: I'll go with `VK_ERROR_UNKNOWN` and leave a note about the fact that the implementation could set errno to an adequate value.

appendices/VK_EXT_acquire_drm_display.txt
cubanismo: I don't know that this is sufficiently clear to readers then. It wasn't to me at least. I would prefer it be clarified (use text closer to what you've used above to explain what connection closed means), but up to you.

cubanismo:
The functionally here looks good overall to me. Still have a suggested wording change inline above, but happy to sign off on this either way on behalf of NV for multi-vendor status.

cubanismo (approved):

appendices/VK_EXT_acquire_drm_display.txt
```diff
@@ -0,0 +1,29 @@
...
+
+=== Description
+
+This extension allows an application to take exclusive control of a display
+using the Direct Rendering Manager (DRM) interface. When acquired, the display
+will be under full control of the application until the display is either
+released, the provided DRM fd is closed, or the physical connection to the
+display is lost.
+
+include::{generated}/meta/{refprefix}VK_EXT_acquire_drm_display.txt[]
```
oddhack, line 20: This include should go at the start of the appendix, followed by '=== Other Extension Metadata' and then the '*Last Modified Date*' etc.

oddhack: Then at this point (end of Description), add the generated interfaces include:

> include::{generated}/interfaces/VK_EXT_acquire_drm_display.txt[]

emersion:
Follow-up for the wording concern from https://github.com/KhronosGroup/Vulkan-Docs/pull/1529#discussion_r640152100: I also think "connection" is confusing here. In KMS terms, "connection" usually refers to a connector's current status (connected/disconnected, see `drmModeConnector`).

Also, the use of "display connection closed or lost" adds more confusion, as I think this refers to unplugging?

Can we completely remove the use of "connection", replacing it either by simply "DRM FD" or "unplug" depending on the intended meaning?

bl4ckb0ne:
I made the change to specify that the DRM fd should not be closed during usage, as discussed with @emersion 

@cubanismo if you want to take a look to be sure that everything is ok.

emersion:
The motivation is that if the application closes the DRM FD behind the Vulkan implementation's back, there's no way for the Vulkan implementation to figure it out. The IOCTLs on the closed FD will fail, or worse if the the FD number is re-used for a completely unrelated file.

It's somewhat tempting to make the Vulkan implementation take ownership over the DRM FD, however this won't work well in the case where the application wants to drive multiple connectors.

I'm seeing the DRM FD as an equivalent of an Xlib `Display`.

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
```diff
@@ -0,0 +1,56 @@
...
+  * pname:display The display the caller wishes Vulkan to control.
+
+All permissions necessary to control the display are granted to the Vulkan
+instance associated with the provided pname:physicalDevice until the display is
+either released or the connector is unplugged.
+If any error is encountered during the acquisition of the display, the call
+must return the error code ename:VK_ERROR_INITIALIZATION_FAILED.
+
+The provided DRM fd should not be closed before the display is released,
+attempting to do it may result in undefined: behaviour.
```
emersion, line 24: Typo: extra `:` here

bl4ckb0ne: It's to avoid [CI failure](https://github.com/bl4ckb0ne/Vulkan-Docs/runs/2784979633?check_suite_focus=true), see https://github.com/KhronosGroup/Vulkan-Docs/blob/b4e8cd820b2487bc892b391fb26b49501473a6a6/style/writing.txt#L254

Although I am not sure if `undefined:` or `undefined{cl}` is prefered, I've seen both in other extensions.

emersion: Oh. My bad.

oddhack: 'undefined:' is a markup convention used to indicate that the term is agreed by implementers in this context, and specifies what sort of undefined (behavior | values | etc.) may occur.

'undefined{cl}' is a markup convention used only in the style guide in places where we want to present markup, not invoke the asciidoctor macro of the same name. If there are any outstanding extension proposals containing this markup, they need to be corrected. I don't see this happening in main branch.

oddhack: https://www.khronos.org/registry/vulkan/specs/1.2/styleguide.html#writing-undefined

oddhack:
@emersion is this ready to be published? If so and with @cubanismo's approval we could put it out as an EXT on the 21st - I'd like to give the working group one last chance to comment on Monday, though.

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
```diff
@@ -0,0 +1,56 @@
...
+[open,refpage='vkAcquireDrmDisplayEXT',desc='Acquire access to a VkDisplayKHR using DRM',type='protos']
+--
+
+To acquire permission to directly a display in Vulkan from the Direct Rendering
+Manager (DRM) interface, call:
+
+include::{generated}/api/protos/vkAcquireDrmDisplayEXT.txt[]
+
+  * pname:physicalDevice The physical device the display is on.
+  * pname:drmFd Connection to the DRM node file descriptor.
```
emersion, line 14: This still reads "connection". It should just read "DRM primary file descriptor".

emersion: Ideally also mention that the DRM FD must have DRM master permissions.

chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
```diff
@@ -0,0 +1,56 @@
...
+--
+
+Before acquiring a display from the DRM interface, the caller may want to
+select a specific sname:VkDisplayKHR handle by identifying it using a
+pname:connectorId. To do so, call:
+
+include::{generated}/api/protos/vkGetDrmDisplayEXT.txt[]
+
+  * pname:physicalDevice The physical device to query the display from.
+  * pname:drmFd Connection to the DRM node file descriptor.
```
emersion, line 39: Ditto, remove "connection" here.

DRM master is not strictly needed here, since this is just used for information gathering purposes.

emersion (approved):
This version LGTM!

stonesthrow:
@oddhack - approved.

bl4ckb0ne:
I added clarifications to `vkAcquireDrmDisplayEXT` to reflect what was done in `VkGetDrmDisplayEXT` in case the supplied DRM FD doesn't match the supplied VkPhysicalDevice

```diff
diff --git a/chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt b/chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
index b7d77dea..d32d30cb 100644
--- a/chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
+++ b/chapters/VK_EXT_acquire_drm_display/acquire_drm_display.txt
@@ -17,7 +17,9 @@ include::{generated}/api/protos/vkAcquireDrmDisplayEXT.txt[]
 All permissions necessary to control the display are granted to the Vulkan
 instance associated with the provided pname:physicalDevice until the display is
 either released or the connector is unplugged.
-The supplied DRM FD mush have DRM master permissions.
+The provided pname:drmFd must correspond to the one owned by the
+pname:physicalDevice. If not, the error code ename:VK_ERROR_UNKNOWN must be
+returned. The DRM FD mush have DRM master permissions.
 If any error is encountered during the acquisition of the display, the call
 must return the error code ename:VK_ERROR_INITIALIZATION_FAILED.
 ```

emersion:
This minor clarification LGTM.

PR #1532, opened 2021-05-21
check whether __cplusplus is defined before comparing the value

luckyxxl:
... to fix compilation with C and  -Wundef


oddhack:
I believe an equivalent fix was done in the 1.2.179 spec update. If you agree, please close this.

luckyxxl:
Yes, this is resolved with the spec update, thanks!

PR #1538, opened 2021-05-25
vk.xml: Include `vk_platform` type-header from `vulkan/vk_platform.h`

MarijnS95:
It is generally expected that "top level" includes are relative to the `./include` directory in this repo.  The video headers, sitting at the root of the `./include` folder in `./include/vk_video/vulkan_video_*.h`, are listed as `vk_video/vulkan_video_*.h` in `vk.xml`.  Likewise the `vk_platform` type-header is supposed to be listed as `vulkan/vk_platform.h` to match its relative location under `./include`.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -74,7 +74,7 @@ branch of the member gitlab server.
     </tags>
 
     <types comment="Vulkan type definitions">
-        <type name="vk_platform" category="include">#include "vk_platform.h"</type>
+        <type name="vk_platform" category="include">#include "vulkan/vk_platform.h"</type>
 
             <comment>WSI extensions</comment>
 


krOoze:
That's pretty weird actually. I think `vk_video/` should be `vulkan/video/`

MarijnS95:
> That's pretty weird actually. I think `vk_video/` should be `vulkan/video/`.

I agree, created #1541 for it to track this separately for follow-up with the Video folks.

oddhack:
@zlatinski Video TSG should sign off on this one.

MarijnS95:
@oddhack I don't think `vk_platform.h` is solely related to Video, it's just the first occasion where we need access to Vulkan headers in addition to `vk.xml` and noticed this inconsistency.

oddhack:
> @oddhack I don't think `vk_platform.h` is solely related to Video, it's just the first occasion where we need access to Vulkan headers in addition to `vk.xml` and noticed this inconsistency.

OK. We'll take a look at it internally. It usually takes a lot of back-and-forthing between the different downstream components like the SDK before everyone signs off on even a simple change like this, so I wouldn't expect a quick response.

oddhack:
We talked about this briefly and are OK with taking it, but not until after next week's updates for internal scheduling reasons.

MarijnS95:
> We talked about this briefly and are OK with taking it, but not until after next week's updates for internal scheduling reasons.

Great, thanks for that and thanks for keeping this issue updated!

zlatinski:
We are expecting Vulkan Video  apps to add to the include paths both ..include/vk_video and ..include/vulkan, assuming the app is not using header file name spaces. The prefered way,, however, would be to only include the top level ../include path. The app, in this case should use the appropriate NS directories along with the included header file.
For example:
#include "vk_video/vulkan_video_codec_h264std_decode.h"
#include "vulkan/vulkan.h"

MarijnS95:
@zlatinski Totally, that makes the most sense. When looking at `vk.xml` that's already possible because it specifies `vk_video/` as include path, but `vk_platform.h` still has to be moved as per this PR to enable that.

PR #1547, opened 2021-06-03
fix wrong struct name in renderpass chapter text

luckyxxl:


oddhack:
Thanks! This has already been fixed internally and will show up in the next spec update by Monday.

PR #1556, opened 2021-06-14
Add missing requires attribute for VkPipelineLayoutCreateFlags

expipiplus1:


oddhack:
The Flags type must exist because it's the type of a structure member. But it does not require the FlagBits type yet, because there are no FlagBits defined yet. When they are, we will add this requirement. This is true of multiple Flags types that don't yet have any extensions defining bits for them.

expipiplus1:
True, I think what makes this case a little weird is that `VkPipelineLayoutCreateFlagBits` does exist in the spec (and is the only case of both `Flags` and `FlagBits` existing but not being linked by a `"requires"`.

I suppose it's the difference between "no FlagBits defined yet" and "an empty set of FlagBits is defined". I interpreted having `VkPipelineLayoutCreateFlagBits` defined in the spec to indicate the latter.

I don't hugely mind either way this being merged or closed, I've already worked around this new situation :)

oddhack:
> True, I think what makes this case a little weird is that `VkPipelineLayoutCreateFlagBits` does exist in the spec (and is the only case of both `Flags` and `FlagBits` existing but not being linked by a `"requires"`.

Where are you seeing it in the spec? I just grepped the generated all-extensions HTML build to double-check and can't find it anywhere.

expipiplus1:
Ah sorry, I meant the xml!

oddhack:
> Ah sorry, I meant the xml/registry (which does declare itself part of the specification)!

It's in a sort of weird half-life position in that a disabled extension is reserving a FlagBit in that type. Other FlagBits will go through the same state as they get bits reserved. But the key point is that the extension is disabled so shouldn't be getting processed.

expipiplus1:
Yeah, it wasn't so much that there are bits mentioned in disabled extensions (which are correctly ignored in my program). It was the existence of this orphaned `VkPipelineLayoutCreateFlagBits`. At the point at which it wants to determine information (such as the width) of `VkPipelineLayoutCreateFlagBits` it doesn't yet know that it's ultimately not required by anything. 

Seems like I should either 

- Find a more robust/correct way of associating `Flags` with `FlagBits` (`s/FlagBits/Flags` perhaps?)

or

- be a little smarter about determining what is eventually required.

From my point of view having `VkPipelineLayoutCreateFlagBits` in the registry but unused is worse/less-consistent than either of not having it at all or, having it "required" (albeit empty) by `VkPipelineLayoutCreateFlags`. Obviously I'm not the sole consumer here, nor the most important one! (and indeed it's no longer a problem at all).

oddhack:
> be a little smarter about determining what is eventually required.

That's my suggestion. All our XML processing is dependency-driven - it starts with a set of core versions and extensions and transitively tags everything required by them, then calls the generator scripts only for the required API elements. So that's how I see this - people are traversing the XML in ways they shouldn't and having to deal with "API"s that actually aren't part of the Vulkan API, but exist in the XML as placeholders. I've actually been surprised that so many people downstream seem eager to write all their own code to deal with XML traversal.

expipiplus1:
Indeed this is pretty much what I do after parsing the spec and annotating
it with some useful values. It just so happened that an assumption of mine
was violated before I could get to the dependency driven part.

oddhack:
It sounds like you have worked around the problem, so closing this. When one of the extensions reserving a bit in the type gets published, the dependency will be updated.

PR #1587, opened 2021-07-21
Correct struct name in some atomic float spirv capabilities

expipiplus1:


oddhack:
Should be fixed in the 1.2.186 spec update. There was an equivalent internal MR that got there first, so closing this.

expipiplus1:
Thanks!

PR #1592, opened 2021-07-29
Marks all pNext fields as optional in `vk.xml`

ashpil:
As discussed in #1396, `pNext` fields should be marked as optional. This wasn't ensured for more recently added changes (which breaks some binding generators), so this PR updates the XML file to match that discussion.

#1396 mentions
> add something to the style guide about setting this attribute on pNext members

which appears to not have happened. It might be worth looking into again.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -2360,18 +2360,18 @@ typedef void <name>CAMetalLayer</name>;
         </type>
         <type category="struct" name="VkImportMemoryZirconHandleInfoFUCHSIA" structextends="VkMemoryAllocateInfo">
             <member values="VK_STRUCTURE_TYPE_IMPORT_MEMORY_ZIRCON_HANDLE_INFO_FUCHSIA"><type>VkStructureType</type> <name>sType</name></member>
-            <member>const <type>void</type>*                      <name>pNext</name></member>
+            <member optional="true">const <type>void</type>*                      <name>pNext</name></member>
             <member optional="true"><type>VkExternalMemoryHandleTypeFlagBits</type> <name>handleType</name></member>
             <member optional="true"><type>zx_handle_t</type>           <name>handle</name></member>
         </type>
         <type category="struct" name="VkMemoryZirconHandlePropertiesFUCHSIA" returnedonly="true">
             <member values="VK_STRUCTURE_TYPE_MEMORY_ZIRCON_HANDLE_PROPERTIES_FUCHSIA"><type>VkStructureType</type> <name>sType</name></member>
-            <member><type>void</type>*                            <name>pNext</name></member>
+            <member optional="true"><type>void</type>*                            <name>pNext</name></member>
             <member><type>uint32_t</type>                         <name>memoryTypeBits</name></member>
         </type>
         <type category="struct" name="VkMemoryGetZirconHandleInfoFUCHSIA">
             <member values="VK_STRUCTURE_TYPE_MEMORY_GET_ZIRCON_HANDLE_INFO_FUCHSIA"><type>VkStructureType</type> <name>sType</name></member>
-            <member>const <type>void</type>*                      <name>pNext</name></member>
+            <member optional="true">const <type>void</type>*                      <name>pNext</name></member>
             <member><type>VkDeviceMemory</type>                   <name>memory</name></member>
             <member><type>VkExternalMemoryHandleTypeFlagBits</type> <name>handleType</name></member>
         </type>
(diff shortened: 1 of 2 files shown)


xml/vk.xml
```diff
@@ -5993,7 +5993,7 @@ typedef void <name>CAMetalLayer</name>;
         </type>
         <type category="struct" name="VkPhysicalDeviceRayTracingMotionBlurFeaturesNV" structextends="VkPhysicalDeviceFeatures2,VkDeviceCreateInfo">
             <member values="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_RAY_TRACING_MOTION_BLUR_FEATURES_NV"><type>VkStructureType</type> <name>sType</name></member>
-            <member noautovalidity="true"><type>void</type>*        <name>pNext</name></member>
+            <member optional="true" noautovalidity="true"><type>void</type>*        <name>pNext</name></member>
```
Tobski, line 5996: I'm not sure what the interaction is going to be like here - we don't usually have optional and notautovalidity on the same parameter - @oddhack  any idea what this might do?

xml/vk.xml
```diff
@@ -2360,18 +2360,18 @@ typedef void <name>CAMetalLayer</name>;
...
             <member values="VK_STRUCTURE_TYPE_IMPORT_MEMORY_ZIRCON_HANDLE_INFO_FUCHSIA"><type>VkStructureType</type> <name>sType</name></member>
-            <member>const <type>void</type>*                      <name>pNext</name></member>
+            <member optional="true">const <type>void</type>*                      <name>pNext</name></member>
             <member optional="true"><type>VkExternalMemoryHandleTypeFlagBits</type> <name>handleType</name></member>
             <member optional="true"><type>zx_handle_t</type>           <name>handle</name></member>
         </type>
         <type category="struct" name="VkMemoryZirconHandlePropertiesFUCHSIA" returnedonly="true">
             <member values="VK_STRUCTURE_TYPE_MEMORY_ZIRCON_HANDLE_PROPERTIES_FUCHSIA"><type>VkStructureType</type> <name>sType</name></member>
-            <member><type>void</type>*                            <name>pNext</name></member>
+            <member optional="true"><type>void</type>*                            <name>pNext</name></member>
```
Tobski, line 2369: I think these are fine, but because the struct is tagged "returnedonly" I don't think this actually does anything? Should probably double check it's not outputting anything. It's probably harmless, but almost definitely pointless from the spec perspective.

Tobski (commented):
Couple of places I'm not sure about so tagging in @oddhack to check what this actually does to generated spec artefacts.

xml/vk.xml
ashpil: Ah, this is the downside of just replacing it all with automatic tools - didn't realize it didn't quite make sense in this situation. I think coming from the binding generator perspective this still makes sense, as you still need the `optional` flag to infer the correct type.

xml/vk.xml
oddhack: Nothing, AFAICT. There are two structures with optional & noautovalidity both "true" in vk.xml - VkCopyCommandTransformInfoQCOM and VkPhysicalDeviceShaderTerminateInvocationFeaturesKHR - and removing the optional="true" has zero effect on generated VUs.

I think the only way to deal with this going forward will be adding a CI check to require optional="true" on pNext members. It is a general problem with people want to infer things from the XML that are not relevant to any of the Khronos uses of it. Fortunately this is a fairly straightforward check.

ashpil: Is this CI check a blocker for this PR?

oddhack: No, I'll track it separately. We have some XML sanity checking tools which aren't currently enabled but can be soon, and this is a sensible check to add when we do that.

xml/vk.xml
oddhack: Because it's returnedonly, the only VU it gets is 'pNext must: be NULL', whether the 'optional' is there or not.

xml/vk.xml
oddhack: @ashpil check out #1597, which (hopefully) implements the check. It would be helpful if you could cherry-pick those changes (or rebase on / retarget that branch) and run scripts/xml_consistency.py, or just enable Github Actions on your repository so our CI will run. I am fairly sure there are at least a few more pNexts your tool didn't catch.

https://github.com/KhronosGroup/Vulkan-Docs/pull/1597/checks?check_run_id=3241250501

ashpil: @oddhack Done. The script was using `is not` when it should've been using `!=`, so it was giving quite a few false positives. 

I fixed that, and one warning remains about `vkGetDrmDisplayEXT` - not sure what that is about. Also `spec-generate` seems to error.

oddhack: vkGetDrmDisplayEXT violated our own naming conventions. There's no way to fix it in the C API at this point, so it is treated as an exception (thus the warning instead of an error that would break CI).

The other problem seems to derive from my updating the Docker image used for CI to add a new component needed by the build. Unfortunately it also updated nodejs from 10.x to 12.x and that is breaking an unrelated postprocessing step used on the generated HTML (via scripts/translate_math.js). I frankly don't know enough about the issue, which is something to do with ES6 import vs. CommonJS require vs. transpilation vs. package.json vs. bundling vs. about 5 other things I've found references to in a short time searching, to understand what can be done to fix this, so I will probably have to figure out how to back the image off to the older nodejs for the time being.

oddhack:
@ashpil I think we could close this in favor of #1597, which mostly fixes the CI issue (there is a remaining problem with Vulkan-Hpp which needs to get fixed on their end) as well as including the pNext optional="true" changes. If you see anything #1597 is missing relative to #1592, probably better to suggest it on that PR. I have updated the CI image so that particular problem should be resolved.

xml/vk.xml
oddhack: OK, resolved now. It was using a more recent version of escape-string-regexp that was apparently an ES6 module, locking the container build to the previously used version is a workaround.

ashpil:
@oddhack Sounds good.

Issue #1601, opened 2020-06-04
VkShaderModuleCreateFlagBits Reference Gives 404 Error


samwalls-arm:
The following link gives me a 404 error currently: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkShaderModuleCreateFlagBits.html

I found the link on the reference page for [VkShaderModuleCreateFlag](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkShaderModuleCreateFlags.html).

oddhack:
This should be fixed in the 1.2.190 spec update.

PR #1609, opened 2021-08-16
Make treatment of empty Vk*FlagBits types consistent

oddhack:
This originated with #1601 noting a cross-link to a nonexistent FlagBits
refpage, then expanded in scope a bit as I dealt with inconsistent treatment
of unused FlagBits types.

All FlagBits types with no bits defined are now treated consistently:
the unused (and currently useless) FlagBits type has no description in
the spec and no dependency on it exists in vk.xml at the point where the
corresponding Flags type is required. This meant removing
VkDeviceQueueCreateFlagBits, VkShaderModuleCreateFlagBits, and
VkPipelineCacheCreateFlagBits from the VK_VERSION_1_0 XML block. The
only semantic header file difference in the current headers is that
VkShaderModuleCreateFlagBits is no longer defined in vulkan_core.h:

```
2125a2126,2129
>
> typedef enum VkShaderModuleCreateFlagBits {
>     VK_SHADER_MODULE_CREATE_FLAG_BITS_MAX_ENUM = 0x7FFFFFFF
> } VkShaderModuleCreateFlagBits;
```

In addition the spec interface listings for VkDeviceQueueCreateFlagBits and
VkPipelineCacheCreateFlagBits drop the

```
// Provided by VK_VERSION_1_0
```

comment, but still describe each actual enum with the version or
extension that introduced it. These two type definitions also moved
around in vulkan_core.h to be with the version or extension that first
added an extending enum to them.

It's possible that some XML processors do not recognize

> \<enum extends="VkSomeFlagBits">

as an explicit requirement of VkSomeFlagBits. There's a temporary
workaround for this, although we plan to remove it before the PR is
accepted, and a note in the schema document to clarify this case.

Futureproofing XML comments about needing to define the FlagBits types
when they're needed have been made consistent.

I think this is a small net win, although it is possible that there is
existing code which defines a VkShaderModuleCreateFlagBits variable
somewhere despite its uselessness, and would not compile with this
change.

The other alternative to remain consistent would be to *always* require
the FlagBits type whenever a Flags type is defined as a member of a new
structure, even when the FlagBits type has no enumerants defined in to
begin with. This would add many new, empty FlagBits enumerations to the
headers, most of which would probably never be used.

We will let this PR sit for feedback for a bit before deciding to move
this direction, or not.

### DIFF
--- a/chapters/descriptorsets.txt
+++ b/chapters/descriptorsets.txt
@@ -2965,7 +2965,7 @@ If a binding has a pname:descriptorCount of zero, it is skipped.
 This behavior applies recursively, with the update affecting consecutive
 bindings as needed to update all pname:descriptorCount descriptors.
 Consecutive bindings must: have identical
-elink::VkDescriptorType, elink::VkShaderStageFlags,
+elink:VkDescriptorType, tlink:VkShaderStageFlags,
 ifdef::VK_VERSION_1_2,VK_EXT_descriptor_indexing[]
 elink:VkDescriptorBindingFlagBits,
 endif::VK_VERSION_1_2,VK_EXT_descriptor_indexing[]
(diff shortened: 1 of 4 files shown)


oddhack:
No public feedback after one week. Vulkan WG has already approved this. Signed off to merge.

PR #1611, opened 2021-08-16
Create Vulkan-Projects.adoc

marty-johnson59:
Moving list of Vulkan project repos to a file in Vulkan-Docs (from Vulkan-Ecosystem which is being archived)

### DIFF
--- a/Vulkan-Projects.adoc
+++ b/Vulkan-Projects.adoc
@@ -1,0 +1,160 @@
+# Map of Vulkan project GitHub repositories
+
+Vulkan is a big place and there's a lot out there that is useful! In no particular order...
+
+## Khronos/Vulkan® Specification
+Vulkan® Specification - formal documentation of the Vulkan API. This includes the main API Specification, the reference (man) pages, the XML API Registry, and related tools and scripts.
+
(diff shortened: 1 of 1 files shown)


marty-johnson59:
Note, the README.adoc file in Vulkan-Docs points to the Vulkan-Ecosystem for a list of other Vulkan related Github repos.  Since we're archiving Vulkan-Ecosytem, I'm suggesting moving this list to an adoc file in Vulkan-Docs instead -and then updating the README file to point here.

oddhack:
I think this makes more sense in the Vulkan-Web-Registry repo, at least for now. There's already a list of repos in the index page for that repository and that should either point to a separate document with this information, or be updated to incorporate it (probably the former).

marty-johnson59:
Yah, probably better to keep it out of Vulkan-docs.  I do think there's value in this being "human-readable" (vs. php), so keeping this in adoc and referencing it from both the index php file and the readme.adoc on Vulkan-docs makes sense to me.  I'll open an new PR on Vulkan-Web-Registry (and close this one) to add this adoc file there.  Might ask you to make the change to the index file as I'm not familiar with php and don't want to break something (is this file used by any tools?).

PR #1628, opened 2021-09-04
Add VK_EXT_non_seamless_cube_map

DadSchoorse:
Core Vulkan always uses seamless cube map edge wrapping. Out of bounds accesses select texels from the neighboring face with linear filter, or get clamp to edge behavior with nearest filter. D3D9 has no seamless cube maps, the sampler wrap modes apply within the selected face. OpenGL has a global toggle with ARB_seamless_cube_map and a per texture toggle with ARB_seamless_cubemap_per_texture.

For DXVK (D3D9 to Vulkan translation) we need D3D9 cube map behavior since games were designed with it in mind and some don't look correct with seamless cube map edge handling. Zink or other OpenGL translation layers might also be interested in non-seamless cube maps to implement the previously mentioned OpenGL extensions.

anv implementation: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/12730
radv implementation: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/12729
some basic CTS tests: https://github.com/KhronosGroup/VK-GL-CTS/pull/286

### DIFF
--- a/chapters/samplers.txt
+++ b/chapters/samplers.txt
@@ -296,6 +296,11 @@ ifdef::VK_EXT_fragment_density_map[]
     If pname:flags includes ename:VK_SAMPLER_CREATE_SUBSAMPLED_BIT_EXT, then
     pname:unnormalizedCoordinates must: be ename:VK_FALSE
 endif::VK_EXT_fragment_density_map[]
+ifdef::VK_EXT_non_seamless_cube_map[]
+  * If the <<features-nonSeamlessCubeMap, pname:nonSeamlessCubeMap>> feature
+    is not enabled, pname:flags must: not include
+    ename:VK_SAMPLER_CREATE_NON_SEAMLESS_CUBE_MAP_BIT_EXT
+endif::VK_EXT_non_seamless_cube_map[]
 ifdef::VK_EXT_custom_border_color[]
   * [[VUID-VkSamplerCreateInfo-borderColor-04011]]
     If pname:borderColor is one of ename:VK_BORDER_COLOR_FLOAT_CUSTOM_EXT or
@@ -351,7 +356,14 @@ ifdef::VK_EXT_fragment_density_map[]
     specifies that the implementation may: use approximations when
     reconstructing a full color value for texture access from a subsampled
     image.
+endif::VK_EXT_fragment_density_map[]
+ifdef::VK_EXT_non_seamless_cube_map[]
+  * ename:VK_SAMPLER_CREATE_NON_SEAMLESS_CUBE_MAP_BIT_EXT
+    specifies that <<textures-cubemapedge, cube map edge handling>> is not
+    performed.
+endif::VK_EXT_non_seamless_cube_map[]
 
+ifdef::VK_EXT_fragment_density_map[]
 [NOTE]
 .Note
 ====
--- a/appendices/VK_EXT_non_seamless_cube_map.txt
+++ b/appendices/VK_EXT_non_seamless_cube_map.txt
@@ -1,0 +1,29 @@
+// Copyright 2021 Georg Lehmann
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_EXT_non_seamless_cube_map.txt[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2021-09-04
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Georg Lehmann
+
+=== Description
+
+This extension provides functionality to disable <<textures-cubemapedge, cube map edge handling>>
+on a per sampler level which matches the behavior of other graphics APIs.
+
+This extension may be useful for building translation layers for those APIs
+or for porting applications that rely on this cube map behavior.
+
+include::{generated}/interfaces/VK_EXT_non_seamless_cube_map.txt[]
+
+=== Version History
+
+  * Revision 1, 2021-09-04 (Georg Lehmann)
+    - First Version
--- a/chapters/features.txt
+++ b/chapters/features.txt
@@ -5238,6 +5238,24 @@ include::{generated}/validity/structs/VkPhysicalDeviceShaderEarlyAndLateFragment
 --
 endif::VK_AMD_shader_early_and_late_fragment_tests[]
 
+ifdef::VK_EXT_non_seamless_cube_map[]
+[open,refpage='VkPhysicalDeviceNonSeamlessCubeMapFeaturesEXT',desc='Structure describing features to disable seamless cube maps',type='structs']
+--
+The sname:VkPhysicalDeviceNonSeamlessCubeMapFeaturesEXT structure is defined as:
+
+include::{generated}/api/structs/VkPhysicalDeviceNonSeamlessCubeMapFeaturesEXT.txt[]
+
+This structure describes the following feature:
+
+  * [[features-nonSeamlessCubeMap]] pname:nonSeamlessCubeMap indicates that the
+    implementation supports ename:VK_SAMPLER_CREATE_NON_SEAMLESS_CUBE_MAP_BIT_EXT.
+
+:refpage: VkPhysicalDeviceNonSeamlessCubeMapFeaturesEXT
+include::{chapters}/features.txt[tag=features]
+
+include::{generated}/validity/structs/VkPhysicalDeviceNonSeamlessCubeMapFeaturesEXT.txt[]
+--
+endif::VK_EXT_non_seamless_cube_map[]
 
 [[features-requirements]]
 == Feature Requirements
@@ -5726,6 +5744,10 @@ ifdef::VK_AMD_shader_early_and_late_fragment_tests[]
     `apiext:VK_AMD_shader_early_and_late_fragment_tests` extension is
     supported.
 endif::VK_AMD_shader_early_and_late_fragment_tests[]
+ifdef::VK_EXT_non_seamless_cube_map[]
+  * <<features-nonSeamlessCubeMap, pname:nonSeamlessCubeMap>>, if the
+    `<<VK_EXT_non_seamless_cube_map>>` extension is supported.
+endif::VK_EXT_non_seamless_cube_map[]
 
 All other features defined in the Specification are optional:.
 
--- a/chapters/textures.txt
+++ b/chapters/textures.txt
@@ -603,6 +603,11 @@ then the texel value comes from the value in image memory.
   * If the read is the result of an image sample instruction or image gather
     instruction, and
   * If the image is not a cube image,
+ifdef::VK_EXT_non_seamless_cube_map[]
+    or if a sampler created with
+    ename:VK_SAMPLER_CREATE_NON_SEAMLESS_CUBE_MAP_BIT_EXT is used,
+endif::VK_EXT_non_seamless_cube_map[]
+
 +
 then the texel is a border texel and <<textures-texel-replacement,texel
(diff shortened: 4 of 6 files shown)


oddhack:
Structurally this looks OK. I will make sure to get on the agenda for the next working group meeting to get some more eyes on it. These days we use the "Template for Release Task List" issue template after publishing a new extension to create a tracking issue for various related tasks to the extension being fully implemented and supported, but I'm not sure if it's come up before in the context of an extension developed entirely outside Khronos, so will also ask what we want to do about the checklist.

zmike:
I have some interest in this functionality, so I don't mind being its advocate in WG and working with @DadSchoorse in that capacity. Please tag me on any internal issues that arise.

ShabbyX:
FYI, this can be emulated. OpenGL at least has precise rules on how to select a face, so you basically need to calculate the face to sample from, and where to sample from it, then do a 2D sampling of said face. Here's ANGLE's emulation: https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/angle/src/compiler/translator/tree_ops/RewriteCubeMapSamplersAs2DArray.cpp

It's not terribly efficient, but for the sake of D3D9 it's most likely ok. Nicer to emulate IMO than bring this old baggage to Vulkan.

DadSchoorse:
Yes, it could be emulated for D3D9, but ARB_seamless_cubemap_per_texture isn't possible without hardware support, since you don't know at compile time if you should sample seamless or non seamless.

> It's not terribly efficient, but for the sake of D3D9 it's most likely ok.

While the performance impact of emulation should not be big, D3D9 titles are gpu bound on lower end hardware. It feels wrong to sacrifice performance for correctness if there's hardware capabilities that we could use instead. And that hardware, while you could call it old baggage, is probably here to stay thanks to the OpenGL extension.

jekstrand:
I'm a little hesitant to add D3D9-only features.  If it's something that's continued on into D3D11 and 12 like transform-feedback, then there's probably pretty good reason to add it to Vulkan even if it sucks.  If it's something that died with D3D9 like non-seemless cube maps, then there's a very real possibility that the hardware support for it will eventually go away and supporting it in Vulkan will require a nasty emulation, the cost of which every Vulkan app has to pay.

The obvious counter-argument to this is, "That's fine, you can stop supporting the feature."  However, if that's a valid response, then the emulation layers still need to handle the case where the feature isn't supported.  If they're going to have to go to the effort of writing that code anyway, then my gut says we need perf data saying this actually helps over a competent emulation.

DadSchoorse:
It's not a feature that died with D3D9, non seamless is still the default for OpenGL. And emulating ARB_seamless_cubemap_per_texture (especially with bindless textures) would become very painful, so I think the hardware isn't going to go away as long as anyone cares about OpenGL.

ShabbyX:
> It's not a feature that died with D3D9, non seamless is still the default for OpenGL.

Excuse my ignorance for desktop GL features, but in GLES, non-seamless died with GLES2. When applications ask for a GLES2 context, the driver is perfectly capable of returning a GLES3+ context instead which does seamless sampling. Does desktop GL seriously still use non-seamless sampling by default (or even supports it)?

> And emulating ARB_seamless_cubemap_per_texture (especially with bindless textures) would become very painful, so I think the hardware isn't going to go away as long as anyone cares about OpenGL.

Just don't expose that extension?

DadSchoorse:
> Does desktop GL seriously still use non-seamless sampling by default (or even supports it)?

See section 8.13.1 of the OpenGL 4.6 core specification, which describes the global toggle and that seamless cube maps are initially disabled.

> Just don't expose that extension?

Fair argument, I guess. I just wanted to emphasize that while emulation in the D3D9 context is possible, it's unfeasible for OpenGL if you want to support the full set of features that modern desktop hardware supports.

jekstrand:
> Fair argument, I guess. I just wanted to emphasize that while emulation in the D3D9 context is possible, it's unfeasible for OpenGL if you want to support the full set of features that modern desktop hardware supports.

Unless you support bindless (probably not practical in Zink anyway), even the per-image version is easily enough handled via a shader key bitfield.  No, that's not great, and it gets tricky if you care about indirect sampler derefs (you probably don't, honestly) but it works and isn't massively worse than the other stuff Zink has to key off already.  You can also emit a dynamic switch in the shader based on a uniform and then have a three-value enum in the shader key which says: all seamless, all have seams, it's a mix.  Given that the vast majority of apps have at most one cube texture at a time (the skybox), this will handle 99.9% of cases without emitting a dynamic switch anywhere in the shader.

ShabbyX:
> Given that the vast majority of apps have at most one cube texture at a time (the skybox)

This is probably not accurate. I've seen engines do lighting with cube maps, so every lighting probe would be one, and you'd have one cube map array for lights too.

That said, I do agree that 99.9% of cases don't need emulation, because non-seamless cubemap sampling is just not very useful. It was more of a limitation of old APIs rather than a feature (and some old game came to depend on it because they had no choice, or someone thought they were being clever).

> > Just don't expose that extension?
> This gets a big frowny face from me. Ideally all extensions should be supportable.

I don't quite agree with that. Especially not when that extension seems just so unnecessary. If anyone wants to do non-seamless sampling in this day and age they should use 2D arrays, and any old application that might have taken advantage of that extension must necessarily have a fallback because that extension is not ubiquitous anyway (https://opengl.gpuinfo.org/listreports.php?extension=GL_ARB_seamless_cubemap_per_texture&option=not)

DadSchoorse:
> Especially not when that extension seems just so unnecessary.

Not sure this is true, because without ARB_seamless_cubemap_per_texture you are stuck with **non seamless** cube maps if you use bindless textures. The global state is ignored for those. OpenGL really got its default wrong here.

jekstrand:
> This is probably not accurate. I've seen engines do lighting with cube maps, so every lighting probe would be one, and you'd have one cube map array for lights too.

Right, so my assertion that they're using a single cube-map probably wasn't right.  However, I do suspect that very few (if any!) are using a mix.

> Not sure this is true, because without ARB_seamless_cubemap_per_texture you are stuck with non seamless cube maps if you use bindless textures. The global state is ignored for those. OpenGL really got its default wrong here.

That may be true.  However, if a layered implementation implemented bindless (and I'm not sure this is tractable on Vulkan anyway), you can probably just always do seamless and the chances of hitting an app bug are near zero.  If there is a bug, it's likely that the app forgot to set seamless when that's what they wanted all along.

I think we're seriously running into the danger of designing whole new API for theoretical use-cases at this point.  We've pretty well determined that D3D9 is implementable without this (though a bit more work).  If some super-important app comes along that uses non-seamless cubmapps with bindless on OpenGL and running it pixel-perfect on Zink is a priority, then maybe we can justify plumbing it through Vulkan.  But I suspect Vulkan will be a legacy API before that happens.  The other possible justification is if there's a perf difference that is making a game unplayable for someone.  Absent those two things, we're writing piles of spec, conformance tests, multiple driver implementations, etc. to add a feature that clearly no developer actually wants just to avoid a bit of code-gen in a D3D9 translator.  Sorry to put it in such stark terms, but that's where we are.

zmike:
After discussions in the WG, I don't have strong confidence that this extension would be implemented by any driver other than those in Mesa. It's the determination of the WG that this extension would lack the type of broad support that would make it useful for general purpose emulation, so registering it officially doesn't seem to make sense.

DadSchoorse:
> > FYI, this can be emulated. OpenGL at least has precise rules on how to select a face, so you basically need to calculate the face to sample from, and where to sample from it, then do a 2D sampling of said face.
>
> Yes, it could be emulated for D3D9

I have to retract that. While working on cube emulation in dxvk I noticed that as far as I can tell `texture(s, coord, bias)` can't be emulated correctly because bias is only allowed with OpImageSampleImplicitLod and not with OpImageSampleExplicitLod + Grad.

Emulation with `textureLod(s_array, XYZtoSTF(coord), textureQueryLod(s_cube, coord) + bias)` and  is also not valid because:
- the bias has to be clamped according to `clamp(sampler.bias+shaderOp.bias, −maxSamplerLodBias, maxSamplerLodBias)`
- it completely breaks anisotropic filtering

So at best we trade one slightly broken behavior against a different slightly broken behavior.

zmike:
17 May 2022: WG discussion has determined that this extension would be beneficial and should go ahead if the author is still interested.

pdaniell-nv (approved):
One nit, otherwise looks good. NVIDIA plans to support this.

chapters/samplers.txt
```diff
@@ -296,6 +296,12 @@ ifdef::VK_EXT_fragment_density_map[]
     If pname:flags includes ename:VK_SAMPLER_CREATE_SUBSAMPLED_BIT_EXT, then
     pname:unnormalizedCoordinates must: be ename:VK_FALSE
 endif::VK_EXT_fragment_density_map[]
+ifdef::VK_EXT_non_seamless_cube_map[]
+  * [[VUID-VkSamplerCreateInfo-flags-00000]]
```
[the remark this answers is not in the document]
DadSchoorse: Fixed, thanks. I wasn't sure what do regarding the tag.

chapters/samplers.txt
```diff
@@ -351,7 +356,14 @@ ifdef::VK_EXT_fragment_density_map[]
     specifies that the implementation may: use approximations when
     reconstructing a full color value for texture access from a subsampled
     image.
+endif::VK_EXT_fragment_density_map[]
+ifdef::VK_EXT_non_seamless_cube_map[]
+  * ename:VK_SAMPLER_CREATE_NON_SEAMLESS_CUBE_MAP_BIT_EXT
+    specifies that <<textures-cubemapedge, cube map edge handling>> is not
```
Tobski, line 362: I think this needs to be spelled out in the edge handling section too; right now it blanket says that edge handling happens for cubemaps.

Tobski (changes requested):
One minor change but mostly LGTM.

chapters/samplers.txt
DadSchoorse: I'm not sure I understand where you want to have it spelled out. I already added a section to say that border texel replacement happens for cube maps if this bit is used. So the "Otherwise the texel coordinates lie beyond the edges or corners of the selected cube map face, and Cube map edge handling is performed." doesn't apply anymore.

Tobski: You're not wrong - what's tripping me up as a reader is that the second sentence in that section calls out an exception - and this is another one. Reading that section alone it's thus not clear. I think changing this:

> If the texel coordinates lie beyond the edges or corners of the selected cube map face, the following steps are performed. Note that this does not occur when using VK_FILTER_NEAREST filtering within a mip level, since VK_FILTER_NEAREST is treated as using VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_EDGE.

to

> If the texel coordinates lie beyond the edges or corners of the selected cube map face (as described in the prior section).

Then I think it makes it substantially clearer.

A bigger rewrite is probably warranted at some point, but I don't think it'd be appropriate to ask you to do that in this PR 😬

DadSchoorse: Okay, I added that part. But I kept the note about nearest filtering, I'm not sure why that should be removed.

Tobski: The problem I have with the nearest filtering note is that it's calling out one exception when really there's a bunch of them. This often trips readers up, as they "read between the lines" by assuming that it's the only exception.

However, IMO that little addendum fixes it well enough for this extension, so I'm happy to call this resolved.

Tobski (approved):

PR #1646, opened 2021-09-24
vk.xml: Adding len attribute to VkCuModuleCreateInfoNVX struct member pData.

dave-juicelabs:
Utilizing the python scripts for code generation that was dependent on the len attribute, I found the pData member did not have a len attribute and seemed to be related to the dataSize member though I did not see any official documentation on the matter.


ewerness-nv:
Looks good.

PR #1660, opened 2021-10-11
Turn off local max-width that was overriding stylesheet.

emackey:
For what it's worth, I spend most of my time in glTF land, and know almost nothing of Vulkan.  But I've recently learned a great deal about the AsciiDoctor process that is shared between the two.

This suggestion's diff is tiny, but has a major impact.  The Khronos.css stylesheet has a `max-width` on most of the content, and the effect can be seen on the Vulkan man pages HTML, for example.  But the Vulkan HTML spec itself (the numbered chapters) did not use the max-width requested in the Khronos stylesheet, instead they include a manual style override `:max-width: 100%` inside `vkspec.txt` that caused the spec to stretch across the complete width of even the widest monitors, reducing readability for many.

So this change pulls out the offending style override, causing the spec's chapters to have an upper bound on width, as the man pages already do.  I've also pulled out another pair of `:max-width: 100` that I believe to be already inert due to their missing units.

Visually, this is a big change for the spec, so please consider carefully before merging.  If it's not needed or wanted, feel free to close it.


oddhack:
Good catch. That's been in there all the way from before initial public release, so I'm surprised nobody brought it up previously that I can recall. Agreed this is probably a good idea. We do have some very wide tables that would now require scrolling in some cases, but I don't think that's not a reason to do it. Will run it by the WG on Monday.

PR #1663, opened 2021-10-13
Register remaining newly introduced `vk_video` types in vk.xml

Friz64:
This is a follow-up to #1515, adding all the newly introduced types that are defined in `vk_video` but have not been referenced in vk.xml. I am using this information for the generation of bindings.


oddhack:
@zlatinski could you sign off on this? Seems like it fixes a straightforward oversight and will have no effect on generated headers, so I don't think we need the full Video TSG to conduct oversight.

PR #1688, opened 2021-11-16
[RFC] add VK_MESA_rect_list extension

libcg:
`VK_MESA_rect_list` adds a new RECT_LIST primitive topology to rasterize rectangles from a list of 3 axis-aligned vertices.

The main motivation is to support emulation of AMD's Mantle API (precursor to Vulkan). Battlefield 4 is one game that uses this primitive topology.

It could also be useful for general API emulation, internal blitting operations can use this topology in place of triangle strips or triangle lists. Zink could benefit from this in particular.

This is my first extension, let me know if I missed anything. Thanks.

### DIFF
--- a/appendices/VK_MESA_rect_list.txt
+++ b/appendices/VK_MESA_rect_list.txt
@@ -1,0 +1,28 @@
+// Copyright (c) 2021 Clément Guérin
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_MESA_rect_list.txt[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2021-11-15
+*Contributors*::
+  - Clément Guérin
+
+=== Description
+
+This extension adds a new elink:VkPrimitiveTopology code:enum where rectangles
+can be generated from a series of 3 axis-aligned vertices.
+This can be useful for drawing a rectangle without being split into two
+triangles with an internal edge.
+It is also useful to minimize the number of primitives that need to be
+drawn, particularly for a user interface.
+
+include::{generated}/interfaces/VK_MESA_rect_list.txt[]
+
+=== Version History
+
+  * Revision 1, 2021-11-15 (Clément Guérin)
+    - Internal revisions
--- a/chapters/drawing.txt
+++ b/chapters/drawing.txt
@@ -222,6 +222,10 @@ endif::VK_KHR_portability_subset[]
     with adjacency>>, with consecutive triangles sharing an edge.
   * ename:VK_PRIMITIVE_TOPOLOGY_PATCH_LIST specifies
     <<drawing-patch-lists,separate patch primitives>>.
+ifdef::VK_MESA_rect_list[]
+  * ename:VK_PRIMITIVE_TOPOLOGY_RECT_LIST_MESA specifies a series of
+    <<drawing-rect-lists,separate rectangle primitives>>.
+endif::VK_MESA_rect_list[]
 
 Each primitive topology, and its construction from a list of vertices, is
 described in detail below with a supporting diagram, according to the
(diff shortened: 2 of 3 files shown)


libcg:
RADV implementation: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/13810

polarina:
How does this extension differ from `VK_NV_fill_rectangle`?

appendices/VK_MESA_rect_list.txt
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2021 Clément Guérin
```
oddhack, line 1: The "(c)" is redundant and not needed, BTW. Just "Copyright <date> <author>" is fine.

oddhack:
Structurally this looks fine, although it would be helpful if you could enable Github Actions on your repository clone and pass our CI on the branch.

However, you've claimed an extension number which was reserved by @ShabbyX, so we need to make sure that's acceptable to them. If not, you'll need to reserve a new extension number following process in https://www.khronos.org/registry/vulkan/specs/1.2/styleguide.html#_registering_extensions_and_layers and update the PR once the reservation is published in our 'main' branch.

chapters/drawing.txt
```diff
@@ -710,6 +717,21 @@ The vertices comprising a patch have no implied geometry, and are used as
...
+When the primitive topology is ename:VK_PRIMITIVE_TOPOLOGY_RECT_LIST_MESA,
+each consecutive set of three vertices defines a single rectangle primitive,
+according to the equation:
+
+  {empty}:: [eq]#p~i~ = {v~3i~, v~3i+1~, v~3i+2~}#
+
+The number of primitives generated is equal to
+[eq]#{lfloor}pname:vertexCount/3{rfloor}#.
+
+They cannot be clipped, must be axis aligned, and cannot have depth gradient.
```
r-potter, line 733: These probably want to be RFC keywords (i.e. must:, cannot: etc). Also doesn't define provoking vertex behaviour

xml/vk.xml
```diff
@@ -17450,6 +17450,13 @@ typedef void <name>CAMetalLayer</name>;
                 <enum value="&quot;VK_NV_extension_433&quot;"           name="VK_NV_EXTENSION_433_EXTENSION_NAME"/>
             </require>
         </extension>
+        <extension name="VK_MESA_rect_list" number="434" type="device" author="MESA" contact="Clément Guérin @libcg" supported="vulkan">
+            <require>
+                <enum value="1"                                         name="VK_MESA_RECT_LIST_SPEC_VERSION"/>
+                <enum value="&quot;VK_MESA_rect_list&quot;"             name="VK_MESA_RECT_LIST_EXTENSION_NAME"/>
+                <enum offset="0" extends="VkPrimitiveTopology"          name="VK_PRIMITIVE_TOPOLOGY_RECT_LIST_MESA"/>
+            </require>
+        </extension>
```
r-potter, line 17459: Since the extension modifies logical device behaviour, it also needs a feature structure as per the [Feature Structures](https://github.com/KhronosGroup/Vulkan-Docs/blob/3a94dab07f8359af9130d3a344107cb422691397/style/extensions.txt#L1181) section of the style guide

zmike:
> Zink could benefit from this in particular

I'm not sure this is true? If this extension was for emulating GL_QUADS or GL_QUAD_STRIP I'd be more interested, but I'm not sure how much use I have for something like a three-point rect list.

Cherser-s:
> How does this extension differ from `VK_NV_fill_rectangle`?

1) `VK_NV_fill_rectangle` is a polygon mode, while RECT_LIST is a primitive topology.
2) RECT_LIST does require axis aligned vertices, while `VK_NV_fill_rectangle` calculates the AABB for the passed vertices on its own.

libcg:
@zmike you can use it for Gallium blit operations, similar to what radeonsi does: https://gitlab.freedesktop.org/mesa/mesa/-/blob/011ea325/src/gallium/drivers/radeonsi/si_state_draw.cpp#L2512

zmike:
I could, yes, but given that this is likely to only be supported by 1-2 drivers, I'm not sure it's worth having the extra codepath without any compelling perf numbers to motivate it.

libcg:
I did a bit more research around how other vendors handle rectangle lists, only 3 other vendors could support it:
- Nvidia could support it through the `fill_rectangle` fill mode + triangle list topology
- Intel has a RECTLIST topology, however there are additional constraints (W component must be 1.0) which might not match Mantle
- Adreno also has a RECTLIST topology, but I couldn't find documentation around it, so I assume there could be additional constraints here too

The reality is that this topology is only meant for internal driver consumption, and Mantle exposes it because it was targeting GCN only and the API doesn't expose any blit functionality. It's not yet clear if any game other than BF4 uses the topology.

In light of all this, I'm not gonna pursue this extension. However I would love to see AMD expose it as their own vendor extension so Mantle runs best on their own GPUs (I have some evidence that `VK_PRIMITIVE_TOPOLOGY_RECT_LIST_AMD` is a thing, though not public). Apologies for not doing my homework earlier.

PR #1695, opened 2021-11-18
VK_EXT_forced_multisample extension

Jonathan-Weinstein:
This extension allows rendering with a raster sample count >= 2 while single-sample color attachments are present. Essentially, it is the same as D3D's "Target Independent Rasterization" when ForcedSampleCount is >= 2 and RTV(s) are present: https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#TIR.

This feature can be used for anti-aliased path rendering.

The "raster_multisample" name was taken from GL_EXT_raster_multisample, which is a similar feature. I don't have a preference on using a different name though.

This extension should be straightforward to support in Intel hardware and probably AMD too. The Mesa Anvil Intel driver has expressed interest in implementing it. Vendors supporting VK_NV_framebuffer_mixed_samples could trivially support it.

This extension would be useful to VMware, and maybe dxvk/vkd3d too. It might also be a useful feature in general for Vulkan applications. Since TIR is required in D3D feature levels 11_1 through 12_2 (the highest as of now), I don't think this feature would end up as baggage going forward.

This PR is a work in progress, I'm missing a lot of things and I don't know how to do some stuff since it's my first extension proposal. I figured to post this as is though to see if there is any interest in the extension at all.

Thanks.

One thing I was trying to add/edit is the following validation for `VkPipelineMultisampleStateCreateInfo`, but I can't find where to put it:

---

- [VUID-VkPipelineMultisampleStateCreateInfo-flags-1(?)] If VK_EXT_raster_multisamples is not enabled, `VkPipelineMultisampleStateCreateInfo::flags` must be zero. Otherwise, it may contain `VK_PIPELINE_MULTISAMPLE_STATE_CREATE_RASTER_MULTISAMPLE_BIT_EXT`.
- [VUID-VkPipelineMultisampleStateCreateInfo-flags-2(?)] If `flags` contains `VK_PIPELINE_MULTISAMPLE_STATE_CREATE_RASTER_MULTISAMPLE_BIT_EXT`, then the `pNext` chain must not contain a `VkPipelineCoverageReductionStateCreateInfoNV` structure.


Jonathan-Weinstein:
This could a perhaps be done a different way with very few changes to the spec, and maybe not even a new extension.

What I think is needed:
- Allowing certain combinations of mixed samples + coverage reduction modes without requiring `VK_NV_framebuffer_mixed_samples` or `VK_AMD_mixed_attachment_samples` and a way to query them.
- A particular coverage reduction mode or slightly modifying `VK_COVERAGE_REDUCTION_MODE_MERGE_NV` and a way to specify it.

This could be done with `VK_NV_coverage_reduction_mode` alone by making these 3 changes to the spec:
- Add a sentence to `VK_COVERAGE_REDUCTION_MODE_MERGE_NV`'s explanation (in **bold**): "`VK_COVERAGE_REDUCTION_MODE_MERGE_NV` following specifies that each color sample will be associated with an implementation-dependent subset of samples in the pixel coverage. If any of those associated samples are covered, the color sample is covered. **For the case of a single color sample, the subset is the entire pixel**". That added part seems to be true on a both a Quadro P1000 and RTX 3070. It would also seem weird if it didn't work like that. I can't speak for Nvidia though.
- Make `VK_NV_coverage_reduction_mode` not require any extensions (it requires `VK_NV_framebuffer_mixed_samples` now).
- Also allow mixed samples for combinations returned by `vkGetPhysicalDeviceSupportedFramebufferMixedSamplesCombinationsNV`, instead of just when one of `VK_AMD_mixed_attachment_samples` or `VK_NV_framebuffer_mixed_samples` is enabled. This would change [[VUID-VkGraphicsPipelineCreateInfo-subpass-00757]] from chapters/pipelines.txt.

`VkFramebufferMixedSamplesCombinationNV::depthStencilSamples = 0` means the subpass cannot have a depth/stencil attachment. In my first draft I also required the fragment shader to not output `FragDepth` and for `depthTestEnable`, `stencilTestEnable`, and `depthBoundsTestEnable` to all be false at the time of the draw. Reason being was `GL_EXT_raster_multisample` and the D3D 11.3 functional spec for Target Independent Rasterization say similar. Maybe just requiring a null depth/stencil attachment is enough for Vulkan with its pipeline state objects.

If an implementer would like that most depth state should also be explicitly disabled, `VK_COVERAGE_REDUCTION_MODE_MERGE_NV`'s behavior for a single color sample can't generally be guaranteed, or if `VK_NV_coverage_reduction_mode` can't be changed to not require `VK_NV_framebuffer_mixed_samples`, then I think this would need a new extension. I'm assuming it's fine for AMD/Intel to implement an extension with `NV` in the name, as they do for `VK_NV_compute_shader_derivatives` on some hardware.

If a new extension is needed, it would probably require less changes than my original draft. The extension could mimic a lot of `VK_NV_coverage_reduction_mode` and use as many `typedef`s and enum value aliases as helpful. The main addition would be a new coverage reduction mode (`VK_COVERAGE_REDUCTION_MODE_MERGE_TO_ONE_RESTRICED_EXT = 2` or something). Changes to the spec would be almost the same as the 3 proposed changes using `VK_NV_coverage_reduction_mode` alone. This might be preferable than the original `VK_PIPELINE_MULTISAMPLE_STATE_CREATE_RASTER_MULTISAMPLE_BIT_EXT` that circumvented `VK_NV_coverage_reduction_mode` in maybe a weird way.

Jonathan-Weinstein:
I renamed this from VK_EXT_raster_multisample to VK_EXT_forced_multisample to be similar to D3D's ForcedSampleCount, which is also the name used in Intel graphics documents (like in 3DSTATE_RASTER_BODY here: https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-icllp-vol02d-commandreference-structures_0.pdf).

The latest draft in my forked repo built with no errors and looked fine to me using `makeSpec -clean -spec all html`. I'm unsure where VUID numbers are supposed to come from though and just picked unused ones at random.

I'm closing this because I'm not sure I how much I care anymore, if anyone else wants an extension like this, and I realized I need to make a few changes for VK_KHR_dynamic_rendering. Maybe I'll reopen this or make a new PR that is clearer in the future though.

I still think the idea behind this extension is good. The feature is not complex and D3D 11_1+ requires it, so it should be well supported in desktop hardware going forward. I was also able to implement the original proposal (that had the MSAA state create info flag) quite easily in the Mesa ANV driver. There aren't any good ways to emulate this feature, one of the better ones may be to use VK_EXT_conservative_rasterization and insert a geometry shader that passes barycentric coordinates to the FS. The FS then constructs what the input coverage mask would be doing `interpolateAtOffset(baryCoord, sampleOffset[i])` `ForcedSampleCount`-number of times; if all components are non-negative, coverage bit `i` is set. The geometry shader would make things less efficient and the application would need code to generate all the variants they may need. It's much more ideal to just set a few bits in a hardware register.

oddhack:
VUIDs are assigned after an extension is merged into main branch. For future reference, just write VU statements without VUIDs, per https://www.khronos.org/registry/vulkan/specs/1.2/styleguide.html#appendix-vuid-creating :

> Since these tags are of use only to the published validation layer, they are needed only in the published Specification sources and outputs. Therefore, authors of extensions, or other branches adding valid usage statements, are not themselves responsible for adding tags in their branches. The specification editors will take care of this as part of the process of publishing updates. For reference purposes, this process is described below:

PR #1698, opened 2021-11-19
Minor changes about separate depth or stencil readonly image layout

Richard-Yunchao:
Separate depth or stencil readonly image layouts only exist in
Vulkan 1.1+ or via extension. It should not apprear at VUID 03280
and 03282 in 1.0.

The impacted image layouts are:
  - VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL
  - VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL

### DIFF
--- a/chapters/renderpass.txt
+++ b/chapters/renderpass.txt
@@ -1611,19 +1611,23 @@ attachments on multiple uses via ename:VK_ATTACHMENT_LOAD_OP_CLEAR.
     ename:VK_IMAGE_LAYOUT_PREINITIALIZED
   * [[VUID-VkAttachmentDescription-format-03280]]
     If pname:format is a color format, pname:initialLayout must: not be
-    ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL,
+ifdef::VK_VERSION_1_1,VK_KHR_maintenance2[]
+    ename:VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL,
+    ename:VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL,
+endif::VK_VERSION_1_1,VK_KHR_maintenance2[]
+    ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL, or
     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL,
-    ename:VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, or
-    ename:VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL
   * [[VUID-VkAttachmentDescription-format-03281]]
     If pname:format is a depth/stencil format, pname:initialLayout must: not
     be ename:VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL
   * [[VUID-VkAttachmentDescription-format-03282]]
     If pname:format is a color format, pname:finalLayout must: not be
-    ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL,
+ifdef::VK_VERSION_1_1,VK_KHR_maintenance2[]
+    ename:VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL,
+    ename:VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL,
+endif::VK_VERSION_1_1,VK_KHR_maintenance2[]
+    ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL, or
     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL,
-    ename:VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, or
-    ename:VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL
   * [[VUID-VkAttachmentDescription-format-03283]]
     If pname:format is a depth/stencil format, pname:finalLayout must: not
     be ename:VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL
--- a/chapters/renderpass.txt
+++ b/chapters/renderpass.txt
@@ -1609,37 +1609,35 @@ attachments on multiple uses via ename:VK_ATTACHMENT_LOAD_OP_CLEAR.
   * [[VUID-VkAttachmentDescription-finalLayout-00843]]
     pname:finalLayout must: not be ename:VK_IMAGE_LAYOUT_UNDEFINED or
     ename:VK_IMAGE_LAYOUT_PREINITIALIZED
-ifndef::VK_VERSION_1_1,VK_VERSION_1_2,VK_KHR_maintenance2[]
+ifndef::VK_VERSION_1_1,VK_KHR_maintenance2[]
   * [[VUID-VkAttachmentDescription-format-03280]]
     If pname:format is a color format, pname:initialLayout must: not be
     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL or
     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL
-endif::VK_VERSION_1_1,VK_VERSION_1_2,VK_KHR_maintenance2[]
-ifdef::VK_VERSION_1_1,VK_VERSION_1_2,VK_KHR_maintenance2[]
-  * [[VUID-VkAttachmentDescription-format-03280]]
-    If pname:format is a color format, pname:initialLayout must: not be
+endif::VK_VERSION_1_1,VK_KHR_maintenance2[]
+ifdef::VK_VERSION_1_1,VK_KHR_maintenance2[]
+  * If pname:format is a color format, pname:initialLayout must: not be
     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL,
     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL,
     ename:VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, or
     ename:VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL
-endif::VK_VERSION_1_1,VK_VERSION_1_2,VK_KHR_maintenance2[]
+endif::VK_VERSION_1_1,VK_KHR_maintenance2[]
   * [[VUID-VkAttachmentDescription-format-03281]]
     If pname:format is a depth/stencil format, pname:initialLayout must: not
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
These two enums are not listed in VkImageLayout in Vulkan 1.0: https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VkImageLayout. That is correct. However, they are mentioned in VUID 03280 and 03282 in latest 1.0 spec, which is incorrect.

oddhack:
@Richard-Yunchao our toolchain does not support asciidoctor conditionals imbedded inside VUs, unfortunately. We'd like to resolve this by create a new, parallel VU for the enums in question, and placing the conditionals surrounding it. So the first VU would turn into

``` 
  * If pname:format is a color format, pname:initialLayout must: not be    
    ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL or
    ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL,   
ifdef::VK_VERSION_1_1,VK_KHR_maintenance2[]
  * If pname:format is a color format, pname:initialLayout must: not be
    ename:VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, or
    ename:VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL
endif::VK_VERSION_1_1,VK_KHR_maintenance2[]
```

and similarly for the other. If you want to update the PR accordingly that would be great, but if  it's not entirely clear how we're doing this, reassign the PR to me and I'll create a new one to accomplish this. Thanks!

Richard-Yunchao:
> @Richard-Yunchao our toolchain does not support asciidoctor conditionals imbedded inside VUs, unfortunately. We'd like to resolve this by create a new, parallel VU for the enums in question, and placing the conditionals surrounding it. So the first VU would turn into
> 
> ```
>   * If pname:format is a color format, pname:initialLayout must: not be    
>     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL or
>     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL,   
> ifdef::VK_VERSION_1_1,VK_KHR_maintenance2[]
>   * If pname:format is a color format, pname:initialLayout must: not be
>     ename:VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, or
>     ename:VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL
> endif::VK_VERSION_1_1,VK_KHR_maintenance2[]
> ```
> 
> and similarly for the other. If you want to update the PR accordingly that would be great, but if it's not entirely clear how we're doing this, reassign the PR to me and I'll create a new one to accomplish this. Thanks!

@oddhack , I actually followed the instructions about how to generate .hmtl (https://github.com/KhronosGroup/Vulkan-Docs/blob/main/BUILD.adoc#building) and I tried my PR at my local machine. I was working. So I submitted this PR. I even verified again right now, it works. So, are there any changes for online auto-gen tool against the tool on the instruction webpage?

oddhack:
> @oddhack , I actually followed the instructions about how to generate .hmtl (https://github.com/KhronosGroup/Vulkan-Docs/blob/main/BUILD.adoc#building) and I tried my PR at my local machine. I was working. So I submitted this PR. I even verified again right now, it works. So, are there any changes for online auto-gen tool against the tool on the instruction webpage?

The HTML output is fine, but the validusage target will truncate the VU at the point of the imbedded ifdef, which is why we don't support this usage.

Richard-Yunchao:
@oddhack , I'd like to discuss this issue on my PR (sorry to be late to update it because of Thanksgiving holiday). Please take a look if you have time.

chapters/renderpass.txt
```diff
@@ -1609,21 +1609,37 @@ attachments on multiple uses via ename:VK_ATTACHMENT_LOAD_OP_CLEAR.
...
     pname:finalLayout must: not be ename:VK_IMAGE_LAYOUT_UNDEFINED or
     ename:VK_IMAGE_LAYOUT_PREINITIALIZED
+ifndef::VK_VERSION_1_1,VK_VERSION_1_2,VK_KHR_maintenance2[]
+  * [[VUID-VkAttachmentDescription-format-03280]]
+    If pname:format is a color format, pname:initialLayout must: not be
+    ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL or
+    ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL
+endif::VK_VERSION_1_1,VK_VERSION_1_2,VK_KHR_maintenance2[]
+ifdef::VK_VERSION_1_1,VK_VERSION_1_2,VK_KHR_maintenance2[]
   * [[VUID-VkAttachmentDescription-format-03280]]
```
Richard-Yunchao, line 1619: Can we use the same VUID for the block ifndef-endif-ifdef-endif? Looks like it is working. 

I noticed your replacement, it is a little ugly because it will have some text (like `If pname:format is a color format, pname:initialLayout must: not be`) twice. And that text is a bulletin without VUID. And I also noticed a following PR from someone set a VUID-TODO for it. 

If this solution is OK. It resolved issues in those two PRs. 

BTW, I also added VK_VERSION_1_2 in ifs, otherwise, it is incorrect for VK 1.2 specification. But this turned out that if we have VK 1.3 or VK 2.0, all these following versions need to be added into the list. Is there a solution like if VK_VERSION_1_1+, which means VK 1.1 and all versions above?

oddhack: What you originally did with this PR was to recognize that a couple of the enums in the VUs were not valid without the extension and/or version that introduced those enums defined, and that was a good observation. I re-did the PR in #1708 to account for the limitations of how the validusage statement extractor actually works. The changes you've added to this PR since then are not consistent with how our process works, and we can't accept those. They don't prevent generating a spec but they will cause problems with actually using the VUs in the validation layers process downstream.

 I believe #1708 captures the intent of this PR and would like to go with that one, unless you see something actually wrong (in contrast to appearing inelegant or wordy) with it.

> Can we use the same VUID for the block ifndef-endif-ifdef-endif? Looks like it is working.

No, that's not OK because they are two different valid usage conditions being tested. This doesn't affect generating the spec outputs but it will be a problem for validation layer implementation.

> I noticed your replacement, it is a little ugly because it will have some text (like If pname:format is a color format, pname:initialLayout must: not be) twice.

Yes, it's unfortunate that there is some repetition. But as I commented previously, the valid usage extractor has restrictions that force this mild ugliness on us.

>  And that text is a bulletin without VUID.

In the PR, it does not have a VUID yet. VUIDs are assigned to new VU statements as part of the public spec update process, which is the same point at which the valid usage database validusage.json is generated for consumption by the validation layers. This is a documented part of the process of managing VU statements.

> BTW, I also added VK_VERSION_1_2 in ifs, otherwise, it is incorrect for VK 1.2 specification.

What the original conditional was saying was that if you build a spec with VK_VERSION_1_1 defined, or without VK_VERSION_1_1 but with VK_KHR_maintenance2 (e.g Vulkan 1.0 + maintenance2 extension), then the ifdefed content will be included. VK_VERSION_1_2 by definition requires VK_VERSION_1_1. We will never generate a spec where this is not true. The ifdef needs to contain only the extensions and/or versions where the content is actually first defined.

Richard-Yunchao: @oddhack , I think as a senior folk and a reviewer as you in Vulkan community, it might make more sense to give insights and suggestions on the other two issues at #1700 and #1699, and I really appreciate it if you can do that. They are much more difficult and need real discussions. While polishing this PR is really trivial. And it's my duty.  

For the trivial stuff like fulfilling the requirement of the autogen/infrastructure for Vk spec, I think you can guide me to do that. And that is what a reviewer should do. According to my experience, the reviewer should not submit a new PR/change to compete with the original author unless the original author meet significant hard-to-resolve problem by himself/herself or the original author disappears and doesn't response to the thread. Otherwise, if the reviewer see some problem, he just submits a replacement and ask the original author to close his original PR, I feel like it is not a friendly way to promote this project. 

Regarding the VUID stuff, thanks for your explanation! I revised this PR accordingly. It can handle the ugly repetition problem. And we can assign new VUID afterwards. 

Regarding the VK_VERSION_1_2 problem, it turned out that I used `incorrect` way to generate the html. I tried `./makeSpec -spec core html VERSIONS="VK_VERSION_1_2" `. It should be
`./makeSpec -spec core html VERSIONS="VK_VERSION_1_0 VK_VERSION_1_1 VK_VERSION_1_2" ` or just `./makeSpec -spec core html` for now. Likewise, if we want to generate VK 1.1, we need to add VK 1.0 on the list. You see, if Vulkan gets more and more versions, we need to add all previous versions to generate a particular version spec. That's a little silly. I think the autogen tool can expand versions for a particular version via script. However, I understand that the infrastructure to autogen of Vk spec is only used by a very small amount of people, so it is up to you guys whether to do that or not. 

Once again, I really appreciate if you can comment on issue 1700 and 1699, Jon. I really want to hear the thoughts from senior folks like you in Vulkan community.

chapters/renderpass.txt
```diff
@@ -1609,21 +1609,35 @@ attachments on multiple uses via ename:VK_ATTACHMENT_LOAD_OP_CLEAR.
   * [[VUID-VkAttachmentDescription-finalLayout-00843]]
     pname:finalLayout must: not be ename:VK_IMAGE_LAYOUT_UNDEFINED or
     ename:VK_IMAGE_LAYOUT_PREINITIALIZED
+ifndef::VK_VERSION_1_1,VK_KHR_maintenance2[]
```
oddhack, line 1612: I'm unclear why you added these #ifndefs. AFAIK the desired end state is that none of these 4 enums are valid layouts for a color format. Because these two enums are always defined while the other two are introduced by maintenance4 / VK_VERSION_1_1, it's appropriate to have ifdefs around the latter two, but not to have ifndefs around the first two.

chapters/renderpass.txt
oddhack: I am senior in terms of length of time involved, but that is far from making me an expert on everything in the specification. #1699 has several responses from WG members who are more familiar with it already, and #1700 is being discussed internally by the Working Group and will have a reply posted when there is more information from IHVs available.

There's no problem accepting this PR once it gets to the right place in terms of the markup, but it is not quite there yet. I had not heard back for nearly a week prior to posting #1708 so made that to move things along, but I'm glad you've been able to engage again here.

Richard-Yunchao: Understood. Vulkan is so complicated. It is virtually impossible for a single person to know every details even as senior as you are in the community. And I really thank you for your feedback about 1700 and 1699. Regarding 1699, I will do more investigation about the implementation in the engine I am working on. I am not convinced. And I will reply on that issue. But I am glad to know that you folks have active follow-ups on these issues, especially for issue 1700. 

Regarding the delay, I am sorry for that. It was Thanksgiving holiday, and I had a couple more day off last week. But I will definitely work on this PR.

chapters/renderpass.txt
Richard-Yunchao: I understand you point here. Your idea is from the perspective of this project itself. But I pay more attention on the html file generated. Because the project itself is just like a tool to generate Vulkan spec in html. The former is used by a very small amount of experts, while the latter is used by much more developers. 

And the ifndef-endif-ifdef-endif structure will make it cleaner for the html generated. No subsequent repetition text like `If format is a color format, initialLayout must not be`. 

And this method (ifndef-endif-ifdef-endif) is applied even in current project itself, for example: 
  - VUID 03062 and 04953, 
  - VUID 00896 and 01758. 

However, I agree that Vulkan and Vulkan spec are very verbose (intentionally?). Repetitions/duplications are here and there in its spec (html file). For example, the `initialLayout` and `finalLayout` in this PR can be combined if they are in other specifications. But Vulkan spec doesn't do this. It even divide them into many scattered UVIDs respectively. So I followed your suggestion and didn't apply ifndef-endif-ifdef-endif block. PTAL. Thank you!

Richard-Yunchao:
Thank you very much, @oddhack !

PR #1699, opened 2021-11-20
Revise image layout matching rules

Richard-Yunchao:
In the depth/stencil readonly image layout definition, it says that
VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL specifies a layout
for both the depth and stencil aspects of a depth/stencil format
image allowing read only access as a depth/stencil attachment or
in shaders as a sampled image.

So we should say that VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL is
compatible with depth/stencil readonly image layout.

### DIFF
--- a/chapters/resources.txt
+++ b/chapters/resources.txt
@@ -3158,12 +3158,16 @@ performed through descriptors referring to only a single aspect of the
 image, where the following relaxed matching rules apply:
 
   * Descriptors referring just to the depth aspect of a depth/stencil image
-    only need to match in the image layout of the depth aspect, thus
+    only need to match in the image layout of the depth aspect. It is also
+    allowed to sample from the image in shader. Thus
+    ename:VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL,
     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL and
     ename:VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL are
     considered to match.
   * Descriptors referring just to the stencil aspect of a depth/stencil
-    image only need to match in the image layout of the stencil aspect, thus
+    image only need to match in the image layout of the stencil aspect.
+    It is also allowed to sample from the image in shader. Thus
+    ename:VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL,
     ename:VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL and
     ename:VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL are
     considered to match


Richard-Yunchao:
(I'd like to discuss this via a PR)
I met this problem when I was using a readonly depth/stencil attachment in render pass. 

In the render pass and subpass, its layout is depth/stencil readonly optimal. However, if we sample from the depth/stencil image in shader, the corresponding binding in descriptor set DO NOT know whether the depth/stencil image is used as depth/stencil attachment. So, it is reasonable to set its layout to VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL. Then it will lead to validation error because the depth/stencil texture has 2 incompatible layout according to current layout matching rules: one is shader readonly, another is depth/stencil readonly. 

I have 2 solutions (or workarounds) for this problem, but neither of them are good enough:
1) Set the layout to GENERAL for both layouts in descriptor set and render-pass/subpass. This is not good because depth/stencil readonly layout like VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL intends to specifies a layout for readonly access from both depth/stencil attachment and sampling from shaders. Now I have to use GENERAL for these two readonly access. 
2) Set binding's layout in descriptor set to depth/stencil readonly optimal blindly, no matter the depth/stencil image is used as readonly attachment or writable attachment, or not used as attachment at all. This is bad too. Because I need to add depth/stencil attachment usage flag like VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT blindly, even though I just copy data into depth/stencil image and sampled from it. Otherwise, it will lead to validation error because the depth/stencil readonly layout requires depth/stencil attachment image usage bit, even though the sampled texture may not be used as depth/stencil attachment at all in render pass.   

So, I feel like the best solution should be say that VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL layout is compatible with depth/stencil readonly layout for depth/stencil image. And the new added compatibility falls into the real intention of depth/stencil readonly layout when we define them, per my understanding. Then, if the depth/stencil image is used as depth/stencil readonly attachment, it is compatible. It is fine. If the image is used as writable attachment, the validation can also catch the error because they are not compatible. If the image is not used as attachment at all, it is also good and I may not need to add VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT during image creation.

jekstrand:
> So, I feel like the best solution should be say that VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL layout is compatible with depth/stencil readonly layout for depth/stencil image.

Why not make your layer or engine or whatever look at the image format and always use `VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL` for depth/stencil images when you would normally use `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL`?  This is what DXVK and other layered solutions do.

Richard-Yunchao:
> > So, I feel like the best solution should be say that VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL layout is compatible with depth/stencil readonly layout for depth/stencil image.
> 
> Why not make your layer or engine or whatever look at the image format and always use `VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL` for depth/stencil images when you would normally use `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL`? This is what DXVK and other layered solutions do.

Thanks for your input, @jekstrand . 

That's what exactly I am doing in the engine involved right now. And it is indeed a doable workaround and I also mentioned that workaround at 2) above. It is a little weird and it will introduce one more flaw (or whatever you call it): If I always blindly use `VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL` when sampling from a depth/stencil image,  I need to add depth/stencil attachment usage flag `VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT` blindly in order to not break validation rule, even though I just copy data into depth/stencil image and sample from it but never use it as `depth/stencil attachment` in render pass. 

So I feel like that make `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL` and `VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL` compatible in the `image layout matching rule` is a more elegant solution, only if there is no hardware limitation.

HansKristian-Work:
If you didn't use VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT, then you are not allowed to use that a depth-stencil attachment to begin with. You are allowed to use SHADER_READ_ONLY layout, but only outside render passes.

Richard-Yunchao:
> If you didn't use VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT, then you are not allowed to use that as a depth-stencil attachment to begin with. You are allowed to use SHADER_READ_ONLY layout, but only outside render passes.

> The layered implementation will generally let you know if the resource can be used as a depth-stencil attachment or not (in D3D12, e.g. ALLOW_DEPTH_STENCIL resource flag), so you can settle on the read-only format at resource creation time. It works fine for us in vkd3d-proton.

@HansKristian-Work,  I met these 2 situations in my engine (not sure whether they are the same with yours): 
1) a. copy data into a depth/stencil image, 
    b. sample that depth/stencil image and do something in a shader, 
    c. start a pipeline, and a render pass to use that shader in step b to do something. It it likely that there is a depth/stencil attachment in the render pass (readonly or not, doesn't matter), but it is using another depth/stencil image. 
2) a. depth/stencil image is initialized via copy or rendering (eg. for shadow, etc), 
    b. sample that depth/stencil image and do something in a shader, 
    c. enable readonly depth/stencil attachment for depth test upon the same depth/stencil image in render pass, and use the shader in step b and do something. 

I think it is allowed to use SHADER_READ_ONLY layout in situation 1.b for the depth/stencil image, and it is used inside a render pass later. 
For 1.b and 2.b when I am creating a depth/stencil image and shader and its bindings, I feel like I don't know whether the same image will be used as readonly attachment or not (or it may be used and reused in both kind of render pass) in the upcoming render pass(es) in my engine, especially when the shader/image/pipeline vs attachment/renderpass are created by different module/people/etc. As a result, I am forced to set depth/stencil image in both situations to DEPTH_STENCIL_READ_ONLY image layout, and with DEPTH_STENCIL_ATTACHMENT_BIT image usage in order to make validation happy, even though the image in situation 1 is not used as attachment at all and it doesn't need DEPTH_STENCIL_ATTACHMENT_BIT image usage.

HansKristian-Work:
> As a result, I am forced to set depth/stencil image in both situations to DEPTH_STENCIL_READ_ONLY image layout, and with DEPTH_STENCIL_ATTACHMENT_BIT image usage in order to make validation happy, even though the image in situation 1 is not used as attachment at all and it doesn't need DEPTH_STENCIL_ATTACHMENT_BIT image usage

I don't really understand the problem here. As an application you need to know if your image can be used as a DEPTH_STENCIL attachment or not at some point.

For 1b, you are allowed to use SHADER_READ_ONLY. If that image is later used in a render pass, you'd need to transition SHADER_READ_ONLY to DEPTH_STENCIL_READ_ONLY, unless you did 1b in DEPTH_STENCIL_READ_ONLY already. Since you need to know up-front if the resource is capable of depth-stencil rendering, it's fairly easy to use the appropriate read-only layout.

> I feel like I don't know whether the same image will be used as readonly attachment or not (or it may be used and reused in both kind of render pass) in the upcoming render pass(es) in my engine

This is normal, but based on the image usage flags, you can deduce a useful image layout as mentioned. I don't see why you'd be forced to add VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT.

TomOlson:
The working group has discussed this issue a couple of times. We aren't able to make a retroactive change to compatibility rules, since we cannot guarantee that drivers in the field will be updated. An extension would have the same problem - you would need a workaround anyway for implementations that don't have it. We think @HansKristian-Work and @jekstrand's recommended workarounds are the way to go. I'm afraid that's the best we can do unless we hear of use cases where the workarounds can't be used and there would be value in having an extension that isn't available everywhere.

oddhack:
@Richard-Yunchao - closing based on @TomOlson's reply above and lack of followup. If you do know of cases where these workarounds can't be used, you can reopen.

PR #1708, opened 2021-11-29
Refactor #1698 to avoid conditionals inside new valid usage statements.

oddhack:
@Richard-Yunchao this refactors your change in #1698 as I described there. Please close #1698 if this will work as a replacement.


oddhack:
Closed in favor of #1698.

PR #1711, opened 2021-12-01
Correct interpolation equation rendering in primsrast

martty:


willdla:
hi there what is f stands for here, in my understanding f can stands for normal or z or anything else that need interpolation of the fragment, is that right?

oddhack:
> hi there what is f stands for here, in my understanding f can stands for normal or z or anything else that need interpolation of the fragment, is that right?

The previous paragraph tries to describe that f is a placeholder for shader outputs or the w coordinate, which are all interpolated in the same fashion. Are you asking for additional clarification?

PR #1729, opened 2022-01-04
Add note for Linear colorspace

stonesthrow:
for a linear or non-gamma transfer function color space use the VK_COLOR_SPACE_PASS_THROUGH_EXT


stonesthrow:
Reference: https://gitlab.khronos.org/vulkan/vulkan/-/issues/2827
"What is the equivalent of `EGL_GL_COLORSPACE_LINEAR` in `VkColorSpaceKHR`?"

chapters/VK_KHR_surface/wsi.txt
```diff
@@ -1061,6 +1061,13 @@ This has been updated to indicate that it uses RGB color encoding, not XYZ.
 The old name is deprecated but is maintained for backwards compatibility.
 ====
 
+[NOTE]
+.Note
+====
+For a linear or non-gamma transfer function color space use
+VK_COLOR_SPACE_PASS_THROUGH_EXT
```
gnl21, line 1068: I missed the discussion on this, so apologies if this has already been discussed, but I think it would be better to say, for example
> For a linear or non-gamma transfer function color space with unknown color primaries use ...

because there are other linear transfer function color spaces that would not use PASS_THROUGH like ADOBERGB_LINEAR.

stonesthrow: We had an initial discussion with some of our experts in formats. But we haven't ended our discussion. This was the first proposal, with expectation of further feedback.
You have a good point, I will bring this us at next discussion. Thank you

stonesthrow:
Needs thumbs up or comments to fix. I'll merge if nothing by Wednesday

krOoze:
- I don't immediately understand what the note means.
 - Equally not sure what is the added value over what `VK_COLOR_SPACE_PASS_THROUGH_EXT` text already says.
 - It no longer seems to cover the linear case and the issue about `EGL_GL_COLORSPACE_LINEAR`
 - Seems brittle language if some non-gamma based enum is added in the future (haven't checked if such enum already doesn't exist)

stonesthrow:
Agreed Krooze, the wording got away from intent - to provide guidance for those looking for a "linear" colorspace.

Propose text:
"For traditional "Linear" or non-gamma transfer function color space use VK_COLOR_SPACE_PASS_THROUGH_EXT."

stonesthrow:
Please review new text. Basically returned to original.

stonesthrow:
@chadversary
The commits can be squashed with merge:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges

chadversary:
lgtm

I believe the GitHub UI does not provide the needed button to squash these commits. Because the branch's commits are interleaved with merges from main, the branch must first be rebased onto main before squashing the commits. But, iiuc, GitHub offers two relevant buttons to the person who does the merge: *Rebase and Merge* and *Squash and Merge*; but GitHub iiuc does not provide the combined button, *Rebase and Squash and Merge*.

So, I'm ok with this branch getting merged either way, squashed or not.

For reference, here is how you would do it on the cmdline. First, run `git rebase -i main`. Git will open a todo list in an editor (default is `vi`). For each commit except the first, replace the action `pick` with `squash`.  Save and exit the editor. Finally, force-push the branch: `git push --force stonesthrow`.

chadversary:
@oddhack Could you give me permission in this project to change milestones? I believe this MR is ready for `Signed-off to Merge`.

oddhack:
> I believe the GitHub UI does not provide the needed button to squash these commits. Because the branch's commits are interleaved with merges from main, the branch must first be rebased onto main before squashing the commits.

'Squash and Merge' seems to do something reasonable IME. The result is a one commit/PR history with the end result being correct, which is what (at least I as the release manager) want to happen. No harm in doing it yourself in the source branch as well.

PR #1795, opened 2022-03-10
Using different (less confusing) variable names to describe VkCompareOp

johannesugb:
Found myself being confused by the used variable names `A` and `B`, where further below it is stated that `A`  corresponds to `z_f` (_not_ to `z_a`!) and `B` corresponds to `z_a`. 

`A` and `B` are just some variable names -- why not use different variable names, say, `F` and `A`?! Then the "f" variables would always refer to fragments, and the "a" variables would always refer to the (depth) attachment.

### DIFF
--- a/chapters/fragops.txt
+++ b/chapters/fragops.txt
@@ -1211,16 +1211,16 @@ stencil comparison function, are:
 include::{generated}/api/enums/VkCompareOp.txt[]
 
   * ename:VK_COMPARE_OP_NEVER specifies that the test evaluates to false.
-  * ename:VK_COMPARE_OP_LESS specifies that the test evaluates [eq]#A < B#.
-  * ename:VK_COMPARE_OP_EQUAL specifies that the test evaluates [eq]#A = B#.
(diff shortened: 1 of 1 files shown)


krOoze:
- "Meaningful letter" is marginally better, but I am not quite happy with single letter names in general. Should perhaps introduce proper names for the concept. I would propose perhaps "reference value" and "tested value".
 - It is still mildly confusing when the point of entry into the spec is at the `VkCompareOp`. The F and A (resp. better names above) should be accompanied with what they mean at this place, perhaps this way:
> The sources of the values for {F} and {A} are specified elsewhere in the specification for each kind of test that refers to `VkCompareOp`.
 - It seems {A} and {B} are defined for Depth Test and Stencil Test, but similar is missing for Texel Imput Depth Compare Operation. This should be added.
- Your PR changed it for Depth Test, but forgot to change it for Stencil Test (at the chapter of the same name)

oddhack:
@johannesugb please take a look at #1805 - I've tried to extend your idea to cover all the places VkCompareOp is used and be consistent between those uses. If you agree its an improvement, please close this. Thanks!

PR #1803, opened 2022-03-23
Fix missing structextends=VkRenderPassCreateInfo2 cases in XML

Ralith:
These are (presumably) intended to be useable through the newer `VkRenderPassCreateInfo2` interface, but that was not expressed in the XML.


Tobski:
Hi @Ralith - this actually is not the intended use of these structs, the omission was deliberate. The parameters for these structs were folded into VkRenderPassCreateInfo2 and its substructs so as to avoid multiple parallel arrays. The parameters for VkRenderPassMultiviewCreateInfo are mapped as follows:

- pViewMasks is now VkSubpassDescription2::viewMask
- pViewOffsets is now VkSubpassDependency2::viewOffset
- correlationMaskCount and pCorrelationMasks are now VkRenderPassCreateInfo2::correlatedViewMaskCount and VkRenderPassCreateInfo2::pCorrelatedViewMasks.

For VkRenderPassInputAttachmentAspectCreateInfo, pAspectReferences are captured in VkAttachmentReference2::aspectMask.

Hope this clears the interactions up.

Tobski:
(Also thank you for your contribution!)

PR #1805, opened 2022-03-28
Improve description and references to VkCompareOp

oddhack:
This is a more comprehensive replacement for #1795

It moves the description of elink:VkCompareOp following its first use in
slink:VkSamplerCreateInfo, and changes the way arguments of the compare
operator are described.

The following sections are updated to follow the new naming scheme for
VkCompareOp, and to be more consistent with each other:

* slink:VkSamplerCreateInfo
* slink:VkPipelineDepthStencilStateCreateInfo
* <<textures-depth-compare-operation, Depth Compare Operation>>
* <<fragops-stencil, Stencil Test>>
* <<fragops-depth-comparison, Depth Comparison>>
* flink:vkCmdSetDepthCompareOp


chapters/samplers.txt
```diff
@@ -478,6 +478,51 @@ ifdef::VK_VERSION_1_2[<<features-samplerMirrorClampToEdge,samplerMirrorClampToEd
 endif::VK_VERSION_1_2,VK_KHR_sampler_mirror_clamp_to_edge[]
 --
 
+[open,refpage='VkCompareOp',desc='Comparison operator for depth, stencil, and sampler operations',type='enums']
```
krOoze, line 481: While technically first occurence here, I would prefer it below `VkPipelineDepthStencilStateCreateInfo`, which is more mainstream usage of the enum.

chapters/samplers.txt
```diff
@@ -478,6 +478,51 @@ ifdef::VK_VERSION_1_2[<<features-samplerMirrorClampToEdge,samplerMirrorClampToEd
...
+--
+_Comparison operators_ compare a _reference_ and a _target_ value, and
+return a true or false value depending on the comparison operator chosen.
```
krOoze, line 484: Esthetically, would prefer "test passed" and "test failed" nomenclature rather than raw bools.

Otherwise should replace references to "test" below.

chapters/samplers.txt
```diff
@@ -478,6 +478,51 @@ ifdef::VK_VERSION_1_2[<<features-samplerMirrorClampToEdge,samplerMirrorClampToEd
...
+    [eq]#_reference_ {leq} _target_#.
+  * ename:VK_COMPARE_OP_GREATER specifies that the test evaluates
+    [eq]#_reference_ > _target_#.
+  * ename:VK_COMPARE_OP_NOT_EQUAL specifies that the test evaluates
+    [eq]#_reference_ {neq} _target_#.
+  * ename:VK_COMPARE_OP_GREATER_OR_EQUAL specifies that the test evaluates
+    [eq]#_reference_ {geq} _target_#.
+  * ename:VK_COMPARE_OP_ALWAYS specifies that the test evaluates to true.
+
+Comparison operators are used for:
```
krOoze, line 504: With such a nice navigation map, it is a question if not move it to Common Object Types chapter.

oddhack:
@jeffbolznv @tobski will review before signing off.

chapters/fragops.txt
```diff
@@ -955,22 +956,24 @@ fragments generated by <<primsrast-polygons-basic,back-facing>>
...
 any other primitives.
 
-The comparison performed is based on the elink:VkCompareOp, compare mask
-[eq]#s~c~# , and stencil reference value [eq]#s~r~# of the relevant state
-set.
-The compare mask and stencil reference value are set by either the
+The comparison operation performed is determined by a elink:VkCompareOp
+value set by
+ifdef::VK_VERSION_1_3,VK_EXT_extended_dynamic_state[]
+flink:vkCmdSetDepthCompareOp::pname:depthCompareOp, or by
```
jeffbolznv, line 962: I don't think I've seen this "flink:A::pname:B" usage before (only with slink). But if it renders something reasonable, then no objection.

chapters/textures.txt
```diff
@@ -839,33 +839,16 @@ sparse unbound texel is undefined:.
...
-D & = 0.0 & \text{otherwise}
-\end{aligned}
-+++++++++++++++++++
-
-where [eq]#D~tex~# is the texel depth value and [eq]#D~ref~# is the
-reference value from the SPIR-V operand.
-If the image being sampled has a fixed-point format then the reference value
-is clamped to [0, 1] before the comparison operation.
+The value of the result, [eq]#D#, is [eq]#1.0# if the result of the
+comparison is [eq]#true#, and [eq]#0.0# otherwise.
```
krOoze, line 843: Would make it a separate paragraph after the following two sentences.

Defining `D` is meaningless now, or where is it invoked?

chapters/textures.txt
```diff
@@ -839,33 +839,16 @@ sparse unbound texel is undefined:.
...
+The compare operation is selected by the elink:VkCompareOp value set by
+slink:VkSamplerCreateInfo::pname:compareOp.
+The _reference_ and _target_ arguments in the operation specified by the
```
krOoze, line 846: Previously called "values" not "arguments".

chapters/textures.txt
```diff
@@ -839,33 +839,16 @@ sparse unbound texel is undefined:.
...
+elink:VkCompareOp are the texel depth value [eq]#D~tex~# and the reference
+value from the SPIR-V operand [eq]#D~ref~#, respectively.
+
+If the image being sampled has a fixed-point format, then [eq]#D~ref~# is
```
krOoze, line 850: Should precisely say unsigned normalized fixed-point format.

jeffbolznv (approved):
Just one nitpick, approving.

chapters/fragops.txt
oddhack: It's used in a couple of dozen places, but not often.

Tobski (approved):

chapters/textures.txt
Tobski: We do only have unsigned normalised depth formats in the spec, but what's the benefit of restating it here? The sentence seems true regardless, and if in some future extension we added an SNORM or SCALED format this would just be an interaction we're bound to miss.

chapters/fragops.txt
```diff
@@ -955,22 +956,24 @@ fragments generated by <<primsrast-polygons-basic,back-facing>>
...
+endif::VK_VERSION_1_3,VK_EXT_extended_dynamic_state[]
+slink:VkStencilOpState::pname:compareOp during pipeline creation.
+
+The compare mask [eq]#s~c~# and stencil reference value [eq]#s~r~# of the
+relevant state set determine arguments of the comparison operation.
```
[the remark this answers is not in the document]
Tobski: This is C+P from the original text, it's meant to be referencing either the front or back facing state. I think this was clearer before because it was in the preceeding paragraph, whereas now they are separated.

Perhaps simply:

```suggestion
front or back stencil state set determine arguments of the comparison operation.
```

MIght also be worth reordering these paragraphs though.

chapters/textures.txt
krOoze: Purpose is linguistical precision and less brittleness. As you point out, if you added SCALED, this interaction is bound to be missed.

If it is signed, then it makes less sense for it to be clamped to 0. If it is not normalized, then it makes less sense to be clamped to unit. If you later added SCALED, this language would likely be defective. If you precisely specified the type, then it would "only" be unspecified for SNORM.

In a specification, being wrong is worse that being omitful. Ergo not being precise with the format type here is brittle.

krOoze: PS: Even less brittle or systematic would perhaps be to specify that the float32 is converted as per the conversion chapter (which already does include clamping).

Tobski: Fair point, I read the interaction differently but you're right, this really should only apply to those cases.

chapters/fragops.txt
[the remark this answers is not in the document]
oddhack: The flink: macro will rewrite vkCmdSetStencilOp -> vkCmdSetStencilOpEXT if the former is not supported at build time but the latter is - see LinkInlineMacroBase in config/spec-macros/extension.rb - so that's what comes out in the 1.0-extensions build. That contributed to greatly reducing the number of dead links in the older spec builds when it was introduced.

chapters/fragops.txt
```diff
@@ -955,22 +956,24 @@ fragments generated by <<primsrast-polygons-basic,back-facing>>
...
-during pipeline creation.
-
-The stencil reference and attachment values [eq]#s~r~# and [eq]#s~a~# are
-each independently combined with the compare mask [eq]#s~c~# using a logical
-code:AND operation to create masked reference and attachment values
-[eq]#s'~r~# and [eq]#s'~a~#.
-[eq]#s'~r~# and [eq]#s'~a~# are used as [eq]#A# and [eq]#B#, respectively,
-in the operation specified by elink:VkCompareOp.
+[eq]#s~r~# and [eq]#s~a~# are each independently combined with the compare
+mask [eq]#s~c~# using a logical code:AND operation to create masked
```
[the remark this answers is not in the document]
oddhack: s~a~ is defined back on line 929.

chapters/fragops.txt
```diff
@@ -955,22 +956,24 @@ fragments generated by <<primsrast-polygons-basic,back-facing>>
...
+reference and attachment values [eq]#s'~r~# and [eq]#s'~a~#.
+[eq]#s'~r~# and [eq]#s'~a~# are used as the _reference_ and _target_ values,
```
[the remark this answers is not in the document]
oddhack: I'd rather not have the term 'test' used for yet another purpose.

chapters/samplers.txt
oddhack: The style is to define a common enum in the first place it is used, though I could see a case for this. Will keep it in mind but not as part of this PR.

chapters/textures.txt
oddhack: D is used to refer to the depth component in a couple of places in this chapter, but I will tweak the sentence a bit.

chapters/fragops.txt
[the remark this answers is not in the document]
oddhack: Don't know where the 'logical' came from, good catch. Checked to make sure it's not still in the GL spec where this was derived from, which it's not. Odd.

oddhack:
@krOoze you can take another pass if you want, but I think I've addressed the points you raised, if not always in the way you suggested.

krOoze (commented):
@oddhack Second pass

chapters/fragops.txt
```diff
@@ -723,8 +723,9 @@ endif::VK_ARM_rasterization_order_attachment_access[]
     writes>> are enabled when pname:depthTestEnable is ename:VK_TRUE.
     Depth writes are always disabled when pname:depthTestEnable is
     ename:VK_FALSE.
-  * pname:depthCompareOp is the comparison operator used in the
-    <<fragops-depth,depth test>>.
+  * pname:depthCompareOp is a elink:VkCompareOp value specifying the
+    comparison operator to use in the <<fragops-depth-comparision, Depth
```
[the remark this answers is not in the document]
oddhack: Spaces before anchor text are not covered by the style guide, nor significant to the output. About 50/50 either way.

chapters/textures.txt
```diff
@@ -839,33 +839,17 @@ sparse unbound texel is undefined:.
...
-    \textit{false}                         & \text{for NEVER}
-  \end{cases} \\
-D & = 0.0 & \text{otherwise}
-\end{aligned}
-+++++++++++++++++++
+The compare operation is selected by the elink:VkCompareOp value set by
+slink:VkSamplerCreateInfo::pname:compareOp.
+The _reference_ and _target_ values in the operation specified by the
+elink:VkCompareOp are the texel depth value [eq]#D~tex~# and the reference
+value from the SPIR-V operand [eq]#D~ref~#, respectively.
```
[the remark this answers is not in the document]
oddhack: That was an error, thanks. The "respectively" is intentional and needed for disambiguation as in the 140-odd other places it's used; saying "A and B are treated as S and T" is unclear.

chapters/textures.txt
```diff
@@ -839,33 +839,17 @@ sparse unbound texel is undefined:.
...
-    D_{\textit{ref}} \neq D_{\textit{tex}} & \text{for NOTEQUAL} \\
-    \textit{true}                          & \text{for ALWAYS}   \\
-    \textit{false}                         & \text{for NEVER}
-  \end{cases} \\
-D & = 0.0 & \text{otherwise}
-\end{aligned}
-+++++++++++++++++++
+The compare operation is selected by the elink:VkCompareOp value set by
+slink:VkSamplerCreateInfo::pname:compareOp.
+The _reference_ and _target_ values in the operation specified by the
```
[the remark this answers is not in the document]
oddhack: Rewrote like the stencil description.

chapters/textures.txt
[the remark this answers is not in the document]
krOoze: It matters little if the "respectively" is implied or written explicitly. Do note two three reviewers (incl. myself on first pass) completely skipped passed this introduced spec bug.

If it was written "A is treated as S, and B is treated as T", then the sentence would scream clarity without necessity of mental gymnastics by the reader and more likely to be caught in review.

PR #1823, opened 2022-04-07
Fix format generation and add packed list

sjfricke:
- the first commit I realize the `def getConditions()` function was not working because the original plan for `<format>` was to include the "enable" conditional like it was in the SPIR-V elements in the XML... This works the code to pass the conditional information (ex. `ifdef::`) so it works as intended
- second commit replaces the current manual list of packed formats with being auto-generated from the XML. I checked, all 54 formats are there with the added bonus of properly wrapping the `VK_EXT_ycbcr_2plane_444_formats` not as `VK_KHR_sampler_ycbcr_conversion`


Tobski:
Hi @sjfricke can you provide the output of the generated file, assuming you've already generated it?

sjfricke:
[compatibility.txt](https://github.com/KhronosGroup/Vulkan-Docs/files/8502668/compatibility.txt)
[packed.txt](https://github.com/KhronosGroup/Vulkan-Docs/files/8502669/packed.txt)

These are the 2 files (`packed.txt` is new, `compatibility.txt` is fixed) when running `./makeAllExts formatsinc`

Tobski:
Thanks @sjfricke - generated files LGTM. Little unfortunate that there's so many ifdef spams instead of grouping it in the compatibility file, but it's autogenerated and transient so I don't think it matters as long as it's correct. This has my approval.

@oddhack can you look at merging this?

Tobski (approved):

PR #1827, opened 2022-04-19
Require Wayland WSI implementations to send wl_surface.commit in

swick:
vkQueuePresentKHR for all present modes. Clients already rely on this
behavior and the change merely syncs the spec to reality.

Context: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/15786#note_1337125

### DIFF
--- a/chapters/VK_KHR_wayland_surface/platformCreateSurface_wayland.txt
+++ b/chapters/VK_KHR_wayland_surface/platformCreateSurface_wayland.txt
@@ -68,10 +68,8 @@ objects created from a particular code:wl_display and code:wl_surface.
 Also, calling flink:vkQueuePresentKHR will result in Vulkan sending
 code:wl_surface.commit requests to the underlying code:wl_surface of each
 sname:VkSwapchainKHR objects referenced by pname:pPresentInfo.
-If the swapchain is created with a present mode of
-ename:VK_PRESENT_MODE_MAILBOX_KHR or ename:VK_PRESENT_MODE_IMMEDIATE_KHR,
-then the corresponding code:wl_surface.attach, code:wl_surface.damage, and
-code:wl_surface.commit request must: be issued by the implementation during
+The code:wl_surface.attach, code:wl_surface.damage, and
+code:wl_surface.commit requests must: be issued by the implementation during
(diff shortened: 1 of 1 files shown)


fooishbar:
@swick are you intending to sign the CLA?

I think this change is good and should be landed. There is a lot of dependent protocol state that gets latched on `wl_surface_commit()`. For instance, the following scenario would break (and we've observed this on bad EGL implementations which delay SwapBuffers commits to a thread):
1. client is happily rendering in windowed mode in FIFO
2. compositor requests the app go fullscreen (Vulkan is not notified of this)
3. app acks the compositor's configure request, indicating that it is complying with the request
4. app requests Vulkan resize the swapchain to fullscreen
5. delayed FIFO commit from windowed mode is executed, including `wl_surface_commit()`
6. compositor sends a fatal protocol error to the client

It would be possible to make this work, but would require an elaborate dance between app and Vulkan, including something like `vkWaitForPresentKHR` and various stalls before the app could commit the state.

It would be far preferable to have a Wayland protocol to allow FIFO mode, so the commits would be delayed on the server side rather than on the client. This is analagous to X11 and most other protocols. We did design a protocol to be able to do this, but it was written for the present-timing extension which has changed scope substantially since we rewrote it. It would probably be better to do an MVP which would make FIFO swapchains work (including non-0/1 `eglSwapInterval`), and handle full present-timing targeting on top of that.

swick:
> @swick are you intending to sign the CLA?

Yeah, tried to do that but something stalled internally. Gave it a kick again so hopefully this will happen soon. 

> It would probably be better to do an MVP which would make FIFO swapchains work (including non-0/1 eglSwapInterval), and handle full present-timing targeting on top of that.

Probably a good idea.

stonesthrow:
@chadversary - please put this up for review/feedback from WSI WG.

swick:
Alright, CLA is signed. Anything else you need from me?

chadversary:
@swick , the SI WG will review this in the next meeting. After that, since @fooishbar approves the change, I expect it to get promptly merged.

stonesthrow (approved):
Seems OK to me

PR #1833, opened 2022-04-25
Rewrite dependency chapter

### LINKED ISSUE #1815: Definition of synchronization scopes as sets
I've been trying to understand from the Vulkan spec the precise definition of a synchronization scope. It's not clear to me whether a synchronization scope is:
1. A set of pipeline stages. For example, a synchronization scope could be {TRANSFER, VERTEX_SHADER}.
2. Or, a set of operations. For example, { _Copy region R from buffer B to C_, _Run vertex shader V_ }. That is, a synchronization scope would be the set of actual operations across the relevant pipeline stages.

The consensus seems to be that the first definition is correct. For example, my reading of the answer to this StackOverflow question (https://stackoverflow.com/questions/63587925/what-happens-when-srcstagemask-specifies-a-stage-which-is-not-in-the-pipeline-of) seems to indicate definition 1.

Another example is the blog post at https://themaister.net/blog/2019/08/14/yet-another-blog-explaining-vulkan-synchronization/ in which the discussion of execution dependency chains also seems to imply definition 1.

However, assuming definition 1 is the intention, this seems to potentially present an issue with the definition of _execution dependency_ in S 7.1. Execution and Memory Dependencies of the Vulkan spec.

The spec states:

> Let A and B be separate sets of operations.
> 
> Let S be a synchronization command.
> 
> Let A<sub>S</sub> and B<sub>S</sub> be the synchronization scopes of S.
> 
> Let A' be the intersection of sets A and A<sub>S</sub>.
> 
> Let B' be the intersection of sets B and B<sub>S</sub>.
> 

If the synchronization scopes A<sub>S</sub> and B<sub>S</sub> are sets of _pipeline stages_ according to definition 1, then the use of the term _intersection_ seems like it doesn't match the standard mathematical meaning of set intersection. I.e. A<sub>S</sub> would contain pipeline stages and A would contain operations, so we are taking the intersection of sets containing different types of objects. Is the spec using the term _intersection_ in a more informal sense rather than the specific mathematical concept of set intersection?

I'd really appreciate any insight into whether definition 1 or 2 is correct, and if definition 1, then how should the term _intersection_ be interpreted?

krOoze:
fix #1815

### DIFF
--- a/chapters/synchronization.txt
+++ b/chapters/synchronization.txt
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
 command's two _synchronization scopes_.
 
 [[synchronization-dependencies-scopes]]
-The synchronization scopes define which other operations a synchronization
-command is able to create execution dependencies with.
-Any type of operation that is not in a synchronization command's
+A synchronization scope is a set of selection criteria for operations.
+A synchronization command has a first synchronization scope and a second
+synchronization scope for the purpose od declaring which kinds of operations
+are subject to the dependency introduced by the command.
+Any type of operation that is not selected by the synchronization command's
 synchronization scopes will not be included in the resulting dependency.
+
+[NOTE]
+.Note
+====
 For example, for many synchronization commands, the synchronization scopes
-can: be limited to just operations executing in specific
+can: be limited to just operations executing in only specific
 <<synchronization-pipeline-stages,pipeline stages>>, which allows other
 pipeline stages to be excluded from a dependency.
-Other scoping options are possible, depending on the particular command.
+But other scope to other kinds of operations is possible depending on the
+particular command, such as the <<synchronization-fences-signaling,
+fence signal operation>>.
+====
 
 [[synchronization-dependencies-execution]]
-An _execution dependency_ is a guarantee that for two sets of operations,
-the first set must: _happen-before_ the second set.
-If an operation happens-before another operation, then the first operation
-must: complete before the second operation is initiated.
+An _execution dependency_ is an implementation's guarantee that for two sets
+of operations, the first set must: _happen-before_ the second set.
+If an operation _happens-before_ another operation, it means the first
+operation must: complete before the second operation is initiated.
 More precisely:
 
-  * Let *A* and *B* be separate sets of operations.
+  * Let *Ops* be a set of all submitted operations.
   * Let *S* be a synchronization command.
-  * Let *A~S~* and *B~S~* be the synchronization scopes of *S*.
-  * Let *A'* be the intersection of sets *A* and *A~S~*.
-  * Let *B'* be the intersection of sets *B* and *B~S~*.
-  * Submitting *A*, *S* and *B* for execution, in that order, will result in
-    execution dependency *E* between *A'* and *B'*.
-  * Execution dependency *E* guarantees that *A'* happens-before *B'*.
+  * Let *Scope~src~* be the first synchronization scope of *S*.
+  * Let *Scope~dst~* be the second synchronization scope of *S*.
+  * Let *Ops~src~* be the subset of operations in *Ops* selected by scope
+    *Src~S~*.
+  * Let *Ops~dst~* be the subset of operations in *Ops* selected by scope
+    *Dst~S~*.
+  * Submitting *S* for execution will result in an execution dependency *ED*
+    between *Ops~src~* and *Ops~dst~*.
+  * Execution dependency *ED* guarantees that *Ops~src~* happens-before
+    *Ops~dst~*.
 
 [[synchronization-dependencies-chains]]
-An _execution dependency chain_ is a sequence of execution dependencies that
-form a happens-before relation between the first dependency's *A'* and the
-final dependency's *B'*.
-For each consecutive pair of execution dependencies, a chain exists if the
-intersection of *B~S~* in the first dependency and *A~S~* in the second
-dependency is not an empty set.
-The formation of a single execution dependency from an execution dependency
-chain can be described by substituting the following in the description of
-execution dependencies:
+A sequence of execution dependencies can: form a transitive property called
+an _execution dependency chain_.
+An _execution dependency chain_ is formed between any and all two execution
+dependencies whenever there is an overlap in the first dependency's
+*Scope~dst~* and subsequent dependency's *Scope~src~*.
+There is an overlap between two scopes if it is not possible to reduce the
+combination of the selection criteria of the *Scope~dst~* and the
+*Scope~src~* to nothing.
 
-  * Let *S* be a set of synchronization commands that generate an execution
-    dependency chain.
-  * Let *A~S~* be the first synchronization scope of the first command in
-    *S*.
-  * Let *B~S~* be the second synchronization scope of the last command in
-    *S*.
+[NOTE]
+.Note
+====
+For example, if a flink:vkCmdPipelineBarrier has pname:dstStageMask of
+ename:VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT, and subsequent
+flink:vkCmdPipelineBarrier has pname:srcStageMask of
+ename:VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT, the combined selection
+criterion of both is "commands occuring later than the first barrier and
+earlier than the second barrier reduced to the operations in
+ename:VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT,
+ename:VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT,
+ename:VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT,
+and ename:VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT stages".
+That is a valid non-empty selection criterium, therefore an
+_execution dependency chain_ is formed between these two barriers.
+
+Only the order of submission of the synchronization operations, and its
+scopes matter for determining whether an _execution dependency chain_ has
+been formed. There need not be any operation actually executed inbetween the
+two synchronization operations for a chain to form. Only the selection
+criteria matters for determining this, and whether actual real executed
+operations are being selected or not is irrelevant.
+====
+
+Synchronization operations that formed an _execution dependency chain_ in
+addition to introducing their individual
+<<synchronization-dependencies-execution,execution dependencies>> also
+introduce and additional third execution dependency.
+This additional dependency guarantees that the first dependecy's *Ops~src~*
+_happen-before_ the second dependency's *Ops~dst~*.
+This dependency can: itself also form a chain with another execution
+dependency, which can form an _execution dependency chain_ of more than two
+depenencies.
+_Execution dependency chain_ complements the specification of individual
+execution dependencies with the following description:
+
+  * Let *S~1~*, *S~2~*, ..., *S~N~* be a list of *N* synchronization
+    commands submitted in this order.
+  * Let's assume *S~1~*, *S~2~*, ..., *S~N~* are all part of an
+    _execution dependency chain_ *EC* such that *S~N~* is directly or
+    transitively chained with *S~1~*.
+  * Let *Scope~src~* be the first synchronization scope of *S~1~*.
+  * Let *Scope~dst~* be the second synchronization scope of *S~N~*.
+  * The chain *EC* introduces same
+    <<synchronization-dependencies-execution,execution dependency>>
+    guarantees as if instead an individual synchronization command *S* was
+    executed with these *Scope~src~* and *Scope~dst~* scopes.
 
 Execution dependencies alone are not sufficient to guarantee that values
 resulting from writes in one set of operations can: be read from another set
 of operations.
 
 [[synchronization-dependencies-available-and-visible]]
-Three additional types of operations are used to control memory access.
-_Availability operations_ cause the values generated by specified memory
-write accesses to become _available_ to a memory domain for future access.
-Any available value remains available until a subsequent write to the same
-memory location occurs (whether it is made available or not) or the memory
-is freed.
-_Memory domain operations_ cause writes that are available to a source
-memory domain to become available to a destination memory domain (an example
-of this is making writes available to the host domain available to the
-device domain).
-_Visibility operations_ cause values available to a memory domain to become
-_visible_ to specified memory accesses.
+Three additional types of operations are used to control memory access:
+
+  * _Availability operations_ cause the values generated by specified memory
+    write accesses to become _available_ in a specific memory domain for
+    future access.
+    Any available value remains available until a subsequent write to the
+    same memory location occurs (whether it is made available or not) or the
+    memory is freed.
+
+  * _Memory domain operations_ cause writes that are available in a source
+    memory domain to become available in a destination memory domain (an
+    example of this is making writes available in the host domain available
+    in the device domain).
+
+  * _Visibility operations_ cause values available in a specific memory
+    domain to become _visible_ to specified memory accesses.
+
+ifdef::editing-notes[]
+[NOTE]
+.editing-note
+====
+It feels the following is kinda important and should be part of core 1.0
+Vulkan to some extent.
+====
+endif::editing-notes[]
 
 ifdef::VK_VERSION_1_2,VK_KHR_vulkan_memory_model[]
 Availability, visibility, memory domains, and memory domain operations are
@@ -137,32 +203,34 @@ they can: be read or written by that type of memory access.
 Most synchronization commands in Vulkan define a memory dependency.
 
 [[synchronization-dependencies-access-scopes]]
-The specific memory accesses that are made available and visible are defined
-by the _access scopes_ of a memory dependency.
+The specific memory accesses that are made available and visible are
+selected by the criteria in _access scopes_ of a memory dependency.
 Any type of access that is in a memory dependency's first access scope and
-occurs in *A'* is made available.
+occurs in *Ops~src~* is made available.
 Any type of access that is in a memory dependency's second access scope and
-occurs in *B'* has any available writes made visible to it.
-Any type of operation that is not in a synchronization command's access
-scopes will not be included in the resulting dependency.
+occurs in *Ops~dst~* has any available writes made visible to it.
+Any type of operation and access that is not selected by synchronization
+command's access scopes will not be included in the resulting dependency.
 
 A memory dependency enforces availability and visibility of memory accesses
 and execution order between two sets of operations.
 Adding to the description of <<synchronization-dependencies-chains,
 execution dependency chains>>:
 
-  * Let *a* be the set of memory accesses performed by *A'*.
-  * Let *b* be the set of memory accesses performed by *B'*.
-  * Let *a~S~* be the first access scope of the first command in *S*.
-  * Let *b~S~* be the second access scope of the last command in *S*.
-  * Let *a'* be the intersection of sets *a* and *a~S~*.
-  * Let *b'* be the intersection of sets *b* and *b~S~*.
-  * Submitting *A*, *S* and *B* for execution, in that order, will result in
-    a memory dependency *m* between *A'* and *B'*.
-  * Memory dependency *m* guarantees that:
-  ** Memory writes in *a'* are made available.
-  ** Available memory writes, including those from *a'*, are made visible to
-     *b'*.
+  * Let *OpsMem~src~* be the set of memory accesses performed by *Ops~src~*.
+  * Let *OpsMem~dst~* be the set of memory accesses performed by *Ops~dst~*.
+  * Let *AccessScope~src~* be the first access scope of *S~1~*.
+  * Let *AccessScope~dst~* be the second access scope of *S~N~*.
+  * Let *Accesses~src~* be the subset of memory accesses in *OpsMem~src~*
+    selected by access scope *AccessScope~src~*.
+  * Let *Accesses~dst~* be the subset of memory accesses in *OpsMem~dst~*
+    selected by access scope *AccessScope~dst~*.
+  * Submitting *S* for execution will result in an execution dependency *ED*
+    between *Ops~src~* and *Ops~dst~*.
+  * Memory dependency *MD* guarantees that:
+  ** Memory writes in *Accesses~src~* are made available.
+  ** Available memory writes, including those from *Accesses~src~*, are
+     made visible to *Accesses~dst~*.
 
 [NOTE]
 .Note
@@ -833,13 +901,13 @@ pipeline stage must: not happen-before completion of a logically earlier
 stage.
 This means that including any stage in the source stage mask for a
 particular synchronization command also implies that any logically earlier
-stages are included in *A~S~* for that command.
+stages are included in *Scope~src~* for that command.
 
 Similarly, initiation of a logically earlier pipeline stage must: not
 happen-after initiation of a logically later pipeline stage.
 Including any given stage in the destination stage mask for a particular
 synchronization command also implies that any logically later stages are
-included in *B~S~* for that command.
+included in *Scope~dst~* for that command.
 
 [NOTE]
 .Note


chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
+.Note
+====
+For example, if a flink:vkCmdPipelineBarrier has pname:dstStageMask of
+ename:VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT, and subsequent
+flink:vkCmdPipelineBarrier has pname:srcStageMask of
+ename:VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT, the combined selection
+criterion of both is "commands occuring later than the first barrier and
+earlier than the second barrier reduced to the operations in
+ename:VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT,
+ename:VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT,
```
williamih, line 115: I think this is a really helpful example.

Just one question. I could be wrong about this, but I'm thinking VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT might not be included in the combined selection criteria, because this stage happens before the fragment shader stage - does that seem correct to you?

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
-  * Let *A~S~* and *B~S~* be the synchronization scopes of *S*.
-  * Let *A'* be the intersection of sets *A* and *A~S~*.
-  * Let *B'* be the intersection of sets *B* and *B~S~*.
-  * Submitting *A*, *S* and *B* for execution, in that order, will result in
-    execution dependency *E* between *A'* and *B'*.
-  * Execution dependency *E* guarantees that *A'* happens-before *B'*.
+  * Let *Scope~src~* be the first synchronization scope of *S*.
+  * Let *Scope~dst~* be the second synchronization scope of *S*.
+  * Let *Ops~src~* be the subset of operations in *Ops* selected by scope
+    *Src~S~*.
```
williamih, line 87: Should this be Scope~Src~ not Src~S~? And similarly for the destination scope below?

chapters/synchronization.txt
krOoze: Copypasta error. I think they are listed this way in the enum, but in the graphics pipe there's just late test inbetween

chapters/synchronization.txt
krOoze: Yea. 

Going back and forth on those var names. Suggestions?

williamih: Hmm - Scope~src makes sense to me. I think that may be better than Src S because in some places in the spec the variable **S** is used to mean a synchronization command (rather than a scope).

Other people may have more ideas though!

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
 [[synchronization-dependencies-execution]]
-An _execution dependency_ is a guarantee that for two sets of operations,
-the first set must: _happen-before_ the second set.
-If an operation happens-before another operation, then the first operation
-must: complete before the second operation is initiated.
+An _execution dependency_ is an implementation's guarantee that for two sets
+of operations, the first set must: _happen-before_ the second set.
+If an operation _happens-before_ another operation, it means the first
+operation must: complete before the second operation is initiated.
 More precisely:
```
Tobski, line 80: I'm not sure adding "an implementation's" here has the intended effect. My gut says maybe we need to work harder to separate the definition from who guarantees it.

Maybe:

```suggestion
An _execution dependency_ defines a relationship between two sets
of operations, such that one set happens-before the other.
Applications can: express execution dependencies via synchronization
commands that require: an implementation to uphold this guarantee,
as defined below: 
```

chapters/synchronization.txt
Tobski: \*Scope\~src~* is fine, though I'd be  tempted to call it \*Scope\~first~* (and second) to match the wording in the spec.

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
 command's two _synchronization scopes_.
 
 [[synchronization-dependencies-scopes]]
-The synchronization scopes define which other operations a synchronization
-command is able to create execution dependencies with.
-Any type of operation that is not in a synchronization command's
+A synchronization scope is a set of selection criteria for operations.
```
Tobski, line 56: Not super keen on this wording but I see where you're going. Maybe this:

```suggestion
A synchronization scope identifies a set of operations relative to a synchronization command.
```

?

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
+A synchronization command has a first synchronization scope and a second
+synchronization scope for the purpose od declaring which kinds of operations
+are subject to the dependency introduced by the command.
+Any type of operation that is not selected by the synchronization command's
 synchronization scopes will not be included in the resulting dependency.
```
Tobski, line 61:
```suggestion
synchronization scope identifying two sets of operations that will be synchronized
by the command.
Operations that aren't included in these scopes will not be affected by the
command.
```

I'd rather avoid saying "dependency" in a definition until after the concept has been introduced below (yes I know the text already says it, but we're improving things here 😬)

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
-
-  * Let *S* be a set of synchronization commands that generate an execution
-    dependency chain.
-  * Let *A~S~* be the first synchronization scope of the first command in
-    *S*.
-  * Let *B~S~* be the second synchronization scope of the last command in
-    *S*.
+A sequence of execution dependencies can: form a transitive property called
+an _execution dependency chain_.
+An _execution dependency chain_ is formed between any and all two execution
```
Tobski, line 98:
```suggestion
An _execution dependency chain_ is formed between any two execution
```

Unnecessary verbage, easier to read without IMO.

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
+dependencies whenever there is an overlap in the first dependency's
+*Scope~dst~* and subsequent dependency's *Scope~src~*.
+There is an overlap between two scopes if it is not possible to reduce the
+combination of the selection criteria of the *Scope~dst~* and the
+*Scope~src~* to nothing.
```
Tobski, line 103: This language is a bit awkward. I'm inclined to say this is where using set theory makes sense. This is a definite intersection of two sets of the same type, and I'd suggest rewording it as such.

chapters/synchronization.txt
Tobski: I'm not keen on this specific example because it leans on the pipeline stage order, which isn't defined yet. I'd rather we highlight a simpler example here to avoid new readers getting confused - e.g. something that includes exactly the same stages directly. We could then elaborate on this example later when pipeline stage order is introduced to get that concept highlighted better (which might be a good improvement regardless).

For us I think this example makes sense, because we've read the spec and are familiar - but for a new reader this would be confusing.

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
+addition to introducing their individual
+<<synchronization-dependencies-execution,execution dependencies>> also
+introduce and additional third execution dependency.
+This additional dependency guarantees that the first dependecy's *Ops~src~*
+_happen-before_ the second dependency's *Ops~dst~*.
+This dependency can: itself also form a chain with another execution
+dependency, which can form an _execution dependency chain_ of more than two
+depenencies.
+_Execution dependency chain_ complements the specification of individual
+execution dependencies with the following description:
```
Tobski, line 139: This is a bit looser than I'd like, maybe:

```suggestion
An _execution dependency chain_ implicitly creates another
<<synchronization-dependencies-execution,execution dependency>>.
An execution dependency formed in this way guarantees that *Ops~src~*
in the first dependency in the chain _happens-before_ *Ops~dst~* in the
second dependency in the chain.
This new dependency can: itself become part of another chain, forming an
_execution dependency chain_ of multiple dependencies, up to and including
all operations in the application.
```

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
+
+  * Let *S~1~*, *S~2~*, ..., *S~N~* be a list of *N* synchronization
+    commands submitted in this order.
```
Tobski, line 142: Here you're leaning on submission order, which hasn't been introduced yet. In the execution dependency definition you did an excellent job of getting rid of this, and I think we could repeat that here.

The order of operations is unnecessary if you can formally define the way an execution dependency chain is formed by the intersection of scopes between operations.

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
-_visible_ to specified memory accesses.
+Three additional types of operations are used to control memory access:
+
+  * _Availability operations_ cause the values generated by specified memory
+    write accesses to become _available_ in a specific memory domain for
+    future access.
+    Any available value remains available until a subsequent write to the
+    same memory location occurs (whether it is made available or not) or the
+    memory is freed.
+
```
Tobski, line 166: IIRC these spaces lead to unnecessary gaps in the bullet list, as asciidoctor treats them as separate lists, so would like them removed.

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
+    domain to become _visible_ to specified memory accesses.
+
+ifdef::editing-notes[]
+[NOTE]
+.editing-note
+====
+It feels the following is kinda important and should be part of core 1.0
+Vulkan to some extent.
+====
+endif::editing-notes[]
```
Tobski, line 182: At some point we hope to better integrate the memory model appendix with the synchronization chapter in general, it's just not something we've had the resource to do, _yet_ - I'm not sure an editors note here is necessary as this is tracked internally, but I'm not particularly against including it. For now we've just hidden it behind the extension as some of the definitions only _somewhat_ apply without it.

Mostly just adding some context here, not necessarily suggesting a change.

Tobski (changes requested):
I think I'm happy with the direction this is going in, just needs some tightening on various bits of the new language.

Tobski:
I've requested two extra reviewers from the WG who had heavy input on the original language here, in case they have any input.

chapters/synchronization.txt
stonesthrow: "happens-before" => "is executed" ? "happens" seems ambiguous.

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
+[NOTE]
+.Note
+====
 For example, for many synchronization commands, the synchronization scopes
-can: be limited to just operations executing in specific
+can: be limited to just operations executing in only specific
 <<synchronization-pipeline-stages,pipeline stages>>, which allows other
 pipeline stages to be excluded from a dependency.
-Other scoping options are possible, depending on the particular command.
+But other scope to other kinds of operations is possible depending on the
```
stonesthrow, line 70: Line 70, reword.

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
 command's two _synchronization scopes_.
 
 [[synchronization-dependencies-scopes]]
-The synchronization scopes define which other operations a synchronization
-command is able to create execution dependencies with.
-Any type of operation that is not in a synchronization command's
+A synchronization scope is a set of selection criteria for operations.
+A synchronization command has a first synchronization scope and a second
+synchronization scope for the purpose od declaring which kinds of operations
```
stonesthrow, line 58: "od" -> "of"

chapters/synchronization.txt
stonesthrow: I do think examples might make clear some of the abstract language of scopes and dependencies. So I encourage that.

stonesthrow (commented):
Just some small comments

chapters/synchronization.txt
williamih: To my mind, these two options are not just different wordings, but actually express two different concepts.

I.e. is a synchronization scope the criteria or is it the operations?

In my opinion, either could work (with some refinements to the wording) but they have some different pros/cons.

chapters/synchronization.txt
krOoze: @Tobski I try to make as brutal example here as possible. I suppose it is sort of a desperate attempt to compensate the lack of sufficiently formal description of what "scope overlap" is...

In my mind the concept is pretty straightforward, so it is mindly annoying that it might probably require like half a page to properly explain in the end. And even so will probably require an example to have some certainty people grok it correct.

chapters/synchronization.txt
krOoze: @Tobski The nomenclature is sorta mixed. The API parameters use src/dst. TBH I like it more because first/second does not convey a difference between the scope in how they are used.

chapters/synchronization.txt
```diff
@@ -53,67 +53,133 @@ _memory dependencies_ between two sets of operations defined by the
...
-If an operation happens-before another operation, then the first operation
-must: complete before the second operation is initiated.
+An _execution dependency_ is an implementation's guarantee that for two sets
+of operations, the first set must: _happen-before_ the second set.
+If an operation _happens-before_ another operation, it means the first
+operation must: complete before the second operation is initiated.
 More precisely:
 
-  * Let *A* and *B* be separate sets of operations.
+  * Let *Ops* be a set of all submitted operations.
```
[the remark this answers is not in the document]
krOoze: @Tobski I try to make the scopes do all the work, so if they do not want to include *S* they just need to say so.

I have not had time to wholistically review stuff yet (hence draft PR), but typically the scope is writen like "all operations recorded before S that operate in stage P", therefore S should not never be part of the scope.

chapters/synchronization.txt
krOoze: I think this is basically first definition of what "happens-before" means, which has very formal meaning. But I think it is otherwisely unfortunately specified only in the extension.

chapters/synchronization.txt
krOoze: @Tobski Yea, I try superhard to avoid implying here that scopes are operation. If scopes did "identify set of operations" that basically implies it is an identity with set of operations. I don't want it to be set of operations at all. I want it sort of be a SQL query instead. Which can later on access a set of operations and filter them down.

chapters/synchronization.txt
krOoze: Dependencies are introduced in the first paragraph of the chapter. For that matter it is the name of the chapter.

chapters/synchronization.txt
krOoze: It is sort of a legal doublet. "any and all" means "every".

chapters/synchronization.txt
krOoze: It is best I came up with yet.

Problem is if you want it to be set intersection of same types, then what is the type. Set theory gives the wrong idea here and the wrong type of arithmetic. What I want here is "merger of two SQL queries" and not set intersection. If I wanted set intersection, I would need a static type that forms a static sets. But scopes are sort of dynamic types. They are a predicate. You cannot make a set intersection of predicates. As far as I know. If we really want to go uberformal, we need to introduce bunch of predicate logic theory.

chapters/synchronization.txt
Tobski: I mean, I get it, but we consistently use "any" for this meaning by itself, precisely for the reasons I mentioned above. It'd be inconsistent to change that now.

chapters/synchronization.txt
Tobski: I was trying to state these as the "identifier" rather than the "set of operations" but I see the confusion. I haven't got a better idea so let's just leave it for now.

chapters/synchronization.txt
Tobski: Yea, reading it back it seems this is what we were trying to use as a sort of bootstrapped definition of happens-before, and I don't think it does a good job in light of the memory model definition. "Happens before" is an industry term though (it's on wikipedia!) so I'm kind of inclined to leave it in pending a memory model rewrite that properly introduces the idea later...

chapters/synchronization.txt
Tobski: @krOoze I don't disagree with your intent, but I do think a fresh reader would not be able parse this particular example until they've gone and read a lot more of the chapter. IMO we need something in the style of the Vulkan Guide or a proposal document - where you get a much higher level view of things without all the formality. There you could say a lot of this much more straightforwardly.

In the meantime, I'd rather we simplify this particular example.

chapters/synchronization.txt
Tobski: We landed on first/second for some reason that I honestly can't entirely remember. I'm fine with src/dst here, but might be worth considering a nomenclature change for the spec elsewhere to match - though I'd do that in a separate MR because that's going to be churny.

chapters/synchronization.txt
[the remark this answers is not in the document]
Tobski: Yea that is all in line with my thinking, so that's fine, just something to be aware of.

chapters/synchronization.txt
krOoze: Yea it is just the chain of synch dependencies.

chapters/synchronization.txt
Tobski: You've identified the two scopes as "sets of selection criteria" so you've identified the type already; calling out the intersection of those two sets as non-null suggests the selection criteria is empty. The only ambiguity here is whether this is an inclusive or exclusive set of criteria, which I don't think would be hard to clarify, and would dictate whether to use union or intersection here. The "predicate" idea you've suggested isn't something you've written down in this PR yet, and I'm not sure it would add anything.

We certainly can't introduce SQL queries as a concept in the Vulkan spec though 😛.

chapters/synchronization.txt
krOoze: Don't worry too much about formatting for now. I will self-review stuff if and when we dedraft this and I will render it and run spellcheck.

chapters/synchronization.txt
krOoze: Don't worry too much about typos for now. I will self-review stuff if and when we dedraft this and I will render it and run spellcheck.

chapters/synchronization.txt
krOoze: Well it does not unfortunately work that way. Consider this:

criterion1: "All operations after S1"  
criterion2: "All operations before S2"

The strings are clearly different and the criteria are clearly different, therefore set intersection of criterion1 and criterion2 is an empty set.

But the predicate "All operations after S1" ∧  "All operations before S2" yields an nonempty predicate "All operations after S1 and before S2".

Tobski: I'm confused, that sounds like _exactly_ what we want? Where's the problem here?

Tobski: Wait hang on, lemme rephrase...

Tobski: > The strings are clearly different and the criteria are clearly different, therefore set intersection of set {criterion1} and set {criterion2} is an empty set {}, because criterion1 ≠ criterion2.
> 
> But the predicate criterion1 ∧ criterion2 yields a new nonempty (resp. non-tautological) predicate "All operations after S1 and before S2".

Predicate logic is just a reframing of set logic in many sense - and is an intersection, or is a union. Your argument seems to be based on "the strings are different" against set logic but then you're basically using, set logic by reframing what is in the set. By taking the "and" of two predicates you're doing the mathematical equivalent of taking the union of two sets of conditions.

I don't much care whether we frame it as set or predicate logic, what I _really_ don't want though is to have predicate _and_ set logic in the same chapter, because that's too many concepts to overlap; that was the motivation for using sets in the first place.

krOoze: The problem here is set logic (intersection) yields empty result, wheere predicate logic (logical-and) would yield non-empty result. We want the predicate logic, and we do **not** want the set logic. Therefore we cannot use set intersection to describe this, unless we invoke "ethereal operations", which is its own can of worms.

Tobski: Ok, I see what you're trying to get at here, I don't think the "hypothetical operations" is a much a can of worms as you think it is, and I'm half tempted to go draft that MR so we can compare.

In lieu of doing that though, it sounds like you have an idea to frame this as a sort of database query, which is a perfectly reasonable mental model, but I suspect that if you want to make the spec work that way, you've got a LOT more work to do. I'm concerned that by combining concepts like this PR is currently doing we're heading into a rabbit hole that will just confuse things more.

Like I think you have a vision here for how it's going to work, and it's not the one the spec has currently, and that's fine, but I don't know yet if there's any merit to that framing over what we've currently got.

Based on the discussion in #1815, it felt to me like there was a much smaller fix to that issue, and I'm concerned this PR is starting to drift too far from _just_ solving that original issue. The set of changes to make it deal in "hypothetical operations" I suspect is much smaller.

chapters/synchronization.txt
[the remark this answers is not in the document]
stonesthrow: Possible alternative terms: A is the "prerequisite" operation(s) and B is the "dependent" operation(s).
Synchronization requires the prerequisite to execute before the dependent executes.

chapters/synchronization.txt
krOoze: @Tobski I was kinda strategically aiming for something hard. This is the where reader needs to slow down. The example demonstrates the original Issue. I suppose I can sacrifice the implicit pipeline stage part of the example for now?

chapters/synchronization.txt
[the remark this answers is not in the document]
krOoze: @stonesthrow I don't need A and B here at all, because scopes are generally already defined as downselecting to previous or subsequent operations.

chapters/synchronization.txt
krOoze: @stonesthrow It's a term defined in the memory_model extension, but it is used througout the specification.

chapters/synchronization.txt
krOoze: My problem with "hypothetical operation" is that forms a set of something unreal, magic, and uncountable. Can't say for sure without seeing this alternative, but I feel it will to similar vagueness as the original Issue points out. At least when sufficiently literal reader reads that.

After thinking about it. introducing "hypothetical operations" is sort of a hack\hotfix. And predicate logic is the appropriate indirection method to solve describing this. As it is now in the PR, it probably is not good either. But it feels one approach cannot be sufficiently formalized no matter how much one tries, while the other fundamentally can.

I would avoid the tempting dopamine hit of promptly closing an Issue. I admit I am not a believer in atomic fixes. It is a rare day when something is amenable to fixing atomically in a way it never comes up in subsequent issues. Something top-down and interwoven as the sync system needs some care. At the minimum I consider the whole section to be undivisible and in need to be always edited as a whole.

chapters/synchronization.txt
krOoze: The happens-before\after is frequently used, so should be integrated with the core spec sooner than later...

chapters/synchronization.txt
Tobski: Right, the pipeline stage order is really my main concern here - the rest is fine.

chapters/synchronization.txt
Tobski: I don't think I disagree with much of that, except for your assertion that the other approach can't be formalized. Editing this chapter is hard. I agree that "hypothetical operation" may look like a hotfix and won't know until I try writing it out though.

There's an internal task to rewrite all of this incorporating the memory model language much better, because as you've pointed out there's no formal definition of happens-before without it, and literally nobody is looking forward to it :)
Because of that I'm a tad wary of a major rewrite just now; was hoping that you'd be able to get more with less here. 

Ultimately, I tend to stick by the philosophy that "less is more" - if editing a chapter results in less words then it's a clear win. If it's about the same number of words then it's a much tougher sell. If there's more words, you've probably made it harder. This is a bit of a rough metric but I do find it helps when reviewing. So the thing I'd suggest is if you are planning on doing a lot of work here, plan it out first and try to find ways to say more with less. If you can reduce the number of concepts needed, I'd be much more inclined to approve this.

krOoze: Well, less words I can give you easily. Just delete all the words :p. Unfortunatelly there is another overriding priority: the words must describe completely, correctly, and coherently (4C) the things they describe.

It is unlikely I can substract concepts, because the Issue implies there actually is a concept missing (or not fully formed). What I try to do is first try to find the ideal state, and only then worry how I get there. Planning a route to a destination you don't know is a tricky business. Anyway I am happy with the name changes, which could now be a neat and likely noncontroversial patch that helps reasoning about all this stuff without getting lost.

oddhack:
Closing. There have been many changes to the dependency chapter in the meantime.

PR #1837, opened 2022-05-02
Make names in sync formalism more memorable

krOoze:


ikapoura:
I think that MemOps is more fitting than OpsMem because it fits in how we phrase it in our heads as "memory operations". You also introduce MemDep for "memory dependency" and we would have an inconsistency there. So, we could end with MemOps and ScopedMemOps.

krOoze:
I am not sure memory accesses are "operations" in Vulkan. That would be potentially misleading, because memory is not subject to execution dependency as described in the first formalism bullet list.

chapters/synchronization.txt
```diff
@@ -70,32 +70,37 @@ If an operation happens-before another operation, then the first operation
...
 chain can be described by substituting the following in the description of
 execution dependencies:
 
-  * Let *S* be a set of synchronization commands that generate an execution
-    dependency chain.
-  * Let *A~S~* be the first synchronization scope of the first command in
-    *S*.
-  * Let *B~S~* be the second synchronization scope of the last command in
-    *S*.
+  * Let *SyncChain* be a set of synchronization commands that generate an
```
oddhack, line 98: I don't think this is quite right - it is saying to substitute these terms in the language above, but there \*Sync* is referred to, not \*SyncChain*. It almost works except for the "Submitting \*Ops~1~*, \*Sync* and \*Ops~2~* for execution," clause. Using the same label for \*Sync* and \*SyncChain* would address this.

oddhack (changes requested):
Seems fine once the \*Sync* / \*SyncChain* issue I noted is addressed.

chapters/synchronization.txt
krOoze: Ok, sticking with the original "Sync" name for now.

chapters/synchronization.txt
```diff
@@ -151,18 +156,25 @@ and execution order between two sets of operations.
...
+    *ScopedOps~2~*.
+  * Let *AccessScope~1st~* be the first access scope of the first command in
+    *SyncChain*.
+  * Let *AccessScope~2nd~* be the second access scope of the last command in
+    *SyncChain*.
+  * Let *ScopedOpsMem~1~* be the intersection of sets *OpsMem~1~* and
+    *AccessScope~1st~*.
+  * Let *ScopedOpsMem~2~* be the intersection of sets *OpsMem~2~* and
+    *AccessScope~2nd~*.
+  * Submitting *Ops~1~*, *SyncChain*, and *Ops~2~* for execution, in that
```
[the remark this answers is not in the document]
oddhack: This appears done but not marked auto-resolved by GH for some reason. LGTM now.

oddhack (approved):

oddhack (approved):

PR #1843, opened 2022-05-08
Tighten synchronization scope specification

### LINKED ISSUE #1815: Definition of synchronization scopes as sets

krOoze:
limited version of #1833
fix #1815

Tighten the definition of what "synchronization scope" actually is.
That should resolve confusions about using set intersections in the specification, and confusions about execution dependency chaining,  supplemented by an example.

### DIFF
--- a/chapters/synchronization.txt
+++ b/chapters/synchronization.txt
@@ -48,17 +48,20 @@ synchronization mechanisms are exposed by Vulkan:
 
 An _operation_ is an arbitrary amount of work to be executed on the host, a
 device, or an external entity such as a presentation engine.
-Synchronization commands introduce explicit _execution dependencies_, and
-_memory dependencies_ between two sets of operations defined by the
-command's two _synchronization scopes_.
+Synchronization commands introduce explicit _execution dependency_, and
+_memory dependency_ between two sets of operations as selected by the _first
+synchronization scope_ and the _second synchronization scope_ of the
+command.
 
 [[synchronization-dependencies-scopes]]
-The synchronization scopes define which other operations a synchronization
-command is able to create execution dependencies with.
-Any type of operation that is not in a synchronization command's
-synchronization scopes will not be included in the resulting dependency.
+A _synchronization scope_ is a predicate logic formula that quantifies which
+operations a synchronization command is able to create execution dependency
+with.
+Any type of operation that does not satisfy the synchronization command's
+synchronization scope formula will not be included in the resulting
+dependency.
 For example, for many synchronization commands, the synchronization scopes
-can: be limited to just operations executing in specific
+can: limit synchronization to just those operations executing in a specific
 <<synchronization-pipeline-stages,pipeline stages>>, which allows other
 pipeline stages to be excluded from a dependency.
 Other scoping options are possible, depending on the particular command.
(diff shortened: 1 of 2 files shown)


williamih:
Thanks @krOoze - I like the predicate logic approach. Will have a look at this in more detail when I get a chance (probably in the next few days)

chapters/synchronization.txt
```diff
@@ -70,27 +73,71 @@ If an operation happens-before another operation, then the first operation
...
-  * Let *A'* be the intersection of sets *A* and *A~S~*.
-  * Let *B'* be the intersection of sets *B* and *B~S~*.
-  * Submitting *A*, *S* and *B* for execution, in that order, will result in
-    execution dependency *E* between *A'* and *B'*.
+  * Let *A~S~* be the first synchronization scope of *S*.
+  * Let *B~S~* be the second synchronization scope of *S*.
+  * Let *A'* be the set of operations satisfying *A~S~*.
+  * Let *B'* be the set of operations satisfying *B~S~*.
+  * Submitting *S* for execution will result in an execution dependency
+    *E* between *A'* and *B'*.
```
williamih, line 82: I think the use of predicate logic here is great. This seems intuitive and also more precise than the current spec.

I'm wondering what you think about this potential adjustment, which uses some more of the current wording:
```
* Let *A* and *B* be separate sets of operations.
* Let *S* be a synchronization command.
* Let *A~S~* be the first synchronization scope of *S*.
* Let *B~S~* be the second synchronization scope of *S*.
* Let *A'* be the set of operations in *A* satisfying *A~S~*.
* Let *B'* be the set of operations in *B* satisfying *B~S~*.
* Submitting *A*, *S* and *B* for execution, in that order, will result in
  execution dependency *E* between *A'* and *B'*.
```

The reasoning being that, at least to me, this makes it a bit more clear that `*B'*` can include operations that are submitted later. The note you have below clarifies this, which is great, but I think it would be nice if we could have this be really clear just from the main text of the spec, so that the note is less essential.

Arguably the update I'm suggesting is a bit redundant, because the predicates already "select" the operations, but I do think it reduces the need for the extra note. (My understanding is notes are meant to be more of an optional extra, and not essential for understanding the spec).

krOoze:
Dammit, this got forgotten. Gotta rebase this...

chapters/synchronization.txt
krOoze: Sorry for late answer. Yea it is somewhat annoying **B** is a temporal set. One has to go outside spacetime to see the full extent of **B**.

Notes\comments are always desperate measures. Make no mistake, I would do everyting humanely possible to try to eliminate them.

What you suggest is there originally, and what I wanted to avoid, because it makes several inaccurate preassumptions. For example, let's say **S** is a `vkWaitForFences`. What does it even mean submitting `vkWaitForFences`? There's not even any queue to submit to...

oddhack:
Closing.

PR #1858, opened 2022-05-18
Remove reference to nonexistent parameter.

David-DiGioia:
Removed a mention of `raygenShaderBindingOffset` that incorrectly referred to it as a parameter of `vkCmdTraceRaysKHR`. It is a parameter of `vkCmdTraceRaysNV`.


dgkoch:
That block of text is common for both KHR and NV extensions, so I'd rather change it to refer to both commands and the relevant parameters. I'll propose an alternative.

PR #1860, opened 2022-05-23
Add VK_KHR_maintenance4 relaxed interface VUID

sjfricke:
This came up in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4148

The description of ` maintenance4` feature says

> The interface matching rules allow a larger output vector to match with a smaller input vector, with additional values being discarded.

and there is language in the [interface matching section](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#interfaces-iointerfaces-matching) but there is no explicit VUID to assign to this.

Open as always to suggestions on wording or placement of the VU

### DIFF
--- a/appendices/spirvenv.txt
+++ b/appendices/spirvenv.txt
@@ -1223,9 +1223,9 @@ ifndef::VK_VERSION_1_3,VK_KHR_maintenance4[]
 endif::VK_VERSION_1_3,VK_KHR_maintenance4[]
 ifdef::VK_VERSION_1_3,VK_KHR_maintenance4[]
   * [[VUID-{refpage}-LocalSizeId-06434]]
-    if execution mode code:LocalSizeId is used, <<features-maintenance4,
+    If execution mode code:LocalSizeId is used, <<features-maintenance4,
     pname:maintenance4>> must: be enabled
-  * if <<features-maintenance4, pname:maintenance4>> is not enabled, any
+  * If <<features-maintenance4, pname:maintenance4>> is not enabled, any
     code:OpTypeVector output interface variables must: not have a higher
     code:Component code:Count than a matching code:OpTypeVector input
     interface variable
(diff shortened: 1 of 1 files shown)


appendices/spirvenv.txt
```diff
@@ -1217,11 +1217,18 @@ endif::VK_NV_ray_tracing_motion_blur[]
...
     The execution mode code:LocalSizeId must: not be used
+  * Any code:OpTypeVector output interface variables must: not have a higher
+    code:Component code:Count than a matching code:OpTypeVector input
+    interface variable
 endif::VK_VERSION_1_3,VK_KHR_maintenance4[]
 ifdef::VK_VERSION_1_3,VK_KHR_maintenance4[]
   * [[VUID-{refpage}-LocalSizeId-06434]]
     if execution mode code:LocalSizeId is used, <<features-maintenance4,
     pname:maintenance4>> must: be enabled
+  * if <<features-maintenance4, pname:maintenance4>> is not enabled, any
```
r-potter, line 1228:
```suggestion
  * If <<features-maintenance4, pname:maintenance4>> is not enabled, any
```

gnl21:
This looks good to me, but do we have VUs for where the output component count is less than the corresponding input count? I had a quick look and couldn't find anything.

sjfricke:
>  but do we have VUs for where the output component count is less than the corresponding input count?

No, there are quite a few shader interface validation that the validation layers are checking for (including this) but don't have dedicated VUID yet... been on my todo list to add them, still plan too, but might be better off as another PR

gnl21:
> there are quite a few shader interface validation that [...] don't have dedicated VUID yet [...] but might be better off as another PR

In general, yes, and I don't think we should hold this PR up trying to do too much, but adding VUs for x > y naturally makes me look for what happen when x < y, so I do think this one case is a bit more confusing than the others. It would be really great if it could be added to this PR.

Having said that, of course, everything here is correct, so I don't mean to block it. I can always add the missing thing myself.

oddhack:
@gnl21 going ahead and merging this - you might want to open a tracker on the additions you suggested above.

PR #1864, opened 2022-05-26
minor fixes for 1.3.215 spec

sjfricke:
No need for the `ifdef` wrapper for standalone SPIR-V VUs since its in the unified SPIR-V spec, there isn't one for `VUID-StandaloneSpirv-PerVertexKHR-06777` which also references `PerVertexKHR`


oddhack:
LGTM. This is editorial so approving.

PR #1870, opened 2022-06-08
github/ci: Split `steps` into individual `jobs` like GitLab CI

MarijnS95:
As requested by @oddhack 

---

~Intermediates are also published and communicated through artifacts. Unfortunately this is way slower than the original ±11m (not to mention much slower than GitLab CI, which sits at about ±7m) because the artifact is a whopping 190MB (compared to ±40MB on GitLab CI) despite including less files. This artifact takes about 4 minutes to download per job :(~

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,6 +40,15 @@ jobs:
           scripts/check_spec_links.py --html=gen/out/checks/problems.html > /dev/null || true
           make allchecks
 
+  spec-core:
+    name: Build the core-only spec, to try and catch ifdef errors in extension markup
+    runs-on: ubuntu-latest
+    container: khronosgroup/docker-images:asciidoctor-spec
+
+    steps:
+      - uses: actions/checkout@v3
+      - run: ./makeSpec -clean -spec core -genpath gencore QUIET= -j${nproc} -Otarget chunked html
+
   spec-generate:
     name: Build the vulkan specification and generate any associated files (such as vulkan.h)
     runs-on: ubuntu-latest
@@ -49,8 +58,6 @@ jobs:
       - uses: actions/checkout@v3
       - name: Build the actual spec (both chunked and single-page HTML), and other common targets
         run: ./makeSpec -clean -spec all QUIET= -j${nproc} -Otarget manhtmlpages validusage styleguide registry chunked html
-      - name: Build the core-only spec, to try and catch ifdef errors in extension markup
-        run: ./makeSpec -clean -spec core -genpath gencore QUIET= -j${nproc} -Otarget chunked html
       - name: Build headers, for use by all later stages
         run: make validate install test
         working-directory: xml
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -2,9 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Github CI file for vulkan spec and header generation
-# Several steps should be parallelizable, but it is unclear how to only
-# checkout the repository once for multiple jobs, and multiple steps are not
-# parallelizable at present.
 # See .gitlab-ci.yml for non-Actions comments and step dependencies.
 
 name: CI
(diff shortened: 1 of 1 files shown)


oddhack:
Disappointing, but thanks for pursuing this. On balance ISTM we should stay with the current, more monolithic CI script. But how sure are you that the artifact size relative to gitlab is the issue? When I look at the spec-generate log it says

> The raw size of all the files that were specified for upload is 201031673 bytes  
> The size of all the files that were uploaded is 40376502 bytes. This takes into account any gzip compression used to reduce the upload size, time and storage

and that's consistent with the size of the zipfile I downloaded - I think the "192 MB" shown on the summary page is the uncompressed size being reported. But I do see that the runtime for "Archive generated spec" is 7m44s which seems unreasonable, whatever the actual file size is.

oddhack:
It appears that the performance is so slow because it's performing an HTTPS transaction for every file in the artifact, and there are thousands of them. Alternatively we could make our own zipfile of, say, everything except the headers (since they're used by a downstream CI action) and just upload headers + zipfile: https://stackoverflow.com/questions/68470162/how-to-archive-files-in-artifact-for-github-workflow-actions-in-order-to-fix-thi and also https://github.com/actions/upload-artifact#limitations

MarijnS95:
Good hunting! I hadn't looked further than the `upload-artifact` and `download-artifact` job taking very long, and GitHub stating that it's 192MB:

image

That's probably wrongly reported, and the ~40MB you mentioned is roughly identical to GitLab's. It takes ages to download (presumably because their backend is taping all the files back together) bu the result is indeed 40.2MiB.

I didn't even know that GitHub would upload every file individually, presumably to ever build a browse UI for it? I'll definitely try out the suggestions, probably opting for `tar` as a `zip`-in-`zip` doesn't seem very helpful to me and recursive compression is needless.

MarijnS95:
0s for `tar`ing and `16s` for uploading, that's perfect.

Unfortunately GH CI is a bit slower than GitLab but we're still under the 11+ minutes baseline, clocking in at 10m 17s.

A big time-eater is initializing the docker containers, logs seem to indicate that none are cached and the layers are pulled fresh every time, for every individual job - all taking about 30-40s. The C/hpp steps shouldn't need it as GitHub's virtual environment is full of tools already, let's see if we can drop it.

oddhack:
There appears to be some possibility of explicitly telling Actions to cache container images (what, are we in the 1980s still?!) though I'm still researching. It might have something to do with the syntax of the uses: tag.

oddhack:
There's also an explicit cache for communicating artifacts between steps, I think: https://github.com/actions/cache

This all feels much more painful that it should be (perhaps I am triggered this morning after spending 45 minutes trying and failing to accomplish something on Citibank's also painful online banking system :-( ).

.github/workflows/CI.yml
```diff
@@ -17,45 +14,78 @@ on:
...
+  spec-generate:
+    name: Build the vulkan specification and generate any associated files (such as vulkan.h)
+    runs-on: ubuntu-latest
+    container: khronosgroup/docker-images:asciidoctor-spec
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build the actual spec (both chunked and single-page HTML), and other common targets
+        run: ./makeSpec -clean -spec all QUIET= -j${nproc} -Otarget manhtmlpages validusage styleguide registry chunked html
+      - name: Build the core-only spec, to try and catch ifdef errors in extension markup
```
oddhack, line 52: This is looking good - thanks very much for taking a run at this. BTW the two makeSpec builds are entirely independent, and could be done separately if you want to add that. The first run takes much longer than the second since it's building 1500-odd HTML reference pages as well as the spec, but even so the second adds 17s+ to the overall runtime. I am a little surprised at how slow the first build is, though - maybe we don't get many cores in the build environment. I usually build with -j12 or -j24 on my own 6-core machine.

I noticed you split the multiple-line run: | actions into separate name:/run: for each command line. Is that just for clarity and to pull the comments up into metadata in the CI script, or is there another benefit?

MarijnS95: > This is looking good

Thanks!

> BTW the two makeSpec builds are entirely independent, and could be done separately if you want to add that.

Let's separate them and see the results :)

> but even so the second adds 17s+ to the overall runtime.

It'll probably cost more cycles total when running in parallel, but should save on combined run length.

> I am a little surprised at how slow the first build is, though - maybe we don't get many cores in the build environment. I usually build with -j12 or -j24 on my own 6-core machine.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources It's only running on a 2-core VM, no ThreadRippers :/

> I noticed you split the multiple-line run: | actions into separate name:/run: for each command line. Is that just for clarity and to pull the comments up into metadata in the CI script, or is there another benefit?

Yeah, to pull comments into the metadata and show them in the job run overview. Otherwise, when a single step fails you're might be left wondering which of the lines in the script are at fault here (though, it's usually easy to guess). There was a thing in the past where these scripts were running without `set -x` and only the exit code of the last line was considered for job fail/success status.

MarijnS95: > > but even so the second adds 17s+ to the overall runtime.
> 
> It'll probably cost more cycles total when running in parallel, but should save on combined run length.

As expected this job just ran at ±20s but takes an extra ±40s for the container setup. Strangely enough total runtime got decreased to 9m9s, probably quiet time :sweat_smile:

oddhack:
LGTM. There is the possibility of a bit ore speedup by decoupling the hpp steps from the full spec build, since that only needs the headers which only take a few seconds to generate, but this is already a clear improvement and I think we should go ahead. When you're ready please set 'Signed off to Merge' Milestone on this PR (or if you can't do that, assign it to me and comment to that effect), and I'll merge it.

MarijnS95:
> since that only needs the headers

I guess we should do that for the GitLab CI in parallel too then, for consistency (same with decoupling the other step?). What's the exact command, use `makeSpec` with just `registry` to get the C header files?

oddhack:
It's going to be slightly involved to split out the header generation and incorporate it in the hpp-generate step - I'm happy to take that on later in the interests of advancing this now.

PR #1871, opened 2022-06-09
Remove contraction in VK_EXT_non_seamless_cube_map proposal

DadSchoorse:
Just noticed this in the CI.


DadSchoorse:
already fixed in the spec commit

PR #1877, opened 2022-06-17
Add a new driver id VK_DRIVER_ID_MESA_DOZEN

kusma:
This is the driver ID for Dozen, the Vulkan-on-Direct3D driver in Mesa.


jenatali:
@kusma Not sure why you added `WIP` to the title, but if it was to hold off until we got confirmation of whether this should be `MESA_DOZEN` vs `MICROSOFT_DOZEN`, we discussed and agreed that `MESA_DOZEN` was the right option.

kusma:
@jenatali: Yeah, that's exactly the reason! I'll update this on Monday and remove the WIP-status!

jekstrand:
@oddhack, @kusma has dropped the WIP and he, @jenatali and I are agreed on the name.  I think this can go in tomorrow's release unless you think it needs more formal WG sign-off.

PR #1879, opened 2022-06-27
Explicit VU for Uniform being read only

sjfricke:
According to the SPIR-V spec, `Uniform` are not defined as read-only (`UniformConstant` are) and `spirv-val` does validate this, but it is missing an explicit VUID and language in the Vulkan spec

Also combines the 2 VUs talking about which opaque types are read-only to a single VUID
(no need to change the VUID as neither `04661` or `04665` are currently being validated yet https://github.com/KhronosGroup/SPIRV-Tools/issues/4796)

### DIFF
--- a/appendices/spirvenv.txt
+++ b/appendices/spirvenv.txt
@@ -255,8 +255,10 @@ endif::VK_VERSION_1_1[]
     [eq]#v# components
   * [[VUID-{refpage}-OpTypeImage-04661]]
     Objects of types code:OpTypeImage, code:OpTypeSampler,
-    code:OpTypeSampledImage, and arrays of these types must: not be stored
-    to or modified
+    code:OpTypeSampledImage, code:OpTypeAccelerationStructureKHR,
+    and arrays of these types must: not be stored to or modified
+  * Any variable in the code:Uniform storage class must: not be stored to
+    or modified
   * [[VUID-{refpage}-Offset-04662]]
     Any image operation must: use at most one of the code:Offset,
     code:ConstOffset, and code:ConstOffsets image operands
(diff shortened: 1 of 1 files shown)


appendices/spirvenv.txt
```diff
@@ -255,8 +255,10 @@ endif::VK_VERSION_1_1[]
     [eq]#v# components
   * [[VUID-{refpage}-OpTypeImage-04661]]
     Objects of types code:OpTypeImage, code:OpTypeSampler,
-    code:OpTypeSampledImage, and arrays of these types must: not be stored
-    to or modified
+    code:OpTypeSampledImage, code:OpTypeAccelerationStructureKHR,
+    and arrays of these types must: not be stored to or modified
```
r-potter, line 259: Does this need a new VUID since it is effectively collapsing 04661 and 04665 into a single VU statement?

sjfricke: The reason to give a new VUID has always been to notify downstream of the change to the validation implementors... but as posted in the summary, this hasn't been implemented yet and I plan to personally do it, so see no reason to change the VU

appendices/spirvenv.txt
```diff
@@ -255,8 +255,10 @@ endif::VK_VERSION_1_1[]
...
+  * Any variable in the code:Uniform storage class must: not be stored to
```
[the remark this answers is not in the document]
HansKristian-Work: I think so, yes.

appendices/spirvenv.txt
```diff
@@ -255,8 +255,10 @@ endif::VK_VERSION_1_1[]
     [eq]#v# components
   * [[VUID-{refpage}-OpTypeImage-04661]]
     Objects of types code:OpTypeImage, code:OpTypeSampler,
-    code:OpTypeSampledImage, and arrays of these types must: not be stored
-    to or modified
+    code:OpTypeSampledImage, code:OpTypeAccelerationStructureKHR,
```
HansKristian-Work, line 258: The VUID changed here, it refers to OpTypeImage.

sjfricke: Sorry, not sure what you mean, I just added `OpTypeAccelerationStructureKHR` from the other VU to here

oddhack: ITHKM that the VUID should be regenerated since the semantics of the VU statement changed - but since these are usually VUs you yourself are implementing, often that doesn't matter.

sjfricke: just removed the VUID, its just a number and I don't really care what it is, just that there is one to track

HansKristian-Work:
LGTM

PR #1880, opened 2022-06-28
fix mixing up of Flat VUID shader stage

sjfricke:
These 2 VUs seem to have been mixed up, I just created a PR for them assuming this change is correct
https://github.com/KhronosGroup/SPIRV-Tools/pull/4831


HansKristian-Work:
The spec is already correct from what I can tell.

sjfricke:
you are right, sorry, idk how I inversed the `not`

PR #1884, opened 2022-07-05
Fix XML tagging of VkShaderModuleCreateInfo and add explicit VU

### LINKED ISSUE #1883: `VkShaderModuleCreateInfo` missing `pNext` VU
`VkShaderModuleCreateInfo` is missing `pNext` VU. I assume the problem is it has some `structextends` despite the fact that it is 1.0 core struct.

oddhack:
VkShaderModuleCreateInfo appears both as an explicit parameter, and in
the pNext chain of VkPipelineShaderStageCreateInfo, The validity
generator doesn't understand this case fully.

Proposed solution is to add an explicit VU to vkCreateShaderModule
expressing the pNext constraint; make sure implicit VU for pNext are
suppressed; and properly tag the structextends attribute for the single
structure extending VkShaderModuleCreateInfo.

Fixes #1883

### DIFF
--- a/chapters/shaders.txt
+++ b/chapters/shaders.txt
@@ -81,6 +81,13 @@ slink:VkPipelineShaderStageCreateInfo.
 This avoids the overhead of creating and managing an additional object.
 endif::VK_EXT_graphics_pipeline_libraries[]
 
+.Valid Usage
+****
+  * If pname:pCreateInfo is not `NULL`, pname:pCreateInfo->pNext must: be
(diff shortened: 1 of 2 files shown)


krOoze:
This is little hotfixy. And what I am afraid of, a little brittle, and I virtually can guarantee you will forget to update this VU if it gets extended. Plus not sure if Validation Layers will pick this up.

What about adding something like `structextends=_toplevel, VkWhateverStruct` hint.

Issue #1889, opened 2022-07-10
Features should be referenced by their real name


krOoze:
Throughout the specification, features are referenced by colloquial name. That is confusing and hurts searchability. Features should be referenced by their real name.

E.g.:

> If the protected memory feature is not enabled, the `VK_COMMAND_POOL_CREATE_PROTECTED_BIT` bit of `flags` **must** not be set

Should say "`protectedMemory` feature".

sjfricke:
I agree with this as this has made finding VUs by searching the feature name in the spec hard sometimes, also it would be nice if it was consistent.

If the WG also agrees, I went through and found probably most of the places where this would need to be fixed

```
protected memory feature
precise occlusion queries feature
inherited queries feature
pipeline statistics queries
inherited viewport scissor feature
multiple viewports feature 
depth clamping feature
depth bias clamping feature
depth bounds testing feature 
variable multisample rate feature
sparse bindings feature
sparse buffer residency feature
sparse aliased residency feature
multisampled storage images feature
image cube map arrays feature
dedicated allocation image aliasing feature
anisotropic sampling feature
robust buffer access feature 
multi-draw indirect feature 
non-solid fill modes feature
sample rate shading feature
alpha to one feature 
wide lines feature
independent blending feature
logic operations feature
dual source blending feature 

VUID-VkImageCreateInfo-imageType-0097*

stage_mask_common.txt
stage_mask_2_common.txt
```

oddhack:
@sjfricke thanks! Very helpful. Are you signing up to propose a PR or just getting us started down the path?

sjfricke:
I'll do the PR, just wanted to first make sure you also thought it would be a good fix before spent time doing it

PR #1891, opened 2022-07-13
reference features by feature name

sjfricke:
closes https://github.com/KhronosGroup/Vulkan-Docs/issues/1889


r-potter:
We just (this week) had an MR to add `pname:` markup to all existing feature links for consistency. This MR should probably do the same

oddhack (approved):
LGTM. I admit I haven't inspected every single change but all the ones I did, look fine.

oddhack:
@sjfricke I found a few more and those are also in the 1.3.221 spec update, along with a brief style guide addition - thanks!

Probably a similar exercise should be done for limits-* xrefs, although that's a bit more difficult due to increased number of ways the xrefs are imbedded.

PR #1894, opened 2022-07-19
GitHub CI: Regenerate and build-test Rust Vulkan bindings (Ash crate)

MarijnS95:
_This complements a similar change done to the GitLab CI script._

Rust's Vulkan bindings, provided by [the Ash crate], rely on rather strict rules/conventions in vk.xml or otherwise fail to generate.  In order to provide more insight into the effect of spec changes to this project, and hopefully catch mishaps earlier (ie. naming rules for enum variants with respect to the typename of the enum) regenerate and build-test the Rust crate in the CI.

Since Rust bindings are not part of the official Khronos Vulkan SDK / release this check is currently allowed to fail.  Over time this task can be expanded to notify Ash maintainers when PRs merged to `main` break our build, so that we can resolve it quickly or request spec changes before they are released hence set-in-stone.  Additionally release managers can interpret build failures like [enum variant name mismatches] and address them on their own, or for the time being ping the Ash maintainers (me - @MarijnS95) to bring failures to our attention or request clarification/assistance.

[the Ash crate]: https://github.com/ash-rs/ash/
[enum variant name mismatches]: https://github.com/Traverse-Research/ash/blob/c4f1c053ea36e4de91094c63109eeda677ca3ecf/generator/src/lib.rs#L1207-L1210

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -117,6 +117,61 @@ jobs:
           name: hpp-outputs
           path: gen/include/
 
+  ash-generate:
+    name: Generate Rust bindings (Ash crate)
+    runs-on: ubuntu-latest
+    needs: spec-generate
+    continue-on-error: true
+    env:
+      # Cached folder outside of git repo
+      CARGO_TARGET_DIR: ${{ github.workspace }}/ash-target
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+        with:
+          repository: ash-rs/ash
+          path: ash
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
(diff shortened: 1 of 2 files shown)


MarijnS95:
Err, note that the CI script currently has:

https://github.com/KhronosGroup/Vulkan-Docs/blob/d7042a0673f7a091133b562b9e67d7a3297fa458/.github/workflows/CI.yml#L10-L15

Meaning that it only runs for pushes to this repo and not forks (well, it runs on my fork, but you don't see the CI result here). Perhaps we should add `pull_request:` and maybe remove/constrain `push:` if it's only relevant to run that on `main`?

oddhack (approved):
It looks fine, to the extent I can "review" it without actually running CI. Sounds like this will evolve a bit before it's ready to merge, though, to handle the fork/CI scenario you alluded to?

MarijnS95:
Looking good, the entire run succeeded even if individual steps fail!

MarijnS95:
And with that cleaned up, this is now ready to be merged from my side.

PR #1895, opened 2022-07-20
Add missing SPV_KHR_ray_tracing storage class VUID

sjfricke:
Currently, there is code in `spirv-val` to check the `Storage Classes` in `SPV_KHR_ray_tracing` and while I am adding support right now, I realized there is duplicate language in the Vulkan spec that is already well documented in the [SPV_KHR_ray_tracing](https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_ray_tracing.html#_modifications_to_the_spir_v_specification) spec

My argument for the removal is
- The Vulkan Spec was missing `ShaderRecordBufferKHR` so it is not complete in the Vulkan spec currently
- There are much more details missing from the SPIR-V spec missing here (eg. a VU for `HitAttributeKHR` can only be read from `AnyHitKHR` and `ClosestHitKHR`)
- single origin to change in future if another Storage Class interacts with the extension

### DIFF
--- a/appendices/spirvenv.txt
+++ b/appendices/spirvenv.txt
@@ -492,7 +492,7 @@ endif::VK_VERSION_1_1[]
     A variable with code:HitAttributeKHR storage class must: only be written
     to in an intersection shader
   * A variable with code:HitAttributeKHR storage class must: only be read
-    to in an any-hit or closest hit shader
+    from in an any-hit or closest hit shader
   * [[VUID-{refpage}-CallableDataKHR-04704]]
     code:CallableDataKHR storage class must: only be used in ray generation,
     closest hit, miss, and callable shaders
--- a/appendices/spirvenv.txt
+++ b/appendices/spirvenv.txt
@@ -491,6 +491,8 @@ endif::VK_VERSION_1_1[]
   * [[VUID-{refpage}-HitAttributeKHR-04703]]
     A variable with code:HitAttributeKHR storage class must: only be written
     to in an intersection shader
+  * A variable with code:HitAttributeKHR storage class must: only be read
+    to in an any-hit or closest hit shader
   * [[VUID-{refpage}-CallableDataKHR-04704]]
     code:CallableDataKHR storage class must: only be used in ray generation,
     closest hit, miss, and callable shaders
@@ -501,6 +503,8 @@ endif::VK_VERSION_1_1[]
     There must: be at most one variable with the
     code:IncomingCallableDataKHR storage class in the input interface of an
     entry point
+  * code:ShaderRecordBufferKHR storage class must: only be used in ray
+    generation, intersection, any-hit, closest hit, callable, or miss shaders
   * [[VUID-{refpage}-Base-04707]]
     The code:Base operand of code:OpPtrAccessChain must: point to one of the
     following: *Workgroup*, if code:VariablePointers is enabled;
--- a/appendices/spirvenv.txt
+++ b/appendices/spirvenv.txt
@@ -491,8 +491,6 @@ endif::VK_VERSION_1_1[]
   * [[VUID-{refpage}-HitAttributeKHR-04703]]
     A variable with code:HitAttributeKHR storage class must: only be written
     to in an intersection shader
-  * A variable with code:HitAttributeKHR storage class must: only be read
-    from in an any-hit or closest hit shader
   * [[VUID-{refpage}-CallableDataKHR-04704]]
     code:CallableDataKHR storage class must: only be used in ray generation,
     closest hit, miss, and callable shaders


dgkoch:
I'd say the downside of not having this in the Vulkan spec is that it'll be harder to find/easier to miss for validation, and there wouldn't be any VUIDs assigned for the checks.  Will discuss in the Ray TSG though.

sjfricke:
At the end of the day, as long as everything is properly being validated in `spirv-val` (which it is currently not so I want to add) I will be happy 😁, so happy with what the WG decides. (I realize this is the not the first time we hit this situation with SPIR-V related VUs 🙃)

The main thing that caught me off guard at first was there was no VU for `ShaderRecordBufferKHR` so I spent time trying to see if there was something subtle that the Vulkan spec disallowed that was a superset of the SPIR-V validation

dgkoch:
> The main thing that caught me off guard at first was there was no VU for `ShaderRecordBufferKHR` so I spent time trying to see if there was something subtle that the Vulkan spec disallowed that was a superset of the SPIR-V validation

Could we interest you in proposing an MR that adds the stuff that is missing from the SPIR-V VU section instead?

sjfricke:
Of course, will line it up with the Validation PR as well 💪

sjfricke:
@dgkoch hopefully this is good now

appendices/spirvenv.txt
```diff
@@ -491,6 +491,8 @@ endif::VK_VERSION_1_1[]
   * [[VUID-{refpage}-HitAttributeKHR-04703]]
     A variable with code:HitAttributeKHR storage class must: only be written
     to in an intersection shader
+  * A variable with code:HitAttributeKHR storage class must: only be read
+    to in an any-hit or closest hit shader
```
stu-s, line 495:
```suggestion
    from in an any-hit or closest hit shader
```

appendices/spirvenv.txt
```diff
@@ -491,6 +491,8 @@ endif::VK_VERSION_1_1[]
   * [[VUID-{refpage}-HitAttributeKHR-04703]]
     A variable with code:HitAttributeKHR storage class must: only be written
     to in an intersection shader
+  * A variable with code:HitAttributeKHR storage class must: only be read
+    from in an any-hit or closest hit shader
```
dgkoch, line 495: Also intersection (it can be read after being written). 
Depending on exactly how 4703 is parsed, it could be added up there instead..

sjfricke: well "read/written" are taken from the SPIR-V spec

> They can be written to only in IntersectionKHR execution model and read from only in AnyHitKHR and ClosestHitKHR execution models.

but in practice, it turned into `OpLoad`/`OpStore` in the `spirv-val` PR

I guess what is a use case to read the value from the `HitAttributeKHR` variable instead of the variable used to write to it?

dgkoch: I don't believe we have anything in SPIR-V that prohibits from reading back output variables, and we definitely want to allow that at the HLL. In theory you could do a compiler transform that does what you suggested, but I doubt that glslang does that, and I think that is getting us off into the weeds here. 

Looking back at the GLSL we say: 

    4.3.X hitAttributeEXT Variables

    These are allowed only in intersection, any-hit, and closest-hit shaders.
    It is a compile-time error to use them in any other stage. They can be read
    in any-hit, and closest-hit shaders. They can be written to and read from
    only in intersection shaders (values are undefined before being written to).

In the SPIR-V we say: 

    HitAttributeKHR
    ...
    Variables declared with this storage class are allowed only in IntersectionKHR,
    AnyHitKHR and ClosestHitKHR execution models. They can be written to only
    in IntersectionKHR execution model and read from only in AnyHitKHR and
    ClosestHitKHR execution models.

So the intent is to clearly allow R/W in Intersection and R in AnyHit and ClosestHit.

Looking back up a few more lines - we have VUID 04701 which already restricts HitAttributeKHR to Intersection, AnyHit and ClosestHit stages.
So 04703 is trying to say that Intersection is the only stage that is allowed to write to HitAttributeKHR (maybe we need slightly different wording to clarify that).
And your new VU should be trying to say that the only stages allowed to read from HitAttributeKHR are Intersection, AnyHit, and Closest hit.

How about: 

  * [[VUID-{refpage}-HitAttributeKHR-04703]]
    A variable with code:HitAttributeKHR storage class must: not be written to in any stage which is not an intersection shader.
  * A variable with code:HitAttributeKHR storage class must: not be read from in any stage which is not an intersection, any-hit, or closest hit shader.

But then the newly proposed one seems pretty redundant with the existing 04701 VU, so I think my only suggestion here might be to try to reword the 04703 VU, so that it's clear that "writing is only permitted in IS" and NOT "is write-only in IS"

sjfricke: > They can be written to only in IntersectionKHR execution model and read from only in AnyHitKHR and ClosestHitKHR execution models. 

ok... I slightly misread that and "you can ONLY read from those 2 stages" not it was "read only IN those stages"

|   | Any | Closest | Intersection |
|---|-----|---------|--------------|
| W |     |         | X            |
| R | X   | X       | X            |

I think both the `4703` and `4701` already cover this then already! Lets just remove this line and maybe reword the SPIR-V spec to just say

> They can be written to only in IntersectionKHR execution model.

as adding the "read from only" is not need as it would be implied from the above sentence stating it is allowed in all 3 stages

sjfricke, line 495:
```suggestion
```

dgkoch (approved):
LGTM now.

stu-s (approved):
LGTM

dgkoch:
@oddhack please pick this up for the next spec update.

PR #1901, opened 2022-07-25
Adjust wording in vkSetDebugUtilsObjectNameEXT for non-device handles

### LINKED ISSUE #1668: Clarification of `vkSetDebugUtilsObjectName/TagEXT` for instance handles
In section 45.1.1, in the definition of `vkSetDebugUtilsObjectNameEXT`, the `device` parameter is described as follows:
> `device` is the device that created the object.

This, for me, implies that objects passed in `VkDebugUtilsObjectNameInfoEXT::objectHandle` must have been created from that device. However, this can't be the case for instances or handles created from it, like physical devices, or the device itself. There is also no Valid Usage requirement that would require handles to be created from the device.

This also applies to the definition of `vkSetDebugUtilsObjectTagEXT`.

From experimenting, it seems to be the case that it is valid to pass any valid device handle.

Could the description of the `device` parameter in these commands be expanded to define the behaviour of naming handles that weren't created from any device?

pixelcluster:
Closes #1668.

As discussed there, it is currently unclear whether or not it is valid to name handles that are not created from a device. This PR is my attempt at specifying what should happen for those types of handles.

As I see it, there are two types of handles that aren't created from a device: For one, handles that a device can be created from (directly or indirectly), which are `VkPhysicalDevice` and `VkInstance`. The other type is handles that aren't related to the device at all (e. g. `VkSurfaceKHR`).

For the first type, I think all Mesa drivers have an internal struct that the handle points to that contains more data about that struct other than loader data (I'd guess other drivers do the same). I think it is reasonable to assume that if a driver wants to expose a custom implementation of `vkSetDebugUtilsObjectNameEXT`, it might want to access that data. Therefore, it makes sense to forbid apps from passing physical devices of other drivers because accessing the internal data of those devices would likely go horribly wrong there. This is the main purpose of the second VU restriction I added, but it also forbids passing a device and an object handle that were created from different instances.

For `VkSurfaceKHR` (and, I think, all relevant objects of the second type), this restriction does not need to apply since the loader gives a separate "real" surface to each ICD. Only the third VU restriction applies here, which simply forbids passing handles across instances.

I marked this as a draft because no VUID numbers are assigned to the new VU requirements. I tried to assign some myself using the reflow.py script, but was unable to because I was missing the vuidCounts.py file. I'll remove the draft status once this is resolved.

### DIFF
--- a/chapters/VK_EXT_debug_utils.txt
+++ b/chapters/VK_EXT_debug_utils.txt
@@ -42,7 +42,14 @@ fname:vkSetDebugUtilsObjectNameEXT as defined below.
 --
 include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
-  * pname:device is the device that created the object.
+  * pname:device is the name of a valid slink:VkDevice which is related to the object in
+    pname:pNameInfo.
+    If pname:pNameInfo references a slink:VkInstance or a slink:VkPhysicalDevice, 
+    pname:device must: be created from that object.
+    If pname:pNameInfo references any other instance-level object, that object and
+    pname:device must: be created from the same slink:VkInstance.
+    Otherwise, the object passed in pname:pNameInfo must: have been created from
+    the same slink:VkDevice ancestor as listed in pname:device.
   * pname:pNameInfo is a pointer to a slink:VkDebugUtilsObjectNameInfoEXT
     structure specifying parameters of the name to set on the object.
 
@@ -52,6 +59,16 @@ include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
     pname:pNameInfo->objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
   * [[VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-02588]]
     pname:pNameInfo->objectHandle must: not be dlink:VK_NULL_HANDLE
+  * If the object referred to by pname:pNameInfo->objectHandle was
+    created, allocated or retrieved from a slink:VkDevice,
+    pname:device must: be the device the object was created,
+    allocated or retrieved from.
+  * If pname:pNameInfo->objectType is ename:VK_OBJECT_TYPE_INSTANCE or
+    ename:VK_OBJECT_TYPE_PHYSICAL_DEVICE, pname:device must: have been
+    created from the object referred to by pname:pNameInfo->objectHandle.
+  * If pname:pNameInfo->objectType is not ename:VK_OBJECT_TYPE_INSTANCE,
+    both of pname:device, and pname:pNameInfo->objectHandle must: have the
+    same slink:VkInstance as their ancestor.
 ****
 
 include::{generated}/validity/protos/vkSetDebugUtilsObjectNameEXT.txt[]


oddhack:
@KarenGhavam-lunarG could you reassign this to whomever is currently responsible for this extension?

chapters/VK_EXT_debug_utils.txt
```diff
@@ -42,7 +42,11 @@ fname:vkSetDebugUtilsObjectNameEXT as defined below.
 --
 include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
-  * pname:device is the device that created the object.
+  * pname:device is the device that created the object, or, if the device is a
```
r-potter, line 45:
```suggestion
  * pname:device is the device that created the object, or, if the object is a
```
Was this intended to be `object` rather than `device`? I'm having difficultly parsing the sentence otherwise

chapters/VK_EXT_debug_utils.txt
```diff
@@ -52,6 +56,19 @@ include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
...
+  * [[VUID-vkSetDebugUtilsObjectNameEXT-device]]
+    If the object referred to by pname:pNameInfo->objectHandle was
+    created, allocated or retrieved from a slink:VkDevice,
+    pname:device must: be the device the object was created,
+    allocated or retrieved from.
+  * [[VUID-vkSetDebugUtilsObjectNameEXT-device]]
+    If pname:pNameInfo->objectType is ename:VK_OBJECT_TYPE_INSTANCE or
+    ename:VK_OBJECT_TYPE_PHYSICAL_DEVICE, pname:device must: have been
+    created from the object referred to by pname:pNameInfo->objectHandle.
+  * [[VUID-vkSetDebugUtilsObjectNameEXT-device]]
```
MarkY-LunarG, line 68: All of these VUID's are the same.  I'm not sure that's allowed.  I think you need them generated so they have unique values

chapters/VK_EXT_debug_utils.txt
```diff
@@ -42,7 +42,11 @@ fname:vkSetDebugUtilsObjectNameEXT as defined below.
 --
 include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
-  * pname:device is the device that created the object.
+  * pname:device is the device that created the object, or, if the device is a
+    slink:VkInstance or a slink:VkPhysicalDevice, a device that was created from the object.
+    If the object does not have any relationship to a slink:VkDevice (for example a
+    slink:VkSurfaceKHR), it is a device that was created from the same slink:VkInstance as
+    the object.
```
MarkY-LunarG, line 49: Perhaps the following sentence might be better...

```
* pname:device is the name of a valid slink:VkDevice which is related to the object in
  pname:pNameInfo.
  If pname:pNameInfo references an instance-level object (for example a
  slink:VkInstance or slink:VkPhysicalDevice), pname:device must be created from that object.
  Otherwise, the object passed in pname:pNameInfo must have been created from
  the same slink:VkDevice ancestor as listed in pname:device.
```
Or something like that.  I don't know if there's too much in that statement as well.

chapters/VK_EXT_debug_utils.txt
pixelcluster: Oops, indeed.

chapters/VK_EXT_debug_utils.txt
pixelcluster: It seems I misread the documentation in vuid.txt, I thought this was the required header in order to generate the VUIDs for merging. Just removing the `[[VUID-*]]` blocks (while keeping the asterisk before the statement) would work?

chapters/VK_EXT_debug_utils.txt
pixelcluster: Good idea, this would be clearer than what I have at the moment. Your statement about instance-level objects might be problematic though, because there are some instance-level objects one can't create a device from (for example `VkSurfaceKHR`). I would like to split these instance-level objects from `VkInstance` and `VkPhysicalDevice`. Maybe
```
* pname:device is the name of a valid slink:VkDevice which is related to the object in
  pname:pNameInfo.
  If pname:pNameInfo references a slink:VkInstance or a slink:VkPhysicalDevice, 
  pname:device must be created from that object.
  If pname:pNameInfo references any other instance-level object, that object and
  pname:device must be created from the same slink:VkInstance.
  Otherwise, the object passed in pname:pNameInfo must have been created from
  the same slink:VkDevice ancestor as listed in pname:device.
```
would work?

chapters/VK_EXT_debug_utils.txt
oddhack: That's right. If you will point to the misleading statement I'll attempt to clarify it.

pixelcluster: Upon rereading, everything was pretty obvious, I don't think there is a need to clarify it :)
I think I can also remove the draft status now. Thanks for helping!

oddhack:
@MarkY-LunarG are you able to sign off on this? @r-potter is OK with it.

chapters/VK_EXT_debug_utils.txt
MarkY-LunarG: That sounds great to me.  Sorry for the delayed response.

pixelcluster: Done.

MarkY-LunarG (approved):

chapters/VK_EXT_debug_utils.txt
```diff
@@ -52,6 +59,16 @@ include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
...
+  * If the object referred to by pname:pNameInfo->objectHandle was
+    created, allocated or retrieved from a slink:VkDevice,
+    pname:device must: be the device the object was created,
+    allocated or retrieved from.
+  * If pname:pNameInfo->objectType is ename:VK_OBJECT_TYPE_INSTANCE or
+    ename:VK_OBJECT_TYPE_PHYSICAL_DEVICE, pname:device must: have been
+    created from the object referred to by pname:pNameInfo->objectHandle.
+  * If pname:pNameInfo->objectType is not ename:VK_OBJECT_TYPE_INSTANCE,
+    both of pname:device, and pname:pNameInfo->objectHandle must: have the
+    same slink:VkInstance as their ancestor.
```
krOoze, line 71: What if they have the same Instance as ancestor but not Physical Device, as it can happen for `VkDisplayKHR`?

VUs should not end with periods.

chapters/VK_EXT_debug_utils.txt
```diff
@@ -42,7 +42,14 @@ fname:vkSetDebugUtilsObjectNameEXT as defined below.
 --
 include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
-  * pname:device is the device that created the object.
+  * pname:device is the name of a valid slink:VkDevice which is related to the object in
+    pname:pNameInfo.
+    If pname:pNameInfo references a slink:VkInstance or a slink:VkPhysicalDevice, 
+    pname:device must: be created from that object.
+    If pname:pNameInfo references any other instance-level object, that object and
+    pname:device must: be created from the same slink:VkInstance.
```
krOoze, line 50: Looks like valid usage. Should not be here among the prose.

chapters/VK_EXT_debug_utils.txt
```diff
@@ -42,7 +42,14 @@ fname:vkSetDebugUtilsObjectNameEXT as defined below.
 --
 include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.txt[]
 
-  * pname:device is the device that created the object.
+  * pname:device is the name of a valid slink:VkDevice which is related to the object in
+    pname:pNameInfo.
```
krOoze, line 46: Should say what the parameter is directly, not what its relation to other things is. I would keep the cookie-cutter original, because what it is is a dispatch parameter as usual for all the commands...

krOoze:
`vkSetDebugUtilsObjectTagEXT` would have identical problems, right?

chapters/VK_EXT_debug_utils.txt
MarkY-LunarG: First, these are very good points. 

Background: (which you probably already know but I'll put here for those who don't).  The original design simply copied the VK_EXT_debug_marker functionality but into a common extension with the old VK_EXT_debug_report to ease sharing of types and functionality which was proving difficult in the original two extensions.  In addition to expanding additional output info (like the object types).  Originally, it was believe the developers wouldn't need this functionality for instance and physical devices. 

The only other thing we could do here if we don't want to relax this as the modifications are allowing is to create a brand new extension that would allow debug labels for Instance and physical device. 

Personally, I'm ok with allowing the tracking of instance and physical device data within this entrypoint, but I think you're point of clearly definnig the lifecycle is important and it should be noted somewhere any instance/physical device tracking will go away with the device.  Or that it goes away with the destruction of the messenger.  My concern is that if we allow it for the messenger it could open more cans of worms (like if different devices labelled the instance differently).

MarkY-LunarG:
> vkSetDebugUtilsObjectTagEXT would have identical problems, right?

Good point.

chapters/VK_EXT_debug_utils.txt
MarkY-LunarG: I wonder if the last bullet point is even needed.  Or maybe

```
 * If pname:pNameInfo->objectType is an instance object (like ename:VK_OBJECT_TYPE_PHYSICAL_DEVICE)  then both of pname:device and pname:pNameInfo->objectHandle must have the same slink:VkInstance as their ancestor
```

Slightly diffierent, but I think satsifies @krOoze's concern?  Or do we not differentiate instance and device objects in that way?

pixelcluster:
Superseded by #2034.

PR #1905, opened 2022-07-28
Drop wayland-client.h include

jengelh:
With https://github.com/KhronosGroup/Vulkan-Tools/pull/670 resolved, wayland-client.h can be removed from Headers. This has been cooking in openSUSE for a while, and other vulkan infrastructure (vulkan-{tools,validationlayers}, glslang, spirv-tools, shaderc) still build fine.

### DIFF
--- a/include/vulkan/vulkan.h
+++ b/include/vulkan/vulkan.h
@@ -38,7 +38,6 @@
 
 
 #ifdef VK_USE_PLATFORM_WAYLAND_KHR
-#include <wayland-client.h>
 #include "vulkan_wayland.h"
 #endif
 


fooishbar:
My preferred solution would be to forward-declare the struct types required (i.e. `wl_display` and `wl_surface`) inside `vulkan_wayland.h`, rather than just completely dropping it here. Without doing that, we're requiring people to include both `wayland-client-core.h` (for `wl_display`) and `wayland-client-protocol.h` (for `wl_surface`) everywhere that includes `vulkan.h` when `VK_USE_PLATFORM_WAYLAND_KHR` is set (which may not necessarily touch Wayland), if they don't want to get compliation errors.

stonesthrow:
@chadversary@google.com, for WSI discussion.

oddhack:
> My preferred solution would be to forward-declare the struct types required (i.e. `wl_display` and `wl_surface`) inside `vulkan_wayland.h`, rather than just completely dropping it here. Without doing that, we're requiring people to include both `wayland-client-core.h` (for `wl_display`) and `wayland-client-protocol.h` (for `wl_surface`) everywhere that includes `vulkan.h` when `VK_USE_PLATFORM_WAYLAND_KHR` is set (which may not necessarily touch Wayland), if they don't want to get compliation errors.

Somewhat to my surprise, it seems the declaration of pointers to these inside the Vulkan struct type counts as a forward declaration. Frex AFAICT this is legal C99 and C++:

```
typedef struct {
    struct wl_display *d;
    struct wl_surface *s;
} aggregate;

aggregate A = { 0, 0 };
```

Although, someone trying to generate a binding to another language might be unpleasantly surprised by this usage.

oddhack:
This should be rebased on main branch to fix the CI issue - we ran into some Docker container caching problems and can't find a way to clear the Actions container cache.

jengelh:
I believe I had enabled [X] Enable edits by maintainers.

fooishbar:
> > My preferred solution would be to forward-declare the struct types required (i.e. `wl_display` and `wl_surface`) inside `vulkan_wayland.h`, rather than just completely dropping it here. Without doing that, we're requiring people to include both `wayland-client-core.h` (for `wl_display`) and `wayland-client-protocol.h` (for `wl_surface`) everywhere that includes `vulkan.h` when `VK_USE_PLATFORM_WAYLAND_KHR` is set (which may not necessarily touch Wayland), if they don't want to get compliation errors.
> 
> Somewhat to my surprise, it seems the declaration of pointers to these inside the Vulkan struct type counts as a forward declaration. Frex AFAICT this is legal C99 and C++:
> 
> ```
> typedef struct {
>     struct wl_display *d;
>     struct wl_surface *s;
> } aggregate;
> 
> aggregate A = { 0, 0 };
> ```

Fascinating. I indeed can't find a way to make gcc warn about this.

chadversary:
Discussed today at the SI meeting.

The MR looks good. But first, this Friday, I'll upload a MR that adds explicit forward declarations to `vulkan_wayland.h`. After that MR gets merged, we will merge this one.

oddhack:
> Discussed today at the SI meeting.
> 
> The MR looks good. But first, this Friday, I'll upload a MR that adds explicit forward declarations to `vulkan_wayland.h`. After that MR gets merged, we will merge this one.

@chadversary did this MR get proposed? I don't recall seeing it internally. TBH not seeing why explicit forward declarations are needed, per my experiment above.

Issue #1832, opened 2022-04-25
Feature Request: Extension for specifying constraints on the destination address of vkMapMemory.


Guy1524:
Background: [The Wine project](https://www.winehq.org/about) is currently progressing towards a new method of supporting 32-bit windows applications within processes matching the host's architecture.  In practice, this means supporting 32-bit consumers of the Vulkan (and other) APIs using 64-bit Vulkan drivers.  One complication in this process is relaying the pointer returned from vkMapMemory to the 32-bit API consumer, if the pointer happens to fall out of the 32-bit range.

To help solve this problem, an extension could be introduced allowing applications to either specify where they want a mapping to be placed, or to constrain the possible area where resources can be mapped to.  One possibility would be a adding a flag to VkMemoryMapFlags specifying that the resource should be mapped into the lower 32-bit address space.  An implementation using mmap could then simply pass this through by setting the MAP_32BIT flag.

Does this sound appropriate for a Vulkan extension?  I think it's worth noting that we are discussing other opportunities to solve our problem like remapping the returned pointer to a compatible address, but as this solution wouldn't involve any implicit assumptions on how the driver maps memory, it could be the best long-term solution for us.

More details and discussion of the problem faced on the wine mailing list: https://www.winehq.org/pipermail/wine-devel/2022-April/214733.html

jekstrand:
Assuming `MAP_32BIT` works (which it probably will), this should be a trivial extension to draft and implement.  Do you have any interest in trying your hand at some extension text?  Or are you looking for volunteers?

To help with figuring out if this will work at all, I've thrown together a driver branch for you that should work on Intel, at least in theory: https://gitlab.freedesktop.org/jekstrand/mesa/-/commits/anv/VK_EXT_mmap_32_bit. On modern kernels, the Intel drivers pass everything through the standard `mmap()` ioctl so I think `MAP_32BIT` should work.  I've not written any detection code about exactly when to expose the feature yet but that should be enough that you can play around.

zfigura:
I think it'd be better to be able to specify a fixed address, and not rely (solely) on MAP_32BIT. The latter is not portable, and we'd also like the ability to restrict the address to the first 3 or 4 GiB if the process is compatible.

jekstrand:
[...]

PR #1906, opened 2022-07-28
Add a new extension: VK_EXT_map_memory_placed

gfxstrand:
This extension adds support for requesting that a `VkDeviceMemory` object be mapped at a particular location in the client's address space.  This is similar to the `addr` parameter to the `mmap()` ioctl.  In order for this to succeed, the client **may** have to first reserve the address range via a platform-specific mechanism.

Fixes: https://github.com/KhronosGroup/Vulkan-Docs/issues/1832

## External links
 - [Intel Linux implementation](https://gitlab.freedesktop.org/jekstrand/mesa/-/commits/anv/VK_EXT_map_memory_placed)
 - [Crucible test](https://gitlab.freedesktop.org/jekstrand/crucible/-/commits/VK_EXT_memory_map_placed)

### DIFF
--- a/appendices/VK_EXT_map_memory_placed.adoc
+++ b/appendices/VK_EXT_map_memory_placed.adoc
@@ -1,0 +1,33 @@
+// Copyright 2022-2023 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_EXT_map_memory_placed.adoc[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2023-03-21
+*IP Status*::
+    No known IP claims.
+*Interactions and External Dependencies*::
+  - Depends on apitext:VK_KHR_map_memory2
+  - Interacts with apitext:VK_EXT_external_memory_host
+*Contributors*::
+  - Faith Ekstrand, Collabora
+  - Tobias Hector, AMD
+  - James Jones, NVIDIA
+  - Georg Lehmann, Valve
+  - Derek Lesho, Codeweavers
+
+=== Description
+
+This extension allows a client to request that flink:vkMapMemory2KHR
+attempt to place the memory map at a particular virtual address.
+
+include::{generated}/interfaces/VK_EXT_map_memory_placed.adoc[]
+
+=== Version History
+
+  * Revision 0, 2024-01-14 (Faith Ekstrand)
+  ** Internal revisions
--- a/chapters/limits.adoc
+++ b/chapters/limits.adoc
@@ -4566,6 +4566,30 @@ include::{generated}/validity/structs/VkPhysicalDeviceRenderPassStripedPropertie
 
 endif::VK_ARM_render_pass_striped[]
 
+ifdef::VK_EXT_map_memory_placed[]
+[open,refpage='VkPhysicalDeviceMapMemoryPlacedPropertiesEXT',desc='Structure describing the alignment requirements of placed memory maps for a physical device',type='structs']
+--
+The sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT structure is
+defined as:
+
+include::{generated}/api/structs/VkPhysicalDeviceMapMemoryPlacedPropertiesEXT.adoc[]
+
+The members of the sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT
+structure describe the following:
+
+  * [[limits-minPlacedMemoryMapAlignment]]
+    pname:minPlacedMemoryMapAlignment is the minimum alignment required for
+    memory object offsets and virtual address ranges when using placed
+    memory mapping.
+
+:refpage: VkPhysicalDeviceMapMemoryPlacedPropertiesEXT
+include::{chapters}/limits.adoc[tag=limits_desc]
+
+include::{generated}/validity/structs/VkPhysicalDeviceMapMemoryPlacedPropertiesEXT.adoc[]
+--
+endif::VK_EXT_map_memory_placed[]
+
+
 [[limits-minmax]]
 == Limit Requirements
 
@@ -5472,6 +5496,9 @@ ifdef::VK_ARM_render_pass_striped[]
 | pname:renderPassStripeGranularity               | - | (64,64) | max
 | pname:maxRenderPassStripes                      | - | 32 | min
 endif::VK_ARM_render_pass_striped[]
+ifdef::VK_EXT_map_memory_placed[]
+| pname:minPlacedMemoryMapAlignment     | - | 65536 | max
+endif::VK_EXT_map_memory_placed[]
 |====
 
 1::
--- a/chapters/memory.adoc
+++ b/chapters/memory.adoc
@@ -3490,7 +3490,13 @@ include::{generated}/api/protos/vkMapMemory.adoc[]
   * pname:size is the size of the memory range to map, or
     ename:VK_WHOLE_SIZE to map from pname:offset to the end of the
     allocation.
+ifndef::VK_EXT_map_memory_placed[]
   * pname:flags is reserved for future use.
+endif::VK_EXT_map_memory_placed[]
+ifdef::VK_EXT_map_memory_placed[]
+  * pname:flags is a bitmask of elink:VkMemoryMapFlagBits specifying
+    additional parameters of the memory map operation.
+endif::VK_EXT_map_memory_placed[]
   * pname:ppData is a pointer to a code:void* variable in which a
     host-accessible pointer to the beginning of the mapped range is
     returned.
@@ -3576,17 +3582,38 @@ ifdef::VK_KHR_device_group[]
   * [[VUID-vkMapMemory-memory-00683]]
     pname:memory must: not have been allocated with multiple instances
 endif::VK_KHR_device_group[]
+ifdef::VK_EXT_map_memory_placed[]
+  * ename:VK_MEMORY_MAP_PLACED_BIT_EXT must: not be set in pname:flags
+endif::VK_EXT_map_memory_placed[]
 ****
 
 include::{generated}/validity/protos/vkMapMemory.adoc[]
 --
 
-[open,refpage='VkMemoryMapFlags',desc='Reserved for future use',type='flags']
+ifdef::VK_EXT_map_memory_placed[]
+[open,refpage='VkMemoryMapFlagBits',desc='Bitmask specifying additional parameters of a memory map',type='enums']
+--
+Bits which can: be set in flink:vkMapMemory::pname:flags and
+slink:VkMemoryMapInfoKHR::pname:flags, specifying additional properties of
+a memory map, are:
+
+include::{generated}/api/enums/VkMemoryMapFlagBits.adoc[]
+
+  * ename:VK_MEMORY_MAP_PLACED_BIT_EXT requests that the implementation
+    place the memory map at the virtual address specified by the client via
+    slink:VkMemoryMapPlacedInfoEXT::pname:pPlacedAddress, replacing any
+    existing mapping at that address.
+    This flag must: not be used with flink:vkMapMemory as there is no way
+    to specify the placement address.
+--
+endif::VK_EXT_map_memory_placed[]
+
+[open,refpage='VkMemoryMapFlags',desc='Bitmask of VkMemoryMapFlagBits',type='flags']
 --
 include::{generated}/api/flags/VkMemoryMapFlags.adoc[]
 
-tname:VkMemoryMapFlags is a bitmask type for setting a mask, but is
-currently reserved for future use.
+tname:VkMemoryMapFlags is a bitmask type for setting a mask of zero or more
+elink:VkMemoryMapFlagBits.
 --
 
 ifdef::VK_KHR_map_memory2[]
@@ -3622,7 +3649,8 @@ include::{generated}/api/structs/VkMemoryMapInfoKHR.adoc[]
   * pname:sType is a elink:VkStructureType value identifying this structure.
   * pname:pNext is `NULL` or a pointer to a structure extending this
     structure.
-  * pname:flags is reserved for future use.
+  * pname:flags is a bitmask of elink:VkMemoryMapFlagBits specifying
+    additional parameters of the memory map operation.
   * pname:memory is the slink:VkDeviceMemory object to be mapped.
   * pname:offset is a zero-based byte offset from the beginning of the
     memory object.
@@ -3649,11 +3677,89 @@ ifdef::VK_KHR_device_group[]
   * [[VUID-VkMemoryMapInfoKHR-memory-07963]]
     pname:memory must: not have been allocated with multiple instances
 endif::VK_KHR_device_group[]
+ifdef::VK_EXT_map_memory_placed[]
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags, the
+    <<features-memoryMapPlaced, pname:memoryMapPlaced>> feature must: be
+    enabled
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags, the
+    pname:pNext chain must: include a slink:VkMemoryMapPlacedInfoEXT
+    structure and sname:VkMemoryMapPlacedInfoEXT::pname:pPlacedAddress
+    must: not be `NULL`
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is not enabled, pname:offset must: be zero
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is not enabled, pname:size must: be ename:VK_WHOLE_SIZE
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is enabled, pname:offset must: be aligned to an integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is enabled, pname:size must: be ename:VK_WHOLE_SIZE or be aligned to an
+    integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
+ifdef::VK_EXT_external_memory_host[]
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags, the memory
+    object must: not have been imported from a handle type of
+    ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_ALLOCATION_BIT_EXT or
+    ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_MAPPED_FOREIGN_MEMORY_BIT_EXT
+endif::VK_EXT_external_memory_host[]
+endif::VK_EXT_map_memory_placed[]
 ****
 
 include::{generated}/validity/structs/VkMemoryMapInfoKHR.adoc[]
 --
 endif::VK_KHR_map_memory2[]
+
+ifdef::VK_EXT_map_memory_placed[]
+[open,refpage='VkMemoryMapPlacedInfoEXT',desc='Structure containing memory map placement parameters',type='structs']
+--
+If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in
+sname:VkMemoryMapInfoKHR::pname:flags and the pname:pNext chain of
+slink:VkMemoryMapInfoKHR includes a sname:VkMemoryMapPlacedInfoEXT
+structure, then that structure specifies the placement address of the
+memory map.
+The implementation will place the memory map at the specified address,
+replacing any existing maps in the specified memory range.
+Replacing memory maps in this way does not implicitly unmap Vulkan memory
+objects.
+Instead, the client must: ensure no other Vulkan memory objects are mapped
+anywhere in the specified virtual address range.
+If successful, pname:ppData will be set to the same value as
+sname:VkMemoryMapPlacedInfoEXT::pname:pPlacedAddress and
+fname:vkMapMemory2KHR will return ename:VK_SUCCESS.
+If it cannot place the map at the requested address for any
+reason, the memory object is left unmapped and fname:vkMapMemory2KHR will
+return ename:VK_ERROR_MEMORY_MAP_FAILED.
+
+The sname:VkMemoryMapPlacedInfoEXT structure is defined as:
+
+include::{generated}/api/structs/VkMemoryMapPlacedInfoEXT.adoc[]
+
+  * pname:sType is the type of this structure.
+  * pname:pNext is `NULL` or a pointer to a structure extending this
+    structure.
+  * pname:pPlacedAddress is the virtual address at which to place the
+    address.  If sname:VkMemoryMapInfoKHR::pname:flags does not contain
+    ename:VK_MEMORY_MAP_PLACED_BIT_EXT, this value is ignored.
+
+.Valid Usage
+****
+  * If sname:VkMemoryMapInfoKHR::pname:flags contains
+    ename:VK_MEMORY_MAP_PLACED_BIT_EXT, pname:pPlacedAddress must: not be
+    `NULL`
+  * pname:pPlacedAddress must: be aligned to an integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
+  * The address range specified by pname:pPlacedAddress and
+    sname:VkMemoryMapInfoKHR::pname:size must: not overlap any existing
+    Vulkan memory object mapping.
+****
+
+include::{generated}/validity/structs/VkMemoryMapPlacedInfoEXT.adoc[]
+--
+endif::VK_EXT_map_memory_placed[]
 
 Two commands are provided to enable applications to work with non-coherent
 memory allocations: fname:vkFlushMappedMemoryRanges and
@@ -3889,24 +3995,60 @@ include::{generated}/api/structs/VkMemoryUnmapInfoKHR.adoc[]
   * pname:sType is a elink:VkStructureType value identifying this structure.
   * pname:pNext is `NULL` or a pointer to a structure extending this
     structure.
-  * pname:flags is reserved for future use.
+  * pname:flags is a bitmask of elink:VkMemoryUnmapFlagBitsKHR specifying
+    additional parameters of the memory map operation.
   * pname:memory is the slink:VkDeviceMemory object to be unmapped.
 
 .Valid Usage
 ****
   * [[VUID-VkMemoryUnmapInfoKHR-memory-07964]]
     pname:memory must: be currently host mapped
+ifdef::VK_EXT_map_memory_placed[]
+  * If ename:VK_MEMORY_UNMAP_RESERVE_BIT_EXT is set in pname:flags,
+    the <<features-memoryUnmapReserve, pname:memoryUnmapReserve>>
+    must: be enabled
+ifdef::VK_EXT_external_memory_host[]
+  * If ename:VK_MEMORY_UNMAP_RESERVE_BIT_EXT is set in pname:flags, the
+    memory object must: not have been imported from a handle type of
+    ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_ALLOCATION_BIT_EXT or
+    ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_MAPPED_FOREIGN_MEMORY_BIT_EXT
+endif::VK_EXT_external_memory_host[]
+endif::VK_EXT_map_memory_placed[]
 ****
 
 include::{generated}/validity/structs/VkMemoryUnmapInfoKHR.adoc[]
 --
 
-[open,refpage='VkMemoryUnmapFlagsKHR',desc='Reserved for future use',type='flags']
+ifdef::VK_EXT_map_memory_placed[]
+[open,refpage='VkMemoryUnmapFlagBitsKHR',desc='Bitmask specifying additional parameters of a memory unmap',type='enums']
+--
+Bits which can: be set in slink:VkMemoryUnmapInfoKHR::pname:flags, specifying
+additional properties of a memory unmap, are:
+
+include::{generated}/api/enums/VkMemoryUnmapFlagBitsKHR.adoc[]
+
+  * ename:VK_MEMORY_UNMAP_RESERVE_BIT_EXT requests that virtual address
+    range currently occupied by the memory map remain reserved after the
+    flink:vkUnmapMemory2KHR call completes.
+    Future system memory map operations or calls to flink:vkMapMemory or
+    flink:vkMapMemory2KHR will not return addresses in that range unless
+    the range has since been unreserved by the client or the mapping is
+    explicitly placed in that range by calling flink:vkMapMemory2KHR with
+    ename:VK_MEMORY_MAP_PLACED_BIT_EXT, or doing the system memory map
+    equivalent.
+    When ename:VK_MEMORY_UNMAP_RESERVE_BIT_EXT is set, the memory unmap
+    operation may: fail, in which case the memory object will remain host
+    mapped and flink:vkUnmapMemory2KHR will return
+    ename:VK_ERROR_MEMORY_MAP_FAILED.
+--
+endif::VK_EXT_map_memory_placed[]
+
+[open,refpage='VkMemoryUnmapFlagsKHR',desc='Bitmask of VkMemoryUnmapFlagBitsKHR',type='flags']
 --
 include::{generated}/api/flags/VkMemoryUnmapFlagsKHR.adoc[]
 
-tname:VkMemoryMapFlagsKHR is a bitmask type for setting a mask, but is
-currently reserved for future use.
+tname:VkMemoryUnmapFlagsKHR is a bitmask type for setting a mask of zero or
+more elink:VkMemoryUnmapFlagBitsKHR.
 --
 endif::VK_KHR_map_memory2[]
 
--- a/proposals/VK_EXT_map_memory_placed.adoc
+++ b/proposals/VK_EXT_map_memory_placed.adoc
@@ -1,0 +1,180 @@
+// Copyright 2022-2023 The Khronos Group, Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+# VK_EXT_map_memory_placed
+:toc: left
+:refpage: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/
+:sectnums:
+
+This document proposes adding support client-controlled virtual address
+placement of `VkDeviceMemory` maps.
+
+## Problem Statement
+
+There are certain cases in which it can be useful to directly control the
+address used by a memory map operation.
+One example of this is in emulation environments when the application under
+emulation is 32-bit but the host userspace is 64-bit.
+In order to avoid additional copies or address translation, the emulator
+can choose an address in the lower 32 bits of the address space and request
+that the map be placed there.
+
+This functionality is already supported by the standard UNIX `mmap()`
+system call which provides an address hint flag as its first parameter as
+well as a `MAP_FIXED` flag which causes the `mmap()` to fail if it cannot
+place the map in exactly that location.
+This extension proposes to add similar functionality to `vkMapMemory()`.
+
+
+## Solution Space
+
+It is possible to emulate this functionality using
+VK_EXT_external_memory_host by creating a placed memory mapping in some
+other way (such as a Linux memfd combined with conventional `mmap()`) and
+then importing the resulting map via VK_EXT_external_memory_host.
+However, this requires every mappable Vulkan memory allocation to be a host
+pointer import which places restrictions available memory types and heaps
+and may have significant performance costs.
+
+One alternative approach to the emulation use-case would be to add a
+`VK_MEMORY_MAP_32BIT_ADDRESS_BIT_EXT` which does exactly what emulation
+wants and places the map in the lower 32 bits.
+This is equivalent to the Linux `MAP_32BIT` flag.
+However, this is less flexible and may be difficult to implement on Arm
+platforms where Linux `mmap()` does not support `MAP_32BIT` since drivers
+are likely relying on `mmap()` to implement `vkMapMemory()`.
+
+## Proposal
+
+### Dependencies
+
+ - VK_EXT_map_memory2
+
+### API Features
+
+The following features are exposed by this extension:
+
+[source,c]
+----
+typedef struct VkPhysicalDeviceMapMemoryPlacedFeaturesEXT {
+    VkStructureType    sType;
+    void*              pNext;
+    VkBool32           memoryMapPlaced;
+    VkBool32           memoryMapRangePlaced;
+    VkBool32           memoryUnmapReserve;
+} VkPhysicalDeviceMapMemoryPlacedFeaturesEXT;
+
+typedef struct VkPhysicalDeviceMapMemoryPlacedPropertiesEXT {
+    VkStructureType    sType;
+    void*              pNext;
+    VkDeviceSize       minPlacedMemoryMapAlignment;
+} VkPhysicalDeviceMapMemoryPlacedPropertiesEXT;
+
+typedef enum VkMemoryMapFlagBits {
+    VK_MEMORY_MAP_PLACED_BIT_EXT = 0x00000001,
+} VkMemoryMapFlagBits;
+
+typedef struct VkMemoryMapPlacedInfoEXT {
+    VkStructureType    sType;
+    const void*        pNext;
+    void*              pPlacedAddress;
+} VkMemoryMapPlacedInfoEXT;
+
+typedef enum VkMemoryUnmapFlagBitsKHR {
+    VK_MEMORY_UNMAP_RESERVE_BIT_EXT = 0x00000001,
+} VkMemoryMapFlagBitsKHR;
+----
+
+ . `memoryMapPlaced` indicates support for placed memory maps
+ . `memoryMapRangePlaced` indicates support for placed memory maps of
+   subranges of the `VkDeviceMemory` object.
+ . `memoryUnmapReserve` indicates support for
+   `VK_MEMORY_UNMAP_RESERVE_BIT_EXT`.
+
+When `VK_MEMORY_MAP_PLACED_BIT_EXT` is set in the `flags` member of
+`VkMemoryMapInfoKHR` and a `VkMapMemoryPlacedInfoEXT` structure is included
+in the `pNext` chain of `VkMemoryMapInfoKHR`, the implementation attempts
+to place the map at the address specified by
+`VkMapMemoryPlacedInfoEXT::pPlacedAddress`.
+If the map cannot be placed exactly at the specified virtual address,
+`vkMapMemory2KHR()` returns `VK_ERROR_MEMORY_MAP_FAILED`.
+
+When `VK_MEMORY_UNMAP_RESERVE_BIT_EXT` is set in the `flags` member of
+`VkMemoryUnmapInfoEXT`, the memory object will be unmapped but the
+address range will remain reserved so that another call to `mmap()` or
+`vkMapMemory()` will not re-use that address range without an address hint
+or `VK_MEMORY_MAP_PLACED_BIT_EXT`, respectively.
+
+## Examples
+
+The following example allocates a memory object and maps it in the lower 32
+bit address space, using `mmap()` to select the address.
+
+[source,c]
+----
+VkDeviceMemory mem;
+const VkMemoryAllocateInfo allocInfo = {
+    .sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO,
+   .allocationSize = /* compute size */,
+   .memoryTypeIndex = /* compute memory type */,
+};
+VK_CHECK(vkAllocateMemory(device, &allocInfo, NULL, &mem));
+
+void *reserved = mmap(NULL, allocInfo.allocationSize, PROT_NONE,
+                      MAP_ANONYMOUS | MAP_PRIVATE | MAP_32BIT, -1, 0);
+CHECK(reserved != MAP_FAILED);
+
+const VkMemoryMapPlacedInfoEXT mapPlacedInfo = {
+    .sType = VK_STRUCTURE_TYPE_MEMORY_MAP_PLACED_INFO_EXT,
+    .pPlacedAddress = reserved,
+};
+const VkMemoryMapInfoKHR mapInfo = {
+    .sType = VK_STRUCTURE_TYPE_MEMORY_MAP_INFO_KHR,
+    .pNext = &mapPlacedInfo,
+    .memory = mem,
+    .offset = 0,
+    .size = VK_WHOLE_SIZE,
+    .flags = VK_MEMORY_MAP_PLACED_BIT_EXT,
+};
+void *map;
+VK_CHECK(vkMapMemory2KHR(device, &mapInfo, &map));
+CHECK(map == reserved);
+----
+
+## Issues
+
+1) How should the requested address be passed to `vkMapMemory2KHR()`?
+
+*RESOLVED*: When `VK_MEMORY_MAP_PLACED_BIT_EXT` is passed to
+`vkMapMemory2KHR()` and a `VkMemoryMapPlacedInfoEXT` structure is present
+in the `pNext` chain, the implementation attempts to map to whatever
+address is provided by `VkMemoryMapPlacedInfoEXT::pPlacedAddress`.
+
+2) What should happen if the implementation cannot place the memory map at
+the requested address?
+
+*RESOLVED*: The memory object should be left unmapped and
+flink:vkMapMemory2KHR should return ename:VK_ERROR_MEMORY_MAP_FAILED.
+
+3) How can a client atomically re-reserve the address range on unmap?
+
+*RESOLVED*: When `VK_MEMORY_UNMAP_RESERVE_BIT_EXT` is passed to
+`vkUnmapMemory2KHR()`, the implementation unmaps the memory range in such a
+way that the range is automatically re-reserved.
+With `mmap()`, this is accomplished by simply mapping over the range with
+another anonymous mapping.
+However, allowing this with `vkMapMemory()` would break Vulkan's concept of
+when a memory object is or is not mapped.
+
+4) Should a placed memory map replace existing maps in the specified
+range or fail if a map exists?
+
+*RESOLVED*: It should replace existing maps.
+If a client wants try-map behavior, it can get that by using mmap with
+`MAP_ANONYMOUS` and an address and only call `vkMapMemory2KHR()` to do a
+placed map if that succeeds.
+Without replacement, there is no way for the application to atomically
+exchange maps and prevent races between reservation and the final map.
+Also, any attempt at a middle ground in which it might replace would make
+the spec less well-defined.
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -338,8 +338,8 @@ typedef void* <name>MTLSharedEvent_id</name>;
         <type requires="VkCommandBufferResetFlagBits"     category="bitmask">typedef <type>VkFlags</type> <name>VkCommandBufferResetFlags</name>;</type>
         <type requires="VkCommandBufferUsageFlagBits"     category="bitmask">typedef <type>VkFlags</type> <name>VkCommandBufferUsageFlags</name>;</type>
         <type requires="VkQueryPipelineStatisticFlagBits" category="bitmask">typedef <type>VkFlags</type> <name>VkQueryPipelineStatisticFlags</name>;</type>
-        <type                                             category="bitmask">typedef <type>VkFlags</type> <name>VkMemoryMapFlags</name>;</type>
-        <type                                             category="bitmask">typedef <type>VkFlags</type> <name>VkMemoryUnmapFlagsKHR</name>;</type>
+        <type requires="VkMemoryMapFlagBits"              category="bitmask">typedef <type>VkFlags</type> <name>VkMemoryMapFlags</name>;</type>
+        <type requires="VkMemoryUnmapFlagBitsKHR"         category="bitmask">typedef <type>VkFlags</type> <name>VkMemoryUnmapFlagsKHR</name>;</type>
         <type requires="VkImageAspectFlagBits"            category="bitmask">typedef <type>VkFlags</type> <name>VkImageAspectFlags</name>;</type>
         <type requires="VkSparseMemoryBindFlagBits"       category="bitmask">typedef <type>VkFlags</type> <name>VkSparseMemoryBindFlags</name>;</type>
         <type requires="VkSparseImageFormatFlagBits"      category="bitmask">typedef <type>VkFlags</type> <name>VkSparseImageFormatFlags</name>;</type>
@@ -619,6 +619,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
         <type name="VkFormat" category="enum"/>
         <type name="VkFormatFeatureFlagBits" category="enum"/>
         <type name="VkFrontFace" category="enum"/>
+        <type name="VkMemoryMapFlagBits" category="enum"/>
         <type name="VkImageAspectFlagBits" category="enum"/>
         <type name="VkImageCreateFlagBits" category="enum"/>
         <type name="VkImageLayout" category="enum"/>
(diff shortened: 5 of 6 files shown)


Tobski:
@jekstrand can you add a proposal document for this extension?

oddhack:
Will be handled internally by the Development call - assigning to Tom as a placeholder for that.

jekstrand:
I've updated the PR as well as all the driver/test branches with the following changes:
 - It's now all based on a new VK_EXT_memory_map2
 - VK_EXT_map_memory_placed now has a proposal doc as per @Tobski
 - There is now a `VK_MEMORY_UNMAP_RESERVE_BIT_EXT` which cuases the driver to `mmap(MAP_ANONYMOUS | MAP_PRIVATE)` over the memory region instead of `munmap()`.  This should give the re-reserve behavior the Wine folks want.
 - My `UNMAP_RESERVE` test found some bugs in the ANV implementation which are now fixed.

jekstrand:
cc @Guy1524 @DadSchoorse

proposals/VK_EXT_map_memory_placed.asciidoc
```diff
@@ -0,0 +1,145 @@
...
+
+*PROPOSED*: When `VK_MEMORY_MAP_PLACED_BIT_EXT` is passed to
+`vkMapMemory()`, the `ppData` parameter becomes an input and the
+implementation attempts to map to whatever address is pointed to by
+`ppData`.
+
+2) How can a client atomically re-reserve the address range on unmap?
+
+*PROPOSED*: When `VK_MEMORY_MAP_RESERVE_ON_UNMAP_BIT_EXT` is passed to
+`vkMapMemory()`, the implementation remembers this bit and, when
```
Guy1524, line 132: On line 59 a VK_MEMORY_UNMAP_RESERVE_BIT_EXT flag is mentioned for vkUnmapMemory2EXT, which contradicts this statement I think.

jekstrand: Fixed.  I found a couple other simliar issues which I've now fixed up as well.

chapters/memory.txt
```diff
@@ -3096,18 +3096,168 @@ ifdef::VK_KHR_device_group[]
...
+  * pname:memory must: not have been allocated with multiple instances
+endif::VK_KHR_device_group[]
+ifdef::VK_EXT_map_memory_placed[]
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags, the
+    <<features-memoryMapPlaced, pname:memoryMapPlaced>> feature must: be
+    enabled
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags, the
+    pname:pNext chain must: include a slink:VkMemoryMapPlacedInfoEXT
+    structure
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
```
Guy1524, line 3190: What's the point of gating this functionality?  What would prevent any implementation style from subtracting the offset from the desired pointer, and mapping the object starting from the beginning?

DadSchoorse: Then it would use more address space than the application expects.

jekstrand: Yeah, if you're going to be mapping over an explicit range, the map has to match exactly.  Otherwise you'll stomp things.  And, sadly, most Linux kernel drivers don't support an `mmap()` offset today.  I'd like to see that fixed but, even if I typed the kernel patches today, it would be a year or two before they hit users so we need the feature bit.

Guy1524: Okay, and another question, should it be documented how much address space preceding the mapping is used up by vkMapMemory, so we know where and how big to make the reserved mapping?  Would the mapping always just start in the page of the address passed in? If not, there could be a problem w/ MAP_FIXED overwriting preceding pages unexpectedly.

jekstrand: The current plan, if we can get away with it, is that it will map exactly the requested range, like `mmap()` does.  The usual mapping system in DRM allows you to map as little or as much as you want at the beginning of the BO, you just can't offset into it.  As long as any extra space the driver needs is placed at the end, it shouldn't be a problem.

If we need to add something like that, we can, but I'd like to avoid it if we can.

Guy1524: Ah I see, and I just noticed the offset alignment requirement a few lines down, perfect.

xml/vk.xml
```diff
@@ -17243,16 +17293,32 @@ typedef void* <name>MTLSharedEvent_id</name>;
...
+                <enum value="0"                               name="VK_EXT_MAP_MEMORY_2_SPEC_VERSION"/>
+                <enum value="&quot;VK_EXT_map_memory2&quot;"  name="VK_EXT_MAP_MEMORY_2_EXTENSION_NAME"/>
+                <enum offset="0" extends="VkStructureType"    name="VK_STRUCTURE_TYPE_MEMORY_MAP_INFO_EXT"/>
+                <enum offset="1" extends="VkStructureType"    name="VK_STRUCTURE_TYPE_MEMORY_UNMAP_INFO_EXT"/>
+                <type name="VkMemoryMapInfoEXT"/>
+                <type name="VkMemoryUnmapInfoEXT"/>
+                <type name="VkMemoryUnmapFlagsEXT"/>
+                <type name="VkMemoryUnmapFlagBitsEXT"/>
+                <command name="vkMapMemory2EXT"/>
+                <command name="vkUnmapMemory2EXT"/>
```
Tobski, line 17307: Do we need two functions here? If we had a single "map" function then we could make it handle map/remap/unmap rather than having map and unmap and then wondering where to put remap.

Unmapping would then just be passing in an existing mapped address with VK_NULL_HANDLE for the memory object. Remapping would be passing in an existing mapped (or reserved?) address with a different memory object/offset.

I'm not married to this idea, but I do think it would be cleaner.

Tobski (changes requested):
Ok I feel like you'll anticipate what I'm going to say here but, can you write a proposal doc for map memory 2 as well please 😅

It seems like a straightforward extension, but it's not clear why we need it without the motivation written down.

I suppose you could also combine the two extensions with a single proposal document if you'd prefer, but that might be more work.

xml/vk.xml
jekstrand: Short version: I'm trying to avoid feature creep.  The point of this extension is to make `vkMap/UnmapMemory()` extensible so that I can extend them in the next extension.  It's not to re-think all of memory mapping.

Longer: have you seen the WDDM2 intrerfaces?  They both take a handle, don't support subset maps, and don't have a remap:

https://github.com/microsoft/libdxg/blob/main/include/d3dkmthk.h#L4809

As much as I'd love it if Vulkan looked more like `mmap()` but what you're suggesting would require a drvier-side map lookup table to get from map pointers to object handles if we wanted to implement it on Windows.  Also, we'd need VUs saying you could only unmap whole memory object ranges anway so it's not like it would actually help anyone.  Remap would either be a no-op or impossible depending on how we spec'd it.

As for remap, I'm not sure exactly what semantics we'd want there.  Just grow the mapping?  On Linux, I guess we could maybe provide a "real" remap that lets you grow maps but is there actually a call for this?

proposals/VK_EXT_map_memory_placed.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+`VK_MEMORY_MAP_32BIT_ADDRESS_BIT_EXT` which does exactly what emulation
+wants and places the map in the lower 32 bits.  This is equivalent to the
+Linux `MAP_32BIT` flag.  However, this is less flexible and may be
+difficult to implement on Arm platforms where Linux `mmap()` does not
+support `MAP_32BIT` since drivers are likely relying on `mmap()` to
+implement `vkMapMemory()`.
+
+## Proposal
+
+### API Features
```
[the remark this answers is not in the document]
jekstrand: I added something.  It depends on it to be able to extend `vkMapMemory()` and `vkUnmapMemory()`.  Currently, extending them is really awkward.

proposals/VK_EXT_map_memory_placed.asciidoc
```diff
@@ -0,0 +1,143 @@
...
+to place the map at the address specified by
+`VkMapMemoryPlacedInfoEXT::pPlacedAddress`.
+If the map cannot be placed exactly at the specified virtual address,
+`vkMapMemory2EXT()` returns `VK_ERROR_MEMORY_MAP_FAILED`.
+
+When `VK_MEMORY_UNMAP_RESERVE_BIT_EXT` is set in the `flags` member of
+`VkMemoryUnmapInfoEXT`, the address range currently occupied by the memory
+map will be atomically re-reserved by `vkUnmapMemory()` so that another
+call to `mmap()` or `vkMapMemory()` will not re-use that address range
+without an address hint or `VK_MEMORY_MAP_PLACED_BIT_EXT`, respectively.
```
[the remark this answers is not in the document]
jekstrand: Yeah, spec'ing this is tricky.  "remains reserved" is ok, but the fact that it's atomic is necessary.  If all we cared about was the ability to re-reserve, the client could do the reservation with `mmap()` after it returns.  We need to ensure that nothing can get in between to solve the relevant race or there's no point in having this part of the Vulkan extension.  See https://github.com/KhronosGroup/Vulkan-Docs/issues/1832#issuecomment-1199820012 for more details on why this feature is needed.

xml/vk.xml
Tobski: > As much as I'd love it if Vulkan looked more like mmap() but what you're suggesting would require a drvier-side map lookup table to get from map pointers to object handles if we wanted to implement it on Windows.

Ugh, Windows. Fair point. We could get around that but that's... definitely feature creepy.

> As for remap, I'm not sure exactly what semantics we'd want there. Just grow the mapping? On Linux, I guess we could maybe provide a "real" remap that lets you grow maps but is there actually a call for this?

Maybe grow, maybe set it to a completely separate memory binding, but then you run into the same issue with Windows again so blah. Windowwws...

proposals/VK_EXT_map_memory_placed.asciidoc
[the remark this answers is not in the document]
Tobski: Yea but you only need to talk about "atomic re-reserve" if you imply that unmap unreserves the range instead of just unmapping it. So you can shortcut that by saying it's never unreserved in the first place.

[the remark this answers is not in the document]
jekstrand: Possibly?  I'm already waving my hands a lot by talking about "reserved" VA ranges when we have no such concept in the spec and it's really an OS concept that's entirely out of the Vulkan's jurisdiction.  Let me think on it.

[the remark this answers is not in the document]
jekstrand: How about ", the memory object will be unmapped but the address range will remain reserved so that another..."?  I still feel like we need to say somewhere explicitly that it's all atomic and another `mmap()` in another thread can't race with it and steal your VA range.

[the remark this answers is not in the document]
Tobski: Yea that works.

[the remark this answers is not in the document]
jekstrand: Done.

appendices/VK_EXT_map_memory_placed.txt
```diff
@@ -0,0 +1,34 @@
...
+*Last Modified Date*::
+    2022-07-28
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Jason Ekstrand, Collabora
+
+=== Description
+
+This extension allows a client to request that flink:vkMapMemory attempt to
```
Tobski, line 18:
```suggestion
This extension allows a client to request that flink:vkMapMemory2EXT attempt to
```
?

chapters/features.txt
```diff
@@ -5451,6 +5451,33 @@ include::{generated}/validity/structs/VkPhysicalDeviceImageProcessingFeaturesQCO
...
+This structure describes the following features:
+
+  * [[features-memoryMapPlaced]] pname:memoryMapPlaced indicates that
+    the implementation supports placing memory maps at client-specified
+    virtual addresses.
+  * [[features-memoryMapRangePlaced]] pname:memoryMapRangePlaced indicates
+    that the implementation supports placing memory maps of a subrange of a
+    memory object at client-specified virtual addresses.
+  * [[features-memoryUnmapReserve]] pname:memoryUnmapReserve
+    indicates that the implementation supports atomically re-reserving
```
Tobski, line 5471: This text I think should similarly be updated to the "don't unreserve it" language.

chapters/limits.txt
```diff
@@ -3370,6 +3370,28 @@ include::{generated}/validity/structs/VkPhysicalDevicePipelineRobustnessProperti
...
+defined as:
+
+include::{generated}/api/structs/VkPhysicalDeviceMapMemoryPlacedPropertiesEXT.txt[]
+
+The members of the sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT
+structure describe the following:
+
+  * [[limits-minPlacedMemoryMapAlignment]]
+    pname:minPlacedMemoryMapAlignment is the minimum alignment required by
+    placed memory maps.
```
Tobski, line 3386:
```suggestion
    memory ranges when using placed memory mapping.
```

maybe?

It wasn't immediately obvious to me if this was a requirement on the host address or the memory range - the VU clears it up but it'd be good to not have to go digging for it.

chapters/memory.txt
```diff
@@ -3096,18 +3096,168 @@ ifdef::VK_KHR_device_group[]
...
 ****
 
 include::{generated}/validity/protos/vkMapMemory.txt[]
 --
 
-[open,refpage='VkMemoryMapFlags',desc='Reserved for future use',type='flags']
+ifdef::VK_EXT_map_memory_placed[]
+[open,refpage='VkMemoryMapFlagBits',desc='Bitmask specifying additional parameters of a memory map',type='enums']
+--
+Bits which can: be set in flink:vkMapMemory::pname:flags, specifying
```
Tobski, line 3110: Feels a bit weird to call out vkMapMemory when the only flag explicitly _cannot_ be used with it, maybe:

```suggestion
Bits which can: be set in flink:vkMapMemory::pname:flags and flink:vkMapMemory2EXT::pname:flags, specifying
```

chapters/features.txt
Tobski: Also, what's the point in these latter two being features - do these need to be explicitly enabled? Feels like they might make more sense as properties?

Tobski (changes requested):

appendices/VK_EXT_map_memory_placed.txt
jekstrand: Yup.  Done.

chapters/memory.txt
jekstrand: Done.

chapters/limits.txt
jekstrand: Yeah. Good call.  I think I've fixed it.

chapters/features.txt
jekstrand: Do they need to be explicitly enabled?  Probably not.  Are they really properties, though?  I don't think they are.  They both gate functionality.  One says whether or not you can map subranges and the other says whether or not `UNMAP_RESERVE` is supported.  Neither of these feels like a property to me.

jekstrand:
While dealing with the comments from Tobias, I realized I didn't specify interactions with host pointer import.  I've added a couple VUs saying that you can't do a placed maps or unmap with reserve on something imported from a host pointer import.

chapters/features.txt
Tobski: I mean you're not wrong, but at the same time... eh. I usually plop in a single feature for things like this and a bunch of properties, saves developers having to enable multiple things. It's probably not the biggest deal either way though 🤷‍♀️ - if you want to call this resolved with no change that is a'ok with me.

chapters/limits.txt
Tobski: Ok hang on, so is it meant to be _both_? If so, then I think you need a VU to state that the address needs to be aligned too (at least that's how I'm reading your new text?)

jekstrand: Yes, it's on both.  It should just be page size but I figured a limit is better than trying to refer to the system page size and hoping that's always the exact requirement.  There's already a VU on `VkMemoryMapPlacedInfoEXT`.

chapters/memory.txt
oddhack: N.b. it's helpful to have back-references to all the users of a type, but it's fragile against extensions, and can get wordy. In the refpages it automatically generates "See Also" references for all APIs which use the API being described but we don't do that in the spec itself (though we could).

chapters/limits.txt
Tobski: Ah I missed that it was in a different place. Thanks!

appendices/VK_EXT_map_memory_placed.txt
```diff
@@ -0,0 +1,39 @@
...
+attempt to place the memory map at a particular virtual address.
+
+include::{generated}/interfaces/VK_EXT_map_memory_placed.txt[]
+
+=== Issues
+
+1) What should happen if the implementation cannot place the memory map at
+the requested address?
+
+*PROPOSED*: The memory object should be left unmapped and flink:vkMapMemory
```
Tobski, line 33:
```suggestion
*PROPOSED*: The memory object should be left unmapped and flink:vkMapMemory2EXT
```

chapters/memory.txt
```diff
@@ -3096,18 +3096,175 @@ ifdef::VK_KHR_device_group[]
...
+--
+Bits which can: be set in flink:vkMapMemory::pname:flags and
+slink:VkMemoryMapInfoEXT::pname:flags, specifying additional properties of
+a memory map, are:
+
+include::{generated}/api/enums/VkMemoryMapFlagBits.txt[]
+
+  * ename:VK_MEMORY_MAP_PLACED_BIT_EXT requests that the implementation
+    attempt to place the memory map at the virtual address specified by the
+    client via pname:ppData.
```
Tobski, line 3118:
```suggestion
    client via pname:ppData.
    Can: only be used in slink:VkMemoryMapInfoEXT::pname:flags.
```

Worth calling this out?

chapters/memory.txt
```diff
@@ -3096,18 +3096,175 @@ ifdef::VK_KHR_device_group[]
...
+sname:VkMemoryMapPlacedInfoEXT::pname:pPlacedAddress and
+fname:vkMapMemory2EXT will return ename:VK_SUCCESS.
+If it cannot place the map at the requested address for any
+reason, the memory object is left unmapped and fname:vkMapMemory2EXT will
+return ename:VK_ERROR_MEMORY_MAP_FAILED.
+
+[NOTE]
+.Note
+====
+In order for a placed map to succeed, the client may: first have to reserve
```
Tobski, line 3239:
```suggestion
In order for a placed map to succeed, the client may first have to reserve
```

We don't use normative tags in notes.

Tobski: Although thinking about it - should this be a note? Also is this something we should advertise via a limit if it's a static property, or is this more "if you haven't reserved it, then it might have already been snatched up by another allocation"?

chapters/memory.txt
```diff
@@ -3294,6 +3458,86 @@ include::{generated}/api/protos/vkUnmapMemory.txt[]
...
+ifdef::VK_EXT_map_memory_placed[]
+[open,refpage='VkMemoryUnmapFlagBitsEXT',desc='Bitmask specifying additional parameters of a memory unmap',type='enums']
+--
+Bits which can: be set in slink:VkMemoryUnmapInfoEXT::pname:flags, specifying
+additional properties of a memory unmap, are:
+
+include::{generated}/api/enums/VkMemoryUnmapFlagBitsEXT.txt[]
+
+  * ename:VK_MEMORY_UNMAP_RESERVE_BIT_EXT requests that virtual address
+    range currently occupied by the memory map be remain reserved after the
```
Tobski, line 3519:
```suggestion
    range currently occupied by the memory map remains reserved after the
```

chapters/memory.txt
```diff
@@ -3294,6 +3458,86 @@ include::{generated}/api/protos/vkUnmapMemory.txt[]
...
+    flink:vkUnmapMemory2EXT call completes.
+    Future system memory map operations or calls to flink:vkMapMemory or
+    flink:vkMapMemory2EXT will not return addresses in that range unless
+    the range has since been unreserved by the client or the memory map
```
Tobski, line 3523:
```suggestion
    the range has since been unreserved by the client, the memory map
```

chapters/memory.txt
```diff
@@ -3294,6 +3458,86 @@ include::{generated}/api/protos/vkUnmapMemory.txt[]
...
+    operation is done with ename:VK_MEMORY_MAP_PLACED_BIT_EXT or the system
```
Tobski, line 3524:
```suggestion
    operation is done with ename:VK_MEMORY_MAP_PLACED_BIT_EXT, or the system
```

chapters/memory.txt
```diff
@@ -3294,6 +3458,86 @@ include::{generated}/api/protos/vkUnmapMemory.txt[]
...
+    memory map equivalent.
+    When ename:VK_MEMORY_UNMAP_RESERVE_BIT_EXT is set, the memory unmap
+    operation may fail, in which case the memory object will remain host
```
Tobski, line 3527:
```suggestion
    operation may: fail, in which case the memory object will remain host
```

chapters/memory.txt
```diff
@@ -3294,6 +3458,86 @@ include::{generated}/api/protos/vkUnmapMemory.txt[]
...
+    mapped and flink:vkUnmapMemory2EXT will return
+    ename:VK_ERROR_MEMORY_MAP_FAILED.
```
Tobski, line 3529: Is it possible to quantify (perhaps in a note) the circumstances under which this happens?

proposals/VK_EXT_map_memory_placed.asciidoc
```diff
@@ -0,0 +1,147 @@
...
+    .flags = VK_MEMORY_MAP_PLACED_BIT_EXT,
+};
+void *map;
+VK_CHECK(vkMapMemory2EXT(device, &mapInfo, &map));
+CHECK(map == reserved);
+----
+
+## Issues
+
+1) How should the requested address be passed to `vkMapMemory()`?
```
Tobski, line 126:
```suggestion
1) How should the requested address be passed to `vkMapMemory2EXT()`?
```

proposals/VK_EXT_map_memory_placed.asciidoc
```diff
@@ -0,0 +1,147 @@
...
+[source,c]
+----
+typedef enum VkMemoryMapFlagBits {
+    VK_MEMORY_MAP_PLACED_BIT_EXT = 0x00000001,
+} VkMemoryMapFlagBits;
+
+typedef enum VkMemoryUnmapFlagBits {
+    VK_MEMORY_UNMAP_RESERVE_BIT_EXT = 0x00000001,
+} VkMemoryMapFlagBits;
+----
```
Tobski, line 66: Is this supposed to be the feature struct here? The placement of these bit flags doesn't make sense to me. You should also illustrate the VkMemoryMapPlacedInfoEXT struct here somewhere in the proposal section.

chapters/memory.txt
```diff
@@ -3096,18 +3096,175 @@ ifdef::VK_KHR_device_group[]
...
+
+.Valid Usage
+****
+  * If sname:VkMemoryMapInfoEXT::pname:flags contains
+    ename:VK_MEMORY_MAP_PLACED_BIT_EXT, pname:pPlacedAddress must: not be
+    `NULL`
+  * If sname:VkMemoryMapInfoEXT::pname:flags contains
+    ename:VK_MEMORY_MAP_PLACED_BIT_EXT, pname:pPlacedAddress must: be
+    aligned to an integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
```
Tobski, line 3262: Both of these VUs should move to the parent structure (VkMemoryMapInfoEXT) as they reference a value in the parent structure.

chapters/memory.txt
```diff
@@ -3294,6 +3458,86 @@ include::{generated}/api/protos/vkUnmapMemory.txt[]
...
+include::{generated}/api/structs/VkMemoryUnmapInfoEXT.txt[]
+
+  * pname:sType is the type of this structure.
+  * pname:pNext is `NULL` or a pointer to a structure extending this
+    structure.
+  * pname:memory is the slink:VkDeviceMemory object to be unmapped.
+
+.Valid Usage
+****
+  * pname:memory must: be currently host mapped
```
Tobski, line 3493: Just to check - this is expected to allow unmapping of memory mapped with VkMapMemory2EXT (and vice-versa), right?

chapters/memory.txt
```diff
@@ -3294,6 +3458,86 @@ include::{generated}/api/protos/vkUnmapMemory.txt[]
...
+[open,refpage='VkMemoryUnmapInfoEXT',desc='Structure containing parameters of a memory unmap operation',type='structs']
+--
+The sname:VkMemoryUnmapInfoEXT structure is defined as:
+
+include::{generated}/api/structs/VkMemoryUnmapInfoEXT.txt[]
+
+  * pname:sType is the type of this structure.
+  * pname:pNext is `NULL` or a pointer to a structure extending this
+    structure.
+  * pname:memory is the slink:VkDeviceMemory object to be unmapped.
```
Tobski, line 3489: Missing the `flags` parameter?

proposals/VK_EXT_map_memory2.asciidoc
```diff
@@ -0,0 +1,75 @@
...
+    VkDevice                                    device,
+    const VkMemoryMapInfoEXT*                   pMemoryMapInfo,
+    void**                                      ppData);
+
+typedef struct VkMemoryUnmapInfoEXT {
+    VkStructureType          sType;
+    const void*              pNext;
+    VkDeviceMemory           memory;
+    VkMemoryUnmapFlagsEXT    flags;
+} VkMemoryUnmapInfoEXT;
```
Tobski, line 45: Interaction question here that probably needs spelling out (at least in the map_placed proposal) - what happens if you map separate multiple ranges via memoryMapRangePlaced? Language you've got so far says that it unmaps _the_ range, which could plausibly refer to the range of `memory`, which is fine, but firstly I want to double check that, and secondly - should there be a way to unmap a subrange in the same way?

Tobski (changes requested):
Ok this is my most thorough review yet, so I'm fairly sure once all of this is addressed you can finally have my approval :P

chapters/memory.txt
jekstrand: I think the original is correct.  It's a bit annoying the way it's phrased but I don't think the comma works.

chapters/memory.txt
jekstrand: Same.  The last or is sort of a parenthetical.  It's supposed to read as unreserved || (VK_MAP_PLACED || SYS_MAP_FIXED)

chapters/memory.txt
jekstrand: Yes.  There's no reason why you can't mix+match.

chapters/memory.txt
jekstrand: Yup.  Fixed.

chapters/memory.txt
oddhack: I agree with @tobias - the comma disambiguates the grouping of the multiple 'or' clauses. It's somewhat like an 'eats shoots and leaves' exercise.

For complete disambiguation, though, a bullet list is probably better in this case.

chapters/memory.txt
jekstrand: Without digging into kernel source code, I'm not sure when it would fail.  I'm not even sure it can.  But `mmap()` is allowed to return `MAP_FAILED` for pretty arbitrary reasons and I would rather throw an error than silently fall back to `munmap()` and lose the reservation.

jekstrand: Context: https://gitlab.freedesktop.org/jekstrand/mesa/-/commit/348a18c3a559e280d8ef859fc6079ec44c1bde2e#637da2a63a937bb34d4a7112d341f9ea5c652765_168_177

chapters/memory.txt
jekstrand: We can, I guess.  There's a VU for that but we can call it out explicitly, I suppose.  IDK.

proposals/VK_EXT_map_memory2.asciidoc
jekstrand: You're not allowed to map two different ranges.  This VU still holds:
> `VUID-vkMapMemory-memory-00678`
>
> `memory` **must** not be currently host mapped

It unmaps the mapped range.  There's only one mapped range.

If we wanted to allow more direct control of memory mappings, that feels like a different feature.  I don't see why it wouldn't be supportable on Linux but I also don't see any real value.  Not supportable on Windows, of course.

chapters/memory.txt
jekstrand: How about I drop the mention of flags and just say it has to be aligned.  0 is trivially aligned so it should be fine in 99% of the cases where the struct is unused.

chapters/memory.txt
jekstrand: Ok, I think I've improved it.  It now says:

> Future system memory map operations or calls to flink:vkMapMemory or
> `vkMapMemory2EXT` will not return addresses in that range unless
> the range has since been unreserved by the client or the mapping is
> explicitly placed in that range by calling `vkMapMemory2EXT` with
> `VK_MEMORY_MAP_PLACED_BIT_EXT` or doing the system memory map
> equivalent.

chapters/memory.txt
jekstrand: I dropped the mention of flags in the alignment one.  The other is annoying.  It really is a VU on `pPlacedAddress` not the flags.  One option would be to toss the flag and just make it based on `VkMemoryMapPlacedInfoEXT` being present in the `pNext` chain and `pPlacedAddress != NULL`.  I kind-of do like the flag, though.

proposals/VK_EXT_map_memory_placed.asciidoc
jekstrand: Added all the structs.

chapters/memory.txt
jekstrand: Hrm... Looking at the `mmap()` docs, I suspect we may may actually want two flags:

 - `VK_MEMORY_MAP_TRY_PLACED`: Attempts to place the memory map and fails if another mapping currently exists in that range.
 - `VK_MEMORY_MAP_PLACED_REPLACE`: Replaces whatever is at the current location regardless.  This needs a VU that it's not overwritting another Vulkan mapping because we want the "one map at a time" rule to be validatable.

These map to `MAP_FIXED_NOREPLACE` and `MAP_FIXED`, respectively.  Then it would be explicit what happens and I think we could drop this note entirely.  We could also make them different feature bits.

chapters/memory.txt
jekstrand: If we go with the explicit two bits I mentioned above, we could replace all this "reserve" nonsense and say that it replaces the Vulkan memory map with an inaccessible (`PROT_NONE`) anonymous map.

chapters/memory.txt
DadSchoorse: It's simple for applications to emulate `VK_MEMORY_MAP_TRY_PLACED` with a `mmap` call with `MAP_FIXED_NOREPLACE` + `VK_MEMORY_MAP_PLACED_REPLACE`. So I see no reason to have `VK_MEMORY_MAP_TRY_PLACED` at all.

chapters/memory.txt
Tobski: That's better, but I think the (or doing the system memory map equivalent) is something that should be in brackets. The lack of punctuation melts my brain 🙃

chapters/memory.txt
Tobski: I think a naive reader might get confused but probably not for long so eh, take it or leave it.

chapters/memory.txt
Tobski: Ok, if this is basically esoteric OS behavior then there's not much we can state here.

chapters/memory.txt
Tobski: I'm fine with either resolution, though dropping the flag would make the API simpler, and make it more obvious that you can't do it with the vanilla map memory command.

chapters/memory.txt
Tobski: Ok, don't think this needs calling out, just wanted to make sure it was expected.

Tobski: (though this might need testing in the CTS test plan...)

proposals/VK_EXT_map_memory2.asciidoc
Tobski: Ah ok, didn't realise you could only have a single mapped range full stop. I suspect it might be useful with SVM-like features in the future, but on its own right now agree it's pretty useless. Fine with this resolution then.

chapters/memory.txt
Tobski: Resolving this thread but only on the basis that the conversation is continuing in the thread below 🙃

billhollings:
Apologies for the late arrival of commentary here.

I'm trying to figure a way this could work in MoltenVK. 

In Metal, device memory is allocated from a `MTLBuffer` whose address range is set when it is created to underly the `VkDeviceMemory`. So, we could possibly take the `pPlacedAddress` at `VkDeviceMemory` creation time, but at the time we're dealing with `vkMapMemory2EXT`, the memory has been allocated to an address already. 

At that point, because we're dealing with a memory address and not a file descriptor, I don't see how we could use something simple like `mmap()`. We'd probably have to resort to `memcpy()'s` under the covers, which could work (albeit inefficiently) for host-visible memory, but not for host-coherent device memory that is not flushed or unmapped by the app.

cjacek:
> At that point, because we're dealing with a memory address and not a file descriptor, I don't see how we could use something simple like `mmap()`.

macOS has mach_vm_remap() that could probably be used. We have successfully used it for memory mapped by OpenGL and I hope it will work for Metal as well.

billhollings:
> macOS has `mach_vm_remap()` that could probably be used.

Thanks for the suggestion. I came across this separately as well, after I posted above. I'll definitely look into it.

zfigura:
On the other hand, if Metal allows manually placing the mapped memory location, but only at memory allocation time, would it make sense to rewrite the extension around that? It would certainly work just as well for Wine.

DadSchoorse:
> It would certainly work just as well for Wine.

Would it though? If we can only place it at a VA location at allocation time, we would always block that VA space even if the memory object isn't mapped. And that doesn't sound great for wine at all because 32bit memory is very limited.

chapters/memory.txt
jekstrand: That's not going to be atomic.  If something else successfully maps between the two calls, the `PLACED_REPLACE` will stomp it.  I don't know if that's a problem or not.

Guy1524: Whether a flag is added to Vulkan or not, a component using MMAP_FIXED/PLACE_REPLACE on an address it hasn't already reserved is a potential race condition.  Adding the MAP_TRY_PLACED would only plug one side of that race condition, while doing nothing against the Vulkan mapping being overwritten.

jekstrand: Hrm... Maybe I've not woken up yet.  Yes, I think what @DadSchoorse suggested does work.  If `MAP_FIXED_NOREPLACE` succeeds, you've "reserved" the range and can replace it without worrying about stomping.  Ok, I think I"m convinced that the try isn't needed.

zfigura:
> > It would certainly work just as well for Wine.
> 
> Would it though? If we can only place it at a VA location at allocation time, we would always block that VA space even if the memory object isn't mapped. And that doesn't sound great for wine at all because 32bit memory is very limited.

Indeed; it seems I hadn't thought that one through...

chapters/memory.txt
```diff
@@ -3163,12 +3183,89 @@ include::{generated}/api/structs/VkMemoryMapInfoEXT.txt[]
...
+    is not enabled, pname:size must: be ename:VK_WHOLE_SIZE
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is enabled, pname:offset must: be aligned to an integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is enabled, pname:size must: be ename:VK_WHOLE_SIZE or be aligned to an
+    integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
```
cjacek, line 3207: VkDeviceMemory size is not required to be aligned and if caller would like to pass aligned size that includes the end of the buffer, it would break 'If size is not equal to VK_WHOLE_SIZE, size must be less than or equal to the size of the memory minus offset' requirement. I guess it's possible for the caller to use VK_WHOLE_SIZE in this case, but maybe it would make sense to remove this restriction, allow unaligned size from the caller and specify that implementation does the alignment?

cjacek: Nevermind, I think I can see how it's better for drivers. It should be fine to handle that on Wine side.

cjacek:
We have support for wow64 thunks in Wine now and I experimented with winevulkan memory mapping on top of https://gitlab.freedesktop.org/jekstrand/mesa/-/commits/anv/VK_EXT_map_memory_placed. My prototype implementation is here: https://gitlab.winehq.org/jacek/wine/-/commits/wow. It looks promising overall, I was able to run a number of Vulkan tests. (A number of things still don't work, but I it's probably for unrelated reasons and needs more debugging. wow64 support in Wine is still in a rough state and the fact that I used A770 on top of jekstrand's branch adds more uncertainty around my setup.)

chapters/memory.txt
jekstrand: I've updated the spec to require replacement behavior.  This should get rid of any of that annoying "may have to reserve" nonsense.

chapters/memory.adoc
```diff
@@ -3108,18 +3114,171 @@ ifdef::VK_KHR_device_group[]
...
+If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in
+sname:VkMemoryMapInfoEXT::pname:flags and the pname:pNext chain of
+slink:VkMemoryMapInfoEXT includes a sname:VkMemoryMapPlacedInfoEXT
+structure, then that structure specifies the placement address of the
+memory map.
+The implementation will place the memory map at the specified address,
+replacing any existing maps in the specified memory range.
+Replacing memory maps in this way does not implicitly unmap Vulkan memory
+objects.
+Instead, the client must: ensure no other Vulkan memory objects are mapped
```
cubanismo, line 3250: Should this be VU?

chapters/memory.adoc
```diff
@@ -3108,18 +3114,171 @@ ifdef::VK_KHR_device_group[]
...
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is enabled, pname:size must: be ename:VK_WHOLE_SIZE or be aligned to an
+    integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
+ifdef::VK_EXT_external_memory_host[]
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags, the memory
+    object must: not have been imported from a handle type of
+    ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_ALLOCATION_BIT_EXT or
+    ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_MAPPED_FOREIGN_MEMORY_BIT_EXT
```
cubanismo, line 3229: Is this actually required? Doesn't it make implementing a wrapper thing that supports both these extensions harder? Would it be sufficient to instead say the address must match the host pointer imported (accounting for offset and whatnot) in these cases?

chapters/memory.adoc
```diff
@@ -3306,6 +3472,89 @@ include::{generated}/api/protos/vkUnmapMemory.adoc[]
...
+ifdef::VK_EXT_map_memory_placed[]
+[open,refpage='VkMemoryUnmapFlagBitsEXT',desc='Bitmask specifying additional parameters of a memory unmap',type='enums']
+--
+Bits which can: be set in slink:VkMemoryUnmapInfoEXT::pname:flags, specifying
+additional properties of a memory unmap, are:
+
+include::{generated}/api/enums/VkMemoryUnmapFlagBitsEXT.adoc[]
+
+  * ename:VK_MEMORY_UNMAP_RESERVE_BIT_EXT requests that virtual address
+    range currently occupied by the memory map remains reserved after the
```
cubanismo, line 3535: is it "remains" or "remain?"

chapters/limits.adoc
```diff
@@ -3953,6 +3953,29 @@ include::{generated}/validity/structs/VkPhysicalDeviceRayTracingInvocationReorde
...
+defined as:
+
+include::{generated}/api/structs/VkPhysicalDeviceMapMemoryPlacedPropertiesEXT.adoc[]
+
+The members of the sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT
+structure describe the following:
+
+  * [[limits-minPlacedMemoryMapAlignment]]
+    pname:minPlacedMemoryMapAlignment is the minimum alignment required for
+    memory object offsets and virtual address ranges when using placed
```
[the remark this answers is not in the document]
gfxstrand: I don't think so. The CPU memory map alignment should only depend on CPU page table limitations which are currently 4K on x86 and most Arm and 16K on Apple M1. If someone has a counter-example to this, I'm happy to hear it but I'd rather keep it at the physical device level if we can.

chapters/memory.adoc
gfxstrand: Sure.  I can do that. I've added a VU but I think leaving this sentence here makes the spec more clear.

chapters/memory.adoc
gfxstrand: fixed.

chapters/memory.adoc
gfxstrand: Maybe? The current pass-through case for host allocations is easy because you just return the pointer the client provided, possibly with an offset.  It's a bit weird but ultimately harmless.  In this case, we'd be asking the client to jump through a bunch of hoops to guarantee that they only map it at itself and then it's unclear what kind of map you have now. I don't see any value in allowing this and it raises more questions than it answers.

cjacek:
> I've updated the branch which also pulled in 5ish months of Intel Arc bug fixes. It very well may be that your issues are related to some of those.

Thanks! Together with newer kernel, that seems to work better. I've updated Wine branch on top of current Wine as well (meantime I upstreamed a hackish implementation based VK_EXT_external_memory_host, which included parts of my previous VK_EXT_map_memory_placed branch, so the diff is now smaller). If there is anything to do on our side that would help this move forward, please let me know.

xml/vk.xml
```diff
@@ -7675,6 +7677,23 @@ typedef void* <name>MTLSharedEvent_id</name>;
...
             <member optional="true"><type>VkMemoryUnmapFlagsEXT</type>  <name>flags</name></member>
         </type>
+        <type category="struct" name="VkPhysicalDeviceMapMemoryPlacedFeaturesEXT" structextends="VkPhysicalDeviceFeatures2,VkDeviceCreateInfo">
+            <member values="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MAP_MEMORY_PLACED_FEATURES_EXT"><type>VkStructureType</type> <name>sType</name></member>
+            <member optional="true"><type>void</type>*            <name>pNext</name></member>
+            <member><type>VkBool32</type>                         <name>memoryMapPlaced</name></member>
+            <member><type>VkBool32</type>                         <name>memoryMapRangePlaced</name></member>
+            <member><type>VkBool32</type>                         <name>memoryUnmapReserve</name></member>
+        </type>
+        <type category="struct" name="VkPhysicalDeviceMapMemoryPlacedPropertiesEXT" structextends="VkPhysicalDeviceProperties2">
```
cjacek, line 7687: This could use returnedonly attribute.

chapters/memory.adoc
```diff
@@ -3108,18 +3114,174 @@ ifdef::VK_KHR_device_group[]
...
+    is not enabled, pname:size must: be ename:VK_WHOLE_SIZE
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is enabled, pname:offset must: be aligned to an integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags and the
+    <<features-memoryMapRangePlaced, pname:memoryMapRangePlaced>> feature
+    is enabled, pname:size must: be ename:VK_WHOLE_SIZE or be aligned to an
+    integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
```
PanGao-h, line 3224: `VkPhysicalDeviceLimits::minMemoryMapAlignment` in normal `vkMapMemory` has some other constraint for ppData and offset. Is there any interaction between it and `VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::minPlacedMemoryMapAlignment`?

chapters/memory.adoc
```diff
@@ -3108,18 +3114,174 @@ ifdef::VK_KHR_device_group[]
...
+Replacing memory maps in this way does not implicitly unmap Vulkan memory
+objects.
+Instead, the client must: ensure no other Vulkan memory objects are mapped
+anywhere in the specified virtual address range.
+If successful, pname:ppData will be set to the same value as
+sname:VkMemoryMapPlacedInfoEXT::pname:pPlacedAddress and
+fname:vkMapMemory2EXT will return ename:VK_SUCCESS.
+If it cannot place the map at the requested address for any
+reason, the memory object is left unmapped and fname:vkMapMemory2EXT will
+return ename:VK_ERROR_MEMORY_MAP_FAILED.
```
PanGao-h, line 3257: If map failed, what's the returned value should be? NULL or just some random address?

chapters/memory.adoc
```diff
@@ -3108,18 +3114,174 @@ ifdef::VK_KHR_device_group[]
...
+.Valid Usage
+****
+  * If sname:VkMemoryMapInfoEXT::pname:flags contains
+    ename:VK_MEMORY_MAP_PLACED_BIT_EXT, pname:pPlacedAddress must: not be
+    `NULL`
+  * pname:pPlacedAddress must: be aligned to an integer multiple of
+    sname:VkPhysicalDeviceMapMemoryPlacedPropertiesEXT::pname:minPlacedMemoryMapAlignment
+  * The address range specified by pname:pPlacedAddress and
+    sname:VkMemoryMapInfoEXT::pname:size must: not overlap any existing
+    Vulkan memory object mapping.
```
PanGao-h, line 3279: Does `size` need to be larger than the mapped range? If the `size` is larger than actual memory size, what's the remaining data should be? Some random data or zeroed?

chapters/memory.adoc
Guy1524: How I read it:

If `minPlacedMemoryMapAlignment` is just as or less granular than `minMemoryMapAlignment`, there is no relevant interaction here, as `*ppData` = `pPlacedAddress`, and both `pPlacedAddress` and `offset` must be aligned to `minPlacedMemoryMapAlignment`.

But yeah, if `minPlacedMemoryMapAlignment` were ever more granular, capitalising on that granularity would be illegal, because `pPlacedAddress` - `offset` still needs to align to `minMemoryMapAlignment`.

I can't imagine a situation where the minimum placed alignment should be more granular than normal, so I'm not sure if this actually a problem.

chapters/memory.adoc
Guy1524: As I far as I know this isn't defined for the other vkMapMemory failures, why should this be defined behavior here?

chapters/memory.adoc
Guy1524: Why would size need to be larger, I think that's illegal:

`VUID-vkMapMemory-size-00681`
"If size is not equal to VK_WHOLE_SIZE, size must be less than or equal to the size of the memory minus offset"

appendices/VK_EXT_map_memory_placed.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+*RESOLVED*: When `VK_MEMORY_UNMAP_RESERVE_BIT_EXT` is passed to
+`vkUnmapMemory2EXT()`, the implementation unmaps the memory range in such a
+way that the range is automatically re-reserved.  With `mmap()`, this is
+accomplished by simply mapping over the range with another anonymous
+mapping.  However, allowing this with `vkMapMemory()` would break Vulkan's
+concept of when a memory object is or is not mapped.
+
+3) Should a placed memory map replace existing maps in the specified
+range or fail if a map exists?
```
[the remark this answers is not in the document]
gfxstrand: :woman_shrugging: IDK. I put most of the issues in both but I can drop them from the proposal.

xml/vk.xml
gfxstrand: Sure.

chapters/memory.adoc
gfxstrand: Correct. If you get `VK_ERROR_MEMORY_MAP_FAILED`, you should ignore the pointer.

gfxstrand:
Now that VK_KHR_map_memory2 is ratified and released, I've rebased this branch and fixed up a bunch of stuff.  The ANV branch has also been updated.

cubanismo (approved):
Still says WIP, but this LGTM.

gfxstrand:
> Still says WIP, but this LGTM.

Dropped the WIP. We're currently waiting on someone to write some CTS tests. I put together a test plan but have yet to find someone to write the actual tests.

pdaniell-nv:
There is an initial implementation of this extension in our NVIDIA Vulkan Beta drivers here: https://developer.nvidia.com/vulkan-driver

gfxstrand:
@oddhack, I just rebased the branch today and we agreed on the working group call to skip some of the usual process and go ahead and ship this.  Do you want to merge it from here or would you like me to create an internal MR and merge it that way?

Also, this is pretty short notice so it's okay if we miss this week's spec update.

chapters/memory.adoc
gfxstrand: `minMemoryMapAlignment` can be as low as 1 because the driver can fake offsets internally. In fact, Windows drivers have to because the only Windows interface for mapping GPU memory maps the whole memory object every time.  With placed maps, however, you are actually restricted to the OS page size and there's no getting around that.

gfxstrand:
@oddhack To be clear, this is good to ship as soon as you're ready. There's no need to delay until next week. I just didn't want to throw a wrench in your process if you already had other extensions queued up to go in.

TheComputerGuy96:
> And... we now have an NVK implementation: https://gitlab.freedesktop.org/mesa/crucible/-/merge_requests/147

@gfxstrand This is the actual NVK implementation: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/27599 (the link you posted is just some non-Vulkan CTS tests for the extension)

oddhack:
> @oddhack To be clear, this is good to ship as soon as you're ready. There's no need to delay until next week. I just didn't want to throw a wrench in your process if you already had other extensions queued up to go in.

It's fine to merge from here, though I'm glad you're bringing it up early as it's easier to handle the github PRs first and then the internal ones. Should respond to @tobski's suggestion but I'll sign this off and it will go out this week.

chapters/memory.adoc
```diff
@@ -3649,12 +3676,89 @@ ifdef::VK_KHR_device_group[]
   * [[VUID-VkMemoryMapInfoKHR-memory-07963]]
     pname:memory must: not have been allocated with multiple instances
 endif::VK_KHR_device_group[]
+ifdef::VK_EXT_map_memory_placed[]
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags, the
+    <<features-memoryMapPlaced, pname:memoryMapPlaced>> feature must: be
+    enabled
+  * If ename:VK_MEMORY_MAP_PLACED_BIT_EXT is set in pname:flags, the
+    pname:pNext chain must: include a slink:VkMemoryMapPlacedInfoEXT
```
spencer-lunarg, line 3684: (sorry late to the party) 

nit: probably should say 

> pname:pNext chain must: include a slink:VkMemoryMapPlacedInfoEXT structure with a non null pPlacedAddress

gfxstrand:
I've dealt with the comments from @Tobski and @spencer-lunarg. I also fixed a dependency warning.

gfxstrand:
Unfortunately, my fix for the dependency warning caused a new CI issue. It's complaining that I put the requires for `VkMemoryUnmapFlagBitsKHR` in VK_EXT_memory_map_placed which uses an EXT suffix. However, this is consistent with other cases where we add an unused flags field in an extension or core version and then a different extension is the first to add actual flags. How would you like to resolve this?

xml/vk.xml
```diff
@@ -9078,6 +9080,23 @@ typedef void* <name>MTLSharedEvent_id</name>;
...
+        </type>
+        <type category="struct" name="VkPhysicalDeviceMapMemoryPlacedPropertiesEXT" returnedonly="true" structextends="VkPhysicalDeviceProperties2">
+            <member values="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MAP_MEMORY_PLACED_PROPERTIES_EXT"><type>VkStructureType</type> <name>sType</name></member>
+            <member optional="true"><type>void</type>*            <name>pNext</name></member>
+            <member limittype="min,pot"><type>VkDeviceSize</type> <name>minPlacedMemoryMapAlignment</name></member>
+        </type>
+        <type category="struct" name="VkMemoryMapPlacedInfoEXT" structextends="VkMemoryMapInfoKHR">
+            <member values="VK_STRUCTURE_TYPE_MEMORY_MAP_PLACED_INFO_EXT"><type>VkStructureType</type> <name>sType</name></member>
+            <member optional="true">const <type>void</type>*  <name>pNext</name></member>
+            <member noautovalidity="true"><type>void</type>*  <name>pPlacedAddress</name></member>
```
DadSchoorse, line 9098:
```suggestion
            <member noautovalidity="true">const <type>void</type>*  <name>pPlacedAddress</name></member>
```
It's read only.

oddhack:
> Unfortunately, my fix for the dependency warning caused a new CI issue. It's complaining that I put the requires for `VkMemoryUnmapFlagBitsKHR` in VK_EXT_memory_map_placed which uses an EXT suffix. However, this is consistent with other cases where we add an unused flags field in an extension or core version and then a different extension is the first to add actual flags. How would you like to resolve this?
> 
> I've moved it back to VK_KHR_map_memory2

It looks like you squashed away the offending commit? If you happen to recall the diff I'd like to take a look at it for future extensions but it appears this is working OK now.

oddhack:
@gfxstrand this is saying something about @tobski requested changes though I can't figure out what was requested that you haven't already addressed. Maybe git is confused? Also there's the suggestion from @DadSchoorse above. So I will hold off on merging until tomorrow and we can check back on the call.

xml/vk.xml
DadSchoorse: Actually, I'm not so sure anymore, given that it will be passed to mmap which uses plain `void*`.

Tobski:
> @gfxstrand this is saying something about @Tobski requested changes though I can't figure out what was requested that you haven't already addressed. Maybe git is confused? Also there's the suggestion from @DadSchoorse above. So I will hold off on merging until tomorrow and we can check back on the call.

I'm pretty sure I just asked for a proposal, which has now been done. Will review again and clear this up!

appendices/VK_EXT_map_memory_placed.adoc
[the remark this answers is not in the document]
Tobski: We've generally been dropping them from the appendix rather than the proposal - reason being the questions usually come up when you read the proposal, not the appendix. Suggest dropping them here? Not going to gate merging on this but it is weird to have them duped...

proposals/VK_EXT_map_memory_placed.adoc
```diff
@@ -0,0 +1,178 @@
...
+placement of `VkDeviceMemory` maps.
+
+## Problem Statement
+
+There are certain cases in which it can be useful to directly control the
+address used by a memory map operation.  One example of this is in
+emulation environments when the application under emulation is 32-bit but
+the host userspace is 64-bit.  In order to avoid additional copies or
+address translation, the emulator can choose an address in the lower 32
+bits of the address space and request that the map be placed there.
```
Tobski, line 20: Somewhat nitpicky since we aren't running the reflow tools on proposals (maybe we should?), but for the sake of future diffing it would be nice to have sentences start on new lines.

Tobski (approved):
A couple of nitpicky editorial things that I'd _like_ to see fixed but it's nothing I feel needs to be resolved before merging.

proposals/VK_EXT_map_memory_placed.adoc
gfxstrand: Done. I double-checked the appendix as well.

xml/vk.xml
gfxstrand: Also, we return it as a `void *` via the `ppData` out parameter so we'd be effectively casting const away on behalf of the app. That doesn't sound like a great idea, either.

DadSchoorse: Yeah, I think `void*` is better.

PR #1913, opened 2022-08-12
Remove redundant drawIndirectFirstInstance VUID

sjfricke:
There is already 2 VUIDs for `drawIndirectFirstInstance` in the struct and no need for it in the function as well

> VUID-VkDrawIndirectCommand-firstInstance-00501
If the drawIndirectFirstInstance feature is not enabled, firstInstance must be 0

> VUID-VkDrawIndexedIndirectCommand-firstInstance-00554
If the drawIndirectFirstInstance feature is not enabled, firstInstance must be 0

Removing the function in favor of the structure VUID seem more consistent with rest of VUs


oddhack:
LGTM. I think those VUID numbers are so old they may predate getting clarity on putting VUs at the API where all the necc. information to test them is available, but not higher.

PR #1916, opened 2022-08-14
Add standalone VU for OpImageWrite component mismatch

sjfricke:
Currently trying to implement both VUID `04115` and `04469` in the Validation Layers, but when I have the following shader

```swift
 %type_image = OpTypeImage %uint 2D 0 0 0 2 Rgba8ui
    %texelU3 = OpConstantComposite %v3uint %uint_1 %uint_1 %uint_1
               OpImageWrite %image %int_1 %texelU3 ZeroExtend
```

`Rgba8ui` should become `VK_FORMAT_R8G8B8A8_UINT` and only writing 3 components triggers 

> VUID-vkCmdDraw-None-04115
If a VkImageView is accessed using OpImageWrite as a result of this command, then the Type of the Texel operand of that instruction must have at least as many components as the image view’s format

but I would purpose it might be more beneficial to have this checked at `vkCreateShaderModule` time instead of waiting until draw time (unless someone can point to CTS tests verifying all driver's compilers can handle a shader as given above)

This change makes the draw time VU only for `Unknown` `OpTypeImage` and then adds a separate Runtime SPIR-V Validation for the case where it is not `Unknown`

----

Another way to view this issue is the spec currently says

> If the image format of the OpTypeImage is not compatible with the VkImageView’s format, the write causes the contents of the image’s memory to become undefined.

So not sure if this means if I have `OpTypeImage %uint 2D 0 0 0 2 Rgba8ui` with a `OpImageWrite` of 3 components and then use an `VkImageView` format of `VK_FORMAT_R8G8B8_UINT` if we expect it to be **invalid** or **undefined**

### DIFF
--- a/appendices/spirvenv.txt
+++ b/appendices/spirvenv.txt
@@ -614,9 +614,9 @@ ifndef::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
     operand of 2 and an "`Image Format`" operand of code:Unknown must: be
     decorated with code:NonReadable.
 endif::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
-  * code:OpImageWrite to any variable created with a code:Type of
-    code:OpTypeImage must: have the code:Texel operand contain at least
-    as many components as the code:Image code:Format according to the
+  * code:OpImageWrite to any code:Image whose code:Image code:Format is not
+    code:Unknown must: have the code:Texel operand contain at least
+    as many components as the corresponding elink:VkFormat as given in the
     <<spirvenv-image-formats,SPIR-V Image Format compatibility table>>
   * [[VUID-{refpage}-Location-06272]]
     The sum of code:Location and the number of locations the variable it
--- a/appendices/spirvenv.txt
+++ b/appendices/spirvenv.txt
@@ -615,9 +615,8 @@ ifndef::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
     decorated with code:NonReadable.
 endif::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
   * code:OpImageWrite to any variable created with a code:Type of
-    code:OpTypeImage that has a code:Sampled operand of 2 must: have the
-    code:Texel operand contain at least as many components as the
-    code:Image code:Format according to the
+    code:OpTypeImage must: have the code:Texel operand contain at least
+    as many components as the code:Image code:Format according to the
     <<spirvenv-image-formats,SPIR-V Image Format compatibility table>>
   * [[VUID-{refpage}-Location-06272]]
     The sum of code:Location and the number of locations the variable it
--- a/appendices/spirvenv.txt
+++ b/appendices/spirvenv.txt
@@ -614,6 +614,11 @@ ifndef::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
     operand of 2 and an "`Image Format`" operand of code:Unknown must: be
     decorated with code:NonReadable.
 endif::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
+  * code:OpImageWrite to any variable created with a code:Type of
+    code:OpTypeImage that has a code:Sampled operand of 2 must: have the
+    code:Texel operand contain at least as many components as the
+    code:Image code:Format according to the
+    <<spirvenv-image-formats,SPIR-V Image Format compatibility table>>
   * [[VUID-{refpage}-Location-06272]]
     The sum of code:Location and the number of locations the variable it
     decorates consumes must: be less than or equal to the value for the
--- a/chapters/commonvalidity/draw_dispatch_common.txt
+++ b/chapters/commonvalidity/draw_dispatch_common.txt
@@ -215,15 +215,17 @@ ifdef::VK_VERSION_1_1,VK_KHR_sampler_ycbcr_conversion[]
     object must: not use the code:ConstOffset and code:Offset operands
 endif::VK_VERSION_1_1,VK_KHR_sampler_ycbcr_conversion[]
   * [[VUID-{refpage}-None-04115]]
-    If a slink:VkImageView is accessed using code:OpImageWrite as a result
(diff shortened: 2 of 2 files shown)


Tobski:
Discussed this in the WG today, agreed this makes sense.

Tobski (approved):

appendices/spirvenv.txt
```diff
@@ -614,6 +614,11 @@ ifndef::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
     operand of 2 and an "`Image Format`" operand of code:Unknown must: be
     decorated with code:NonReadable.
 endif::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
+  * code:OpImageWrite to any variable created with a code:Type of
+    code:OpTypeImage that has a code:Sampled operand of 2 must: have the
+    code:Texel operand contain at least as many components as the
+    code:Image code:Format according to the
```
gnl21, line 620: It's not immediately clear to me how we would determine the number of components "according to the table". I expected to see a component count in there but there isn't one. (It's not too hard to work out what the correct answer is, but it would be nice to be more explicit here).

chapters/commonvalidity/draw_dispatch_common.txt
```diff
@@ -215,15 +215,17 @@ ifdef::VK_VERSION_1_1,VK_KHR_sampler_ycbcr_conversion[]
...
   * [[VUID-{refpage}-None-04115]]
-    If a slink:VkImageView is accessed using code:OpImageWrite as a result
-    of this command, then the code:Type of the code:Texel operand of that
-    instruction must: have at least as many components as the image view's
-    format
+    If as a result of this command a slink:VkImageView is accessed using
+    code:OpImageWrite with an code:OpTypeImage that has an "`Image Format`"
+    operand of code:Unknown, then the code:Type of the code:Texel operand
+    of that instruction must: have at least as many components as the
+    image view's format
```
gnl21, line 222: I wonder if it's a little bit misleading including "that has an Image Format operand of Unknown" here. It's not that these are allowed if the operand is not unknown, but rather that you can't have got this far in that case. It might be a little nicer to remove that phrase?

appendices/spirvenv.txt
```diff
@@ -614,6 +614,11 @@ ifndef::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
     operand of 2 and an "`Image Format`" operand of code:Unknown must: be
     decorated with code:NonReadable.
 endif::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
+  * code:OpImageWrite to any variable created with a code:Type of
+    code:OpTypeImage that has a code:Sampled operand of 2 must: have the
```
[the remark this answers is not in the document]
sjfricke: agree, I copied this language from another VU, but I am very happy to remove it

appendices/spirvenv.txt
sjfricke: So luckily the `vk.xml` has all this information now (which these tables should be generated from in the spec too), an example

```xml
        <format name="VK_FORMAT_R8G8_UINT" class="16-bit" blockSize="2" texelsPerBlock="1">
            <component name="R" bits="8" numericFormat="UINT"/>
            <component name="G" bits="8" numericFormat="UINT"/>
            <spirvimageformat name="Rg8ui"/>
        </format>
 ```
 
 So it will be possible to generate this check in the validation layers to map `Rg8ui` in SPIR-V to having 2 `<component>` tags

File: chapters/commonvalidity/draw_dispatch_common.txt (hunk 2 above)
Comment (sjfricke) [reply chapters/commonvalidity/draw_dispatch_common.txt:222]: I agree, I guess it was easier to have here to justify **why** the other VU is needed, but agree it makes a hard VU to read even harder to read

**Comment by oddhack** [comment]
@gnl21 are you OK with merging this, with your comment addressed? I think between you and @tobias we can call this good.

File: appendices/spirvenv.txt (hunk 1 above)
Comment (gnl21) [reply appendices/spirvenv.txt]: My concern was more for someone reading the spec, rather than getting this right in the validation layers (although having this autogenerated there is definitely a good thing). My first thought is that this should say something like "at least as many components as are present in the corresponding Vulkan format, as given in table ...". That doesn't work for Unknown though, for which the corresponding format is "Any".

**Comment by gnl21** [comment]
I'm OK with merging this. Ideally I'd like some change based on my last comment, but I'm fine with it as is if you think I'm being too pendantic!

File: appendices/spirvenv.txt (hunk 1 above)
Comment (sjfricke) [reply appendices/spirvenv.txt]: I plan to take on https://github.com/KhronosGroup/Vulkan-Docs/issues/1920 soon (this week probably) and in there I would be more than happy to add some information/column for how many components it is

File: appendices/spirvenv.txt [hunk 4]
```diff
@@ -614,6 +614,10 @@ ifndef::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
     operand of 2 and an "`Image Format`" operand of code:Unknown must: be
     decorated with code:NonReadable.
 endif::VK_VERSION_1_3,VK_KHR_format_feature_flags2[]
+  * code:OpImageWrite to any variable created with a code:Type of
+    code:OpTypeImage must: have the code:Texel operand contain at least
+    as many components as the code:Image code:Format according to the
+    <<spirvenv-image-formats,SPIR-V Image Format compatibility table>>
```
gnl21, line 620:
```suggestion
  * code:OpImageWrite to any code:Image whose code:Image code:Format is not
    code:Unknown must: have the code:Texel operand contain at least
    as many components as the corresponding elink:VkFormat as given in the
    <<spirvenv-image-formats,SPIR-V Image Format compatibility table>>
```

It's a bit wordier than I would like, but I think it's unambiguous. I removed the "variable with type code:OpTypeImage" because that is required for all writes. (I think that's left over from the language about Sampled operand being 2)

sjfricke: This looks way better! Thanks as always for time/energy in helping rewording these "fun" VUs

PR #1938, opened 2022-09-09
Add SPIR-V VU for mesh shader output count

sjfricke:
The VUs removed are already in the SPIR-V section, for example `VUID-VkPipelineShaderStageCreateInfo-stage-07061` is the same as 

> VUID-RuntimeSpirv-MeshEXT-07115
For mesh shaders using the MeshEXT Execution Model the OutputVertices OpExecutionMode must be less than or equal to VkPhysicalDeviceMeshShaderPropertiesEXT::maxMeshOutputVertices.

------

The 2 sets of VUs added make the following 2 shaders illegal

```swift
OpExecutionMode %mainMesh OutputVertices 1
OpExecutionMode %mainMesh OutputPrimitivesEXT 1
OpSetMeshOutputsEXT 2 2 
```

```swift
OpExecutionMode %mainMesh OutputVertices 0
OpExecutionMode %mainMesh OutputPrimitivesEXT 0
; Can't find spec language saying OpSetMeshOutputsEXT MUST be called if no other Output is also not written
```

### DIFF
--- a/appendices/spirvenv.adoc
+++ b/appendices/spirvenv.adoc
@@ -595,6 +595,10 @@ endif::VK_VERSION_1_1[]
     In mesh shaders using the code:MeshEXT {ExecutionModel} variables in
     workgroup or private storage class declared as or containing a composite
     type must: not be accessed by indices that depend on code:ViewIndex
+  * In mesh shaders using the code:MeshEXT {ExecutionModel} the
+    code:OutputVertices code:OpExecutionMode must: be greater than 0
+  * In mesh shaders using the code:MeshEXT {ExecutionModel} the
+    code:OutputPrimitivesEXT code:OpExecutionMode must: be greater than 0
   * [[VUID-{refpage}-Input-07290]]
     Variables with a storage class of code:Input or code:Output and a type
     of code:OpTypeBool must: be decorated with the code:BuiltIn decoration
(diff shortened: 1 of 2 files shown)


appendices/spirvenv.adoc
```diff
@@ -1009,11 +1013,17 @@ ifdef::VK_EXT_mesh_shader[]
     For mesh shaders using the code:MeshEXT {ExecutionModel} the
     code:OutputVertices code:OpExecutionMode must: be less than or equal to
     sname:VkPhysicalDeviceMeshShaderPropertiesEXT::pname:maxMeshOutputVertices.
+  * For mesh shaders using the code:MeshEXT {ExecutionModel} the
```
HansKristian-Work, line 1016: I don't think this belongs here. The value passed to SetMeshOutputsEXT is a dynamic value, and therefore belongs in runtime VU.

There's two parts to this:
- The MaxPrimitive/Vertex value is statically evaluated against the properties struct.
- The SetMeshOutputsEXT one is runtime evaluated against the shader max value.

sjfricke: >  therefore belongs in runtime VU.

This is the SPIRV Runtime VU section

HansKristian-Work: You're right. The other VUs around it didn't look like it and had to scroll up 100 lines to see it.

HansKristian-Work (approved):

PR #1941, opened 2022-09-10
Fix VK_EXT_shader_image_atomic_int64 ifdef location VUID

sjfricke:
Was going to add validation for `04470`,`04471`,`04472`,`04473` and don't think they require `VK_EXT_shader_image_atomic_int64` in order to validate

> VUID-vkCmdDraw-SampledType-04470
If a [VkImageView](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkImageView) with a [VkFormat](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkFormat) that has a 64-bit component width is accessed as a result of this command, the SampledType of the OpTypeImage operand of that instruction must have a Width of 64

### DIFF
--- a/chapters/commonvalidity/draw_dispatch_common.adoc
+++ b/chapters/commonvalidity/draw_dispatch_common.adoc
@@ -234,7 +234,6 @@ endif::VK_VERSION_1_1,VK_KHR_sampler_ycbcr_conversion[]
     of this command, then the code:Type of the code:Texel operand of that
     instruction must: have at least as many components as the buffer view's
     format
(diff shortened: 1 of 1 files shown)


sjfricke:
ok, so without `Int64ImageEXT` you can never get to these VUs, but I guess it was not intuitive that currently it only possible with `VkPhysicalDeviceShaderImageAtomicInt64FeaturesEXT::shaderImageInt64Atomics`

PR #1965, opened 2022-10-03
Clarify VK_EXT_mesh_shader builtin execution modes

sjfricke:
About to add validation support and thought it meant "if another `OpExecutionMode` is used it is invalid" which doesn't make sense so updated it to be what I think it was intended to mean from the SPIR-V spec saying

> Each OpEntryPoint with the MeshEXT Execution Model must have an OpExecutionMode with exactly one of OutputPoints, OutputLinesEXT, or OutputTrianglesEXT Execution Modes.

### DIFF
--- a/chapters/interfaces.adoc
+++ b/chapters/interfaces.adoc
@@ -3368,8 +3368,8 @@ primitives.
     The code:PrimitivePointIndicesEXT decoration must: be used only within
     the code:MeshEXT {ExecutionModel}
   * [[VUID-{refpage}-PrimitivePointIndicesEXT-07042]]
-    The code:PrimitivePointIndicesEXT decoration must: be used only with the
-    the code:OutputPoints {ExecutionMode}
+    The code:PrimitivePointIndicesEXT decoration must: not be used with the
+    the code:OutputLinesEXT or code:OutputTrianglesEXT {ExecutionMode}
   * [[VUID-{refpage}-PrimitivePointIndicesEXT-07043]]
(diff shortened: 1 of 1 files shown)


chapters/interfaces.adoc
```diff
@@ -3368,8 +3368,8 @@ primitives.
     The code:PrimitivePointIndicesEXT decoration must: be used only within
     the code:MeshEXT {ExecutionModel}
   * [[VUID-{refpage}-PrimitivePointIndicesEXT-07042]]
-    The code:PrimitivePointIndicesEXT decoration must: be used only with the
-    the code:OutputPoints {ExecutionMode}
+    The code:PrimitivePointIndicesEXT decoration must: not be used with the
```
HansKristian-Work, line 3371: Wouldn't it be clearer to just say that if a builtin is used then the execution mode must: be equal to OutputPoints for example?

HansKristian-Work: I think that's what the spirit of "used only with" was trying to express.

sjfricke: :100: that seems better to me too, will fix

HansKristian-Work (approved):

chapters/interfaces.adoc
TomOlson: WG 2022-10-05: We had the same idea in discussion - if you list the things it must NOT be, then if we add more execution modes, you have to remember to update these lists. So it seems better to say what it MUST be. But we're willing to be convinced if there's something you're trying to do here that we are not understanding.

PR #1978, opened 2022-11-04
Remove duplicate sentence in VK_EXT_subpass_merge_feedback.

qbojj:
VK_EXT_subpass_merge_feedback had repeated 
"If a sname:VkRenderPassCreationControlEXT structure is included in the pname:pNext chain of slink:VkRenderPassCreateInfo2 and its value of pname:disallowMerging is ename:VK_TRUE, the implementation will disable subpass merging for the entire render pass." 2 times one after the other.

Removed one of them


janharaldfredriksen-arm:
Hi, thanks for looking at this.

These aren't redundant statements, though. The two sentences describe what happens when VkRenderPassCreationControlEXT is included in the pNext chain of VkRenderPassCreateInfo2 and VkSubpassDescription2, respectively.

There may be a way to write this with less duplicated words if you wanted to suggest that.

qbojj:
Oh... I have overlooked that. It could be 'included in the pNext chain VkRenderPassCreateInfo2 or VkSubpassDescription2' but it would be not homogenous with the rest of the API.

PR #1983, opened 2022-11-11
Reserve a driver ID for NVK

jekstrand:


oddhack:
FYI this is in Vulkan-Headers and the HEAD vk.xml for the 1.3.235 spec update, but I got a merge out of order between the public and internal repos, and it does not appear in the spec HTML/PDF - will be corrected next week. I don't think this justifies doing a full re-release so am not planning to do anything about it now.

jekstrand:
No worries!  It caused a bit of confusion with our Mesa header update but I got it sorted now.

oddhack:
Sigh. This somehow managed to mess up the git history once *again* in the .236 update and temporarily remove itself from HEAD. I put it back in and made sure the headers and XML still have it in Vulkan-Headers as well, but it's going to take another week to propagate through to the spec HTML/PDF. Sorry.

PR #1997, opened 2022-12-02
Add `VK_PIPELINE_STAGE_2_ACCELERATION_STRUCTURE_COPY_BIT_KHR` to access masks VUIDs

Rua:
The bit was missing from these VUIDs, so this PR adds it.


Tobski (approved):
Looks good to me!

dgkoch (approved):
This seems to be more correct than what is there right now, but we might need to do some follow ups to clarify the meaning of the sync2 bits.

oddhack:
> This seems to be more correct than what is there right now, but we might need to do some follow ups to clarify the meaning of the sync2 bits.

@dgkoch I'm interpreting that as a signoff - if not let me know by Wed. afternoon.

dgkoch:
Yes - that was signoff to merge.

PR #2000, opened 2022-12-08
Fix a typo in vkCmdSetCoverageToColorLocationNV

Kelebek1:
coverageToColorLocation is in VkPipelineCoverageToColorStateCreateInfoNV, not VkPipelineSampleLocationsStateCreateInfoEXT.

### DIFF
--- a/chapters/fragops.adoc
+++ b/chapters/fragops.adoc
@@ -1849,7 +1849,7 @@ drawing commands when the graphics pipeline is created with
 ename:VK_DYNAMIC_STATE_COVERAGE_TO_COLOR_LOCATION_NV set in
 slink:VkPipelineDynamicStateCreateInfo::pname:pDynamicStates.
 Otherwise, this state is specified by the
-slink:VkPipelineSampleLocationsStateCreateInfoEXT::pname:coverageToColorLocation
+slink:VkPipelineCoverageToColorStateCreateInfoNV::pname:coverageToColorLocation
 value used to create the currently active pipeline.
 
 .Valid Usage


chapters/fragops.adoc
```diff
@@ -1849,7 +1849,7 @@ drawing commands when the graphics pipeline is created with
 ename:VK_DYNAMIC_STATE_COVERAGE_TO_COLOR_LOCATION_NV set in
 slink:VkPipelineDynamicStateCreateInfo::pname:pDynamicStates.
 Otherwise, this state is specified by the
-slink:VkPipelineSampleLocationsStateCreateInfoEXT::pname:coverageToColorLocation
+slink:VkPipelineCoverageToColorStateCreateInfoNV::pname:coverageToColorLocation
```
oddhack, line 1852: I don't think that's a typo - there's no such NV API. This paragraph is saying that you can set the dynamic state using the NV call, or fall back to the default value specified when the pipeline was created, which is set using a separate EXT extension.

Please close the PR if you agree.

Kelebek1: What do you mean there's no such API? https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineCoverageToColorStateCreateInfoNV.html -- it's in here.

Compare these two pages: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdSetCoverageToColorEnableNV.html -- https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdSetCoverageToColorLocationNV.html

ColorEnable has the correct link, ColorLocation links to the wrong struct.

oddhack: Sorry, misread on my part. @dgkoch @pdaniell-nv could you take a look at this?

pdaniell-nv (approved):
Looks good. Thanks for making the fix.

PR #2011, opened 2022-12-18
Fix `VK_FORMAT_R32_SFLOAT` note

teoxoy:
`VK_FORMAT_R32_SFLOAT` should support `VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT` without `VK_EXT_shader_atomic_float`


spencer-lunarg:
@teoxoy this is not rendering as I think you intended it

image

oddhack:
> @teoxoy this is not rendering as I think you intended it (But I agree with the idea)

Asciidoctor conditional directives are very constrained - must be at the start of a line and must end the line following the end of the condition (the ']'). The most direct way to do this is to replicate the table row inside an ifdef:: and an ifndef:: with appropriate changes. Another option is to set an attribute conditionally, e.g.

> ifdef::VK_EXT_shader_atomic_float[:someattrname: {sym2}] 
> ifndef::VK_EXT_shader_atomic_float[:someattrname: {sym2}]  
> ...  
> Use {someattrname} in the table cell instead of {sym2}

I think the first is marginally preferable in this case.

teoxoy:
Ah, I see, so that's why the `ifdef`s for `VK_EXT_shader_image_atomic_int64` were on new lines.
I moved the `ifdef` to a new line to keep things consistent. Let me know if that works for you!

spencer-lunarg:
it did fix it

image

PR #2012, opened 2022-12-18
Gate `VUID-StandaloneSpirv-OpVariable-04734` behind Vulkan 1.3 or ext

teoxoy:
`VUID-StandaloneSpirv-OpVariable-04734` should require `VK_VERSION_1_3` or `VK_KHR_zero_initialize_workgroup_memory`.

### DIFF
--- a/appendices/spirvenv.adoc
+++ b/appendices/spirvenv.adoc
@@ -206,10 +206,12 @@ endif::VK_VERSION_1_1[]
     Any code:OpVariable with an code:Initializer operand must: have
     *Output*, *Private*, *Function*, or *Workgroup* as its *Storage Class*
     operand
+ifdef::VK_VERSION_1_3,VK_KHR_zero_initialize_workgroup_memory[]
   * [[VUID-{refpage}-OpVariable-04734]]
     Any code:OpVariable with an code:Initializer operand and *Workgroup* as
     its *Storage Class* operand must: use code:OpConstantNull as the
     initializer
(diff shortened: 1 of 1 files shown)


spencer-lunarg:
So all `VUID-StandaloneSpirv` should generally not need to be guarded as they can be ran offline (`spirv-val`) without caring about the extensions being used

`VUID-RuntimeSpirv-shaderZeroInitializeWorkgroupMemory-06372` and `VUID-RuntimeSpirv-OpVariable-06373` already cover the `ifdef::VK_VERSION_1_3,VK_KHR_zero_initialize_workgroup_memory[]` part of this

teoxoy:
Ah, I see. I wasn't familiar with the different flavors of `VUID`s.

Although it seems that these lines in the same section of the file are conditioned:
```adoc
ifdef::VK_VERSION_1_1[]
  * [[VUID-{refpage}-None-04642]]
    *Scope* for <<shaders-group-operations,group operations>> must: be
    limited to *Subgroup*
endif::VK_VERSION_1_1[]
ifndef::VK_VERSION_1_1[]
  * [[VUID-{refpage}-SubgroupVoteKHR-06997]]
    If the code:SubgroupVoteKHR or code:SubgroupBallotKHR capability is not
    declared, *Scope* for memory must: not be *Subgroup*
endif::VK_VERSION_1_1[]
```

spencer-lunarg:
That is the only exception because it is not about which extension is enabled, but if it is SPIR-V 1.0 vs SPIR-V 1.3 that matter for that case

PR #935, opened 2019-03-06
Fix random markup

krOoze:

PR #2034, opened 2023-01-11
Clarify DebugUtils object support

MarkY-LunarG:
The DebugUtils commands that allow a developer to associate data with a given object handle were unclear in a number of places. In particular, if the object passed in was an instance-level or physical-device-level object, what were the restrictions. This language attempts to rectify that.

Associated Issues: #1668

NOTE: This replaces PR #1901 which was submitted to resolve this issue but appears to have gone stale.  But @pixelcluster deserves kudos for attempting to get this fixed since it lead to many good discussions with @krOoze as well.

### DIFF
--- a/chapters/VK_EXT_debug_utils.adoc
+++ b/chapters/VK_EXT_debug_utils.adoc
@@ -28,9 +28,19 @@ callback to capture debug messages from a variety of Vulkan components.
 
 It can be useful for an application to provide its own content relative to a
 specific Vulkan object.
+
 The following commands allow application developers to associate
 user-defined information with Vulkan objects.
-
+These commands are device-level commands but they may: reference instance-level
+objects (such as slink:VkInstance) and physical device-level objects (such as
+slink:VkPhysicalDevice) with a few restrictions:
+  * The data for the corresponding object may: still be available after the
+    slink:VkDevice used in the corresponding API call to set it is destroyed,
+    but access to this data is not guaranteed and should be avoided.
+  * Subsequent calls to change the data of the same object across multiple
+    sname:VkDevice objects, may: result in the data being changed to the most
+    recent version for all sname:VkDevice objects and not just the
+    sname:VkDevice used in the most recent API call.
 
 [[debugging-object-naming]]
 ==== Object Naming
@@ -42,7 +52,8 @@ fname:vkSetDebugUtilsObjectNameEXT as defined below.
 --
 include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.adoc[]
 
-  * pname:device is the device that created the object.
+  * pname:device is the device that is associated with the named object passed
+    in via pname:objectHandle.
   * pname:pNameInfo is a pointer to a slink:VkDebugUtilsObjectNameInfoEXT
     structure specifying parameters of the name to set on the object.
 
@@ -52,6 +63,17 @@ include::{generated}/api/protos/vkSetDebugUtilsObjectNameEXT.adoc[]
     pname:pNameInfo->objectType must: not be ename:VK_OBJECT_TYPE_UNKNOWN
   * [[VUID-vkSetDebugUtilsObjectNameEXT-pNameInfo-02588]]
     pname:pNameInfo->objectHandle must: not be dlink:VK_NULL_HANDLE
+  * If pname:pNameInfo->pname:objectHandle is the valid handle of an
+    instance-level object, the slink:VkDevice identified by pname:device must:
+    be a descendent of the same slink:VkInstance as the object identified by
+    pname:pNameInfo->pname:objectHandle
+  * If pname:pNameInfo->pname:objectHandle is the valid handle of a
+    physical-device-level object, the slink:VkDevice identified by pname:device
+    must: be a descendant of the same slink:VkPhysicalDevice as the object
+    identified by pname:pNameInfo->pname:objectHandle
+  * If pname:pNameInfo->pname:objectHandle is the valid handle of a device-level
+    object, that object must: be a descendent of the slink:VkDevice identified
+    by pname:device
 ****
 
 include::{generated}/validity/protos/vkSetDebugUtilsObjectNameEXT.adoc[]
@@ -124,6 +146,21 @@ include::{generated}/api/protos/vkSetDebugUtilsObjectTagEXT.adoc[]
   * pname:device is the device that created the object.
   * pname:pTagInfo is a pointer to a slink:VkDebugUtilsObjectTagInfoEXT
     structure specifying parameters of the tag to attach to the object.
+
+.Valid Usage
+****
+  * If pname:pNameInfo->pname:objectHandle is the valid handle of an
+    instance-level object, the slink:VkDevice identified by pname:device must:
+    be a descendent of the same slink:VkInstance as the object identified by
+    pname:pNameInfo->pname:objectHandle
+  * If pname:pNameInfo->pname:objectHandle is the valid handle of a
+    physical-device-level object, the slink:VkDevice identified by pname:device
+    must: be a descendant of the same slink:VkPhysicalDevice as the object
+    identified by pname:pNameInfo->pname:objectHandle
+  * If pname:pNameInfo->pname:objectHandle is the valid handle of a device-level
+    object, that object must: be a descendent of the slink:VkDevice identified
+    by pname:device
+****
 
 include::{generated}/validity/protos/vkSetDebugUtilsObjectTagEXT.adoc[]
 --


MarkY-LunarG:
@charles-lunarg and @bradgrantham-lunarg, this is the language that some of us were leaning to in the original PR.  But I looked at [GfxReconstruct Issue #935](https://github.com/LunarG/gfxreconstruct/issues/935) and am wondering if this just makes it more likely that we will see crashes?

Was the crash because the loader didn't intercept the call and properly de-reference the instance and/or physical device objects or some other reason?

bradgrantham-lunarg:
It's important that this is fixed but if an app gets the function pointer from **GetInstanceProcAddr** setting a debug name on a VkPhysicalDevice already works.
I don't know why GFXR is getting the wrong function pointer in https://github.com/LunarG/gfxreconstruct/issues/935 from **GetDeviceProcAddr**.  @charles-lunarg had suggested a possible loader bug (is there an issue or PR filed?)
I'm going to see about hand-coding those two functions in GFXR to use the instance procaddr rather than wait for a loader fix.

charles-lunarg:
Loader issue has been created to track it.
https://github.com/KhronosGroup/Vulkan-Loader/issues/1109

oddhack:
@MarkY-LunarG could we close #1901 at this point (with the OP's assent)? Also, please ping me when this MR is ready to merge from your perspective.

pixelcluster:
I'd be happy with the changes in this PR. I'll close #1901.

bradgrantham-lunarg:
FYI we've already applied a fix at https://github.com/LunarG/gfxreconstruct/pull/948 to use GetInstanceProcAddr.  So as long as that continues to work, GFXR should be fine.

chapters/VK_EXT_debug_utils.adoc
```diff
@@ -28,8 +28,18 @@ callback to capture debug messages from a variety of Vulkan components.
...
 user-defined information with Vulkan objects.
+These commands are device-level commands but they may: reference instance-level
+objects with a few restrictions.
+If the object is an instance-level object (such as a slink:VkInstance) or
+physical-device-level object (such as a slink:VkPhysicalDevice or
+slink:VkDisplayKHR), then the lifetime of the debug annotated data provided to
+one or more of the following commands is restricted to the lifetime of the
+associated slink:VkDevice object used as the parameter.
+Therefore, the data is only valid as long as the slink:VkDevice object is valid,
+or until a corresponding call is made to change the data for that object.
```
bradgrantham-lunarg, line 42: This is an interesting restriction.  If an application sets a debug object name on a VkInstance or VkPhysicalDevice from two different VKDevices, what happens?  Is it up to the implementation to pick the name for a debug message?

MarkY-LunarG: This is a device command, so only the VkPhysicalDevice associated with the device used to call would be used.  At least that's what I'm suggesting.

charles-lunarg: I can tell you from the loader's implementation that the restriction of only being able to name the VkPhysicalDevice that created the VkDevice is essential, since its not practical to 'get' the VkPhysicalDevice associated with non-created VkDevices. 
Since the desktop loader 'wrap's the VkPhysicalDevice handles, the loader must intercept and modify the VkPhysicalDevice parameters.

bradgrantham-lunarg: Okay, what if two different Devices created from two different PhysicalDevices are called to set the debug name for an Instance?  It's unlikely, but e.g.
```
vkCreateDevice(physical_device1, ..., &device1);
vkCreateDevice(physical_device2, ..., &device2);
vkSetDebugObjectNameEXT(device1, {VK_STRUCTURE_TYPE_DEBUG_UTILS_OBJECT_NAME_INFO_EXT,..., VK_OBJECT_TYPE_INSTANCE, instance, "an instance"} );
vkSetDebugObjectNameEXT(device2, {VK_STRUCTURE_TYPE_DEBUG_UTILS_OBJECT_NAME_INFO_EXT,..., VK_OBJECT_TYPE_INSTANCE, instance, "my instance"} );
```

charles-lunarg: The loader doesn't care in this case, it just forwards the `vkSetDebugObjectNameEXT` to whoever does. That said, I believe this is just a case of determining whether VkInstance should *always* have a single name, or if there should be a special case here.
I am for a single name, so the above code would result in VkInstance->"my instance".

bradgrantham-lunarg: That is ambiguous with the proposed wording - if device1 sets the instance's name, then device2 sets the name, then device2 is destroyed, the implication is that the instance's name either has to be cleared or reverts to the name set through device1.

I think maybe it's easier to omit the bit about the lifetime of the set name to the VkDevice's lifetime unless that's how implementations already work.

Maybe something about the name most recently set on the instance or the physical device sticks instead.

(The implication from all of this to me is that SetDebugObjectNameExt probably either should have been on Instance or current use to set the name on a Device's parent (e.g. vkquake does it on PhysicalDevice) should become disallowed / invalid usage.  Future maintenance extension or vulkan revision, I guess)

charles-lunarg: Okay after talking with brad, I think specifying the lifetime here causes more ambiguity and implications for implementations. The issue is that a debug utils implementation would have to 'track' *who* set the name and keep a list of names associated with the device who set it. Whereas the lack of lifetime specification would allow implementations to have a single name for the handle, which should cover all reasonable use cases.

MarkY-LunarG: But this doesn't solve the problem.  It depends on where the implementation stores the name.  If the name is associated in the device, it will go away and not be avaliable afterwords.  But, if it associates it with the object directly, then it will stick.  Since we weren't clear originally, I feel like we almost need to support both. 

So are we saying we just strip out the lifetime comment then and leave it:

```
These commands are device-level commands but they may: reference instance-level
objects with a few restrictions.
```

But what about the case where the implementations stores the names with the device.  Shouldn't we have a statement akin to `this data may go away on the destruction of the device`?

That's my concern.  If we don't clarify when the data is useful, then people will try to use it when it's not valid.

bradgrantham-lunarg: You didn't address the issue I brought up above; if names are set on a `VkInstance` or `VkPhysicalDevice` through two different `VkDevice`s, then one `VkDevice` is deleted, which name persists?  I'm trying to get to the fundamental question about to which object the name is attached.

I think the common use case you mentioned to me in person yesterday is that two `VkDevice`s are created, one for graphics use and the other to set debug names.  That implies the name on the `VkInstance` is expected to be valid outside the scope of the debug name `VkDevice`.  So it sounds to me like the name has to be attached to `VkInstance` and `VkPhysicalDevice` with no association with a child `VkDevice`.

We discussed in person yesterday that in the case a layer or implementation wants to print debug information about the `VkInstance` or `VkPhysicalDevice` that we could consider the name only valid in the context of calling a device function.  But that doesn't handle the common use case above.  Now I think the extension should just say "it sets the name, replacing any previously set name".  Then there's no ambiguity about lifetime if a name is set through two separate `VkDevice`s or if it's set through one `VkDevice` but a layer or implementation wants to print information about the `VkInstance` in a function on call on a different `VkDevice`.

So then the language could look like this if you just want to get this done and move on:

```
These commands are device-level commands but they may: reference instance-level
objects.  A name set on an instance-level object replaces any name previously set on
that object.
```

I think to do do this well we should probably just survey what apps and layers and implementations do and then retconn the extension to cover the most common case.  That's probably too much trouble to cover a small corner case that everyone is already working around.

MarkY-LunarG: Perhaps:

```
For example, all object data may be stored in such a way as to relate it to the VkDevice handle and may not be accessible after the destruction of the VkDevice, even if the data is associated with instance or physical device-level objects.
```

bradgrantham-lunarg: Does that mean you can't set instance or physical device names through one device and have them show up in debug prints in another device?

MarkY-LunarG: I think that's possible.  I'm trying to be vague enough to allow it since implementations of this have been out in the wild for 5 years or so now and I don't want to require new behavior, just clarify existing behavior.

chapters/VK_EXT_debug_utils.adoc
```diff
@@ -28,9 +28,19 @@ callback to capture debug messages from a variety of Vulkan components.
...
+
 The following commands allow application developers to associate
 user-defined information with Vulkan objects.
-
+These commands are device-level commands but they may: reference instance-level
+objects (such as slink:VkInstance) and physical device-level objects (such as
+slink:VkPhysicalDevice) with a few restrictions:
+  * The data for the corresponding object may: still be available after the
+    slink:VkDevice used in the corresponding API call to set it is destroyed,
+    but access to this data is not guaranteed and should be avoided.
```
bradgrantham-lunarg, line 39: You can probably omit `The data for the corresponding object may: still be available after the slink:VkDevice used in the corresponding API call to set it is destroyed, but access to this data is not guaranteed and should be avoided.` - there's no way for an application or layer to query the data anyway, as far as I can tell, and whether a layer or an implementation can still access it in their own data structires is up to how they hang the name off objects.

chapters/VK_EXT_debug_utils.adoc
```diff
@@ -28,9 +28,19 @@ callback to capture debug messages from a variety of Vulkan components.
...
+  * Subsequent calls to change the data of the same object across multiple
+    sname:VkDevice objects, may: result in the data being changed to the most
+    recent version for all sname:VkDevice objects and not just the
+    sname:VkDevice used in the most recent API call.
```
bradgrantham-lunarg, line 43: 👍

bradgrantham-lunarg (approved):
Just a thought that you don't have to mention accessing the data after device deletion, otherwise looks good!

charles-lunarg (approved):

MarkY-LunarG:
@oddhack , I think this is ready when you are.  Thanks

oddhack:
> @oddhack , I think this is ready when you are. Thanks

Thanks. Putting on the Maintenance telecon agenda although we probably will not meet next week since it's after the VF2F.

PR #2042, opened 2023-01-31
Separate reflow doc traversal and logic

ShabbyX:
The reflow code traversed asciidoc files and modified them.  The logic to traverse the files is split into its own class for future reuse.

### DIFF
--- a/scripts/doctransformer.py
+++ b/scripts/doctransformer.py
@@ -1,0 +1,437 @@
+# Copyright 2023 The Khronos Group Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Utilities for automatic transformation of spec sources.  Most of the logic
+has to do with detecting asciidoc markup or block types that should not be
+transformed (tables, code) and ignoring them.  It is very likely there are many
+asciidoc constructs not yet accounted for in the script, our usage of asciidoc
+markup is intentionally somewhat limited.
+"""
+
+import re
+import sys
+from reflib import logDiag, logWarn
+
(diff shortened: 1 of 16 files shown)


ShabbyX:
@oddhack this is preparation for the VU work. It separates the doc-traversal logic from the reflow/vuid logic. Later commits would use the doc-traversal logic for VU formatting.

You mentioned reflow.py is fragile, so I added some unit tests to make sure everything still works as expected, but do please give this a try. A good test would be running with this script on a commit before release and make sure the result is identical to the one that was fixed with the old reflow.py.

With the unit tests for reflow.py bootstrapped, it should be easy to add more tests as necessary.

oddhack:
This seems fine - identical results to the old scripts run on the previous release and also some personal scripts I do strange transformations with. I think we can just mark this signed off as an editorial issue though will mention it on the teleconference today.

oddhack:
BTW since you have developer access to this repo I'd find it helpful to put substantive PRs such as this as branches in the repo itself, which makes it easier to access them. Otherwise I'm left with again looking up the sequence of stuff to create a new remote in my clone, fetch a branch from it, etc. if I want to actually experiment with it.

ShabbyX:
> BTW since you have developer access to this repo I'd find it helpful to put substantive PRs such as this as branches in the repo itself, which makes it easier to access them. Otherwise I'm left with again looking up the sequence of stuff to create a new remote in my clone, fetch a branch from it, etc. if I want to actually experiment with it.

Got it. I'll eventually move the other PRs to branches of this repo when time comes to move forward with them.

PR #2051, opened 2023-02-09
Use {chapters}/ in vkspec.adoc instead of chapters/

ShabbyX:
{chapters}/ is used everywhere else.  This makes it possible to override the macro consistently across the spec.


oddhack:
No harm, although there's actually a reason for all the subsidiary includes to parameterize the path that doesn't apply here (it's not that we're expecting to ever target a different directory at build time, it's that the directory needs to be an absolute path rather than a relative one when building the refpages).

ShabbyX:
> ...doesn't apply here (it's not that we're expecting to ever target a different directory at build time ...

That expectation is broken here: https://github.com/KhronosGroup/Vulkan-Docs/pull/2053 :)

PR #2053, opened 2023-02-10
Add support for Codified VUs

ShabbyX:
The following is implemented in this change:

- Codified VU parsing
- Codified VU reformatting in source (reflow.py)
- Codified VU formatting for build
- Codified VU type checking
- Unit tests

There is no mention of codified VUs in the spec text however at this point. That will be added as VUs are converted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ SPECREMARK = from git branch: $(shell echo `git symbolic-ref --short HEAD 2> /de
 	     commit: $(shell echo `git log -1 --format="%H" 2> /dev/null || echo Git commit not available`)
 
 # Some of the attributes used in building all spec documents:
-#   chapters - absolute path to chapter sources
+#   chapters - absolute path to generated (preprocessed) chapter sources
 #   appendices - absolute path to appendix sources
 #   proposals - absolute path to proposal sources
 #   images - absolute path to images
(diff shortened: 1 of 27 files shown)


ShabbyX:
*Notes*:

- This first three commits in this change have their own PRs, so only the last commit is the subject of this PR
- The PR is a no-op from the point of view of users. Follow up PRs will officially introduce the language, convert a few VUs etc.
- Currently, the pdf output is not styled (justified, no highlighting, not monospace). @oddhack I need your help to point me to how I can make the pdf build process `[style]##something##` correctly.

ShabbyX:
> I'd like to avoid preprocessing the entire spec before building. Having asciidoctor itself call out to the VU generator from an extension will make life easier for most users, particularly when there are build errors that have to be tracked back to where they're happening in the source. Asciidoctor extensions like asciidoctor-diagram do something like this today.
> 
> Re PDF themes, see config/themes/pdf-theme.yml and https://docs.asciidoctor.org/pdf-converter/latest/theme/ for docs. It is unlike CSS and very limited, but you can do some basic stuff.

Thanks for the pointers. I have yet to look at creating an extension that gathers the necessary info and calls into python, but I will at some point. You are right that this can change line numbers during build and that can be annoying.

Before sorting that out (and dealing with pdf), and while we wait for the WG signing off on this direction, I'm going to spend some time sorting out the VVL side of things.

ShabbyX:
> I'd like to avoid preprocessing the entire spec before building. Having asciidoctor itself call out to the VU generator from an extension will make life easier for most users, particularly when there are build errors that have to be tracked back to where they're happening in the source. Asciidoctor extensions like asciidoctor-diagram do something like this today.
> 
> Re PDF themes, see config/themes/pdf-theme.yml and https://docs.asciidoctor.org/pdf-converter/latest/theme/ for docs. It is unlike CSS and very limited, but you can do some basic stuff.

I finally got around to doing this, and with some minor asciidoctor issues worked around, I got things working. This is now building using an asciidoctor extension. It spawns and communicates with a python process to do the job of formatting the VUs.

I also got the pdf output to look decent, though somehow the `bold` style doesn't get applied :shrug: 

Per your previous request, I opened a new PR from a local branch in this repo (https://github.com/KhronosGroup/Vulkan-Docs/pull/2081), so I'm gonna close this one.

PR #2066, opened 2023-02-22
Specify operator precedence in XML `depends` expressions

### LINKED ISSUE #2062: [registry] operator precedence in extension dependencies was left undefined
The new dependencies syntax is documented here:
https://registry.khronos.org/vulkan/specs/1.3/registry.html#_attributes_of_extension_tags

There are 3 operators: `,` OR, `+` AND, and `()` grouping, however, the precedence of the two AND & OR operators is not defined.

Intuitively, I would expect `ext1+ext2,ext3` to be grouped as `(ext1+ext2),ext3`, but the only example in `vk.xml` that uses both AND & OR right now also uses the grouping parentheses in a way that implies that the *opposite* precedence would apply without them:
```xml
<extension name="VK_KHR_buffer_device_address" depends="(VK_KHR_get_physical_device_properties2+VK_KHR_device_group),VK_VERSION_1_1">
```

That seems counter-intuitive, so I think it's worth being explicit in the spec, whichever way was intended.

/cc @gfxstrand

oddhack:
Closes #2062

### DIFF
--- a/registry.adoc
+++ b/registry.adoc
@@ -1392,7 +1392,7 @@ implemented against.
   * attr:type - required if the attr:supported attribute is not
     `'disabled'`.
     Must be either `'device'` or `'instance'`, if present.
-  * attr:depends - optional.
+  * [[depends-expressions]] attr:depends - optional.
     String containing a boolean expression of one or more API core version
     and extension names.
     The extension requires the expression in the string to be satisfied to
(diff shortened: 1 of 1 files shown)


registry.adoc
```diff
@@ -1392,14 +1392,17 @@ implemented against.
...
     String containing a boolean expression of one or more API core version
     and extension names.
     The extension requires the expression in the string to be satisfied to
     use any functionality it defines (for instance extensions), or to use
     any device-level functionality it provides (for device extensions).
     Supported operators include `,` for logical OR, `+` for logical AND, and
     `(` `)` for grouping.
+    `,` and `+` are of equal precedence, and lower than `(` `)`.
+    Expressions must be evaluated left-to-right for operators of the same
+    precedence.
```
1ace, line 1405: @oddhack so `a+b,c` is `(a+b),c` and `a,b+c` is `(a,b)+c`?

I'm not a fan of this (not intuitive for me, which makes it error-prone for anyone else who might think like me), but at least now it's documented, so thank you :+1:

PR #2073, opened 2023-03-08
Allow finding Python in other locations

brad0:
Although a lot of Linux OS's provide Python in /usr/bin that is not always the case and outside of Linux that is not the case.

### DIFF
--- a/scripts/Retired/checkLinks.py
+++ b/scripts/Retired/checkLinks.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 #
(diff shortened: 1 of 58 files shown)


oddhack:
What environment are you trying to build in that this is a problem? We are consciously restrictive about what we support after a few painful years of trying to figure out and maintain how to get the many toolchain components all up to date and running on everybody's desktop, and prefer people to use our Docker Debian image instead.

oddhack:
This has been addressed in the 1.3.290 spec update, since our Docker image has now been updated to a version that requires use of a Python venv.

PR #2076, opened 2023-03-09
Update vk.xml

aharon-abramson:
Add Mobileye vendor ID


oddhack:
> Thanks for merging this. Now I'm looking at https://github.com/KhronosGroup/dfdutils/blob/master/vulkan/vulkan_core.h, and I see that the highest vendor ID in VkVendorId is still VK_VENDOR_ID_MESA (0x10005), and even `VK_VENDOR_ID_POCL = 0x10006` doesn't appear there. Would you happen to know when this file gets regenerated?

FWIW the canonical reservations for vendor IDs are always in vk.xml in this repository (which is maybe not ideal in an abstract sense, but we started reserving them here first and then other APIs also wanted vendor IDs in a single namespace). dfdutils doesn't look like it gets updated very often but I believe @MarkCallow owns that repo and could sync up if needed.

MarkCallow:
The only reason for dfdutils having its own copy of `vulkan_core.h` is that it needs the formats defined by the Vulkan ASTC 3D extension which, as far as I know, is still unreleased. Once it is released my intention is to delete `vulkan_core.h` from dfdutils. Assuming the extension _is_still unreleased, please file an issue at dfdutils to remind me to update `vulkan_core.h`

PR #2081, opened 2023-03-16
Add support for codified VUs

ShabbyX:
The following is implemented in this change:

- Codified VU parsing
- Codified VU reformatting in source (reflow.py)
- Codified VU formatting for build
- Codified VU type checking
- Unit tests

There is no mention of codified VUs in the spec text however at this point.  That will be added as VUs are converted.


ShabbyX:
One limitation I couldn't work around is the inability to get source location information during asciidoctor AST traversal. I backported the relevant feature to asciidoctor v2.0 here https://github.com/asciidoctor/asciidoctor/pull/4417, and when landed and a new version released, needs an update in the docker image.

In the meantime, I've made sure to output enough information so the offending VU (that's failing build) can be identified.

ShabbyX:
Fixed some bugs + added comment support

@oddhack, I want to disable two-line headers (that look like this:

```
This is a header
================
```

asciidoctor's control for that is `Asciidoctor::Compliance.underline_style_section_titles`). Do we use that anywhere I didn't find?

It turns out that feature makes line number tracking messed up in asciidoctor.

oddhack:
> asciidoctor's control for that is `Asciidoctor::Compliance.underline_style_section_titles`). Do we use that anywhere I didn't find?

The style guide says to use only the single-line '== Title' form although I do not know that the other form is *never* used anywhere. Easiest way to find out is generate HTML with and without this control and compare them, probably.

oddhack:
@ShabbyX can we close this, since the WG has not taken up this direction?

ShabbyX:
I guess so

PR #2084, opened 2023-03-20
Clarify which formats are allowed as input attachments

### LINKED ISSUE #2083: Clarify which formats are allowed as input attachments
Section [§14.1.14¶2](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap14.html#descriptorsets-inputattachment) of the Vulkan specification (with all registered extensions) says:
> All image formats that are supported for color attachments (`VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT` or `VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV` ) ... are also supported for input attachments

But [§47.2](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap47.html#VkFormatFeatureFlagBits) says:
> `VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT` specifies that an image view can be used as a framebuffer color attachment that supports blending and as an **input attachment**.

The list of formats in parenthesis in section §14.1.14 should also include `VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT`

evyatark2:
Fixes  #2083

### DIFF
--- a/chapters/descriptorsets.adoc
+++ b/chapters/descriptorsets.adoc
@@ -402,7 +402,8 @@ an <<resources-image-views, image view>> that can: be used for
 fragment shaders.
 
 All image formats that are supported for color attachments
-(ename:VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT
+(ename:VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT,
+ ename:VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BLEND_BIT,
 ifdef::VK_NV_linear_color_attachment[]
  or ename:VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV
 endif::VK_NV_linear_color_attachment[]


chapters/descriptorsets.adoc
```diff
@@ -402,7 +402,8 @@ an <<resources-image-views, image view>> that can: be used for
 fragment shaders.
 
 All image formats that are supported for color attachments
-(ename:VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT
+(ename:VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT,
```
HansKristian-Work, line 405: I think the redundant part here is the mention of BLEND_BIT supporting input attachment. That is already covered by COLOR_ATTACHMENT_BIT, so I think the real fix is the remove the mention of input attachment from the description of BLEND_BIT.

HansKristian-Work: Ok, I'll propose the change internally, no problem.

HansKristian-Work: You can close the PR if you wish, but the issue can stay open.

PR #2087, opened 2023-03-20
Address feedback on enum validation 

### LINKED ISSUE #2077: Enumerant validity rules seem incorrect
1

> Any parameter of an enumerated type **must** be a valid enumerant for that type.

Sentence feels confusing. Probably means "any argument". Even so the sentence feels weirdly formed; it misses verbs or something and partly redundant to the next sentence, and the "of that type" seem partly redundant to the first bullet point.

----
2

Rules do not cover functions without dispatchable handle, such as `vkCreateInstance`. This impacts use of `VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR`, enumerants usable in the `pNext` chain, and potential new enumerants in the `VkAllocationCallbacks` structure.

---
3

> suffixed with _MAX_ENUM¹.

The "¹" points nowhere now.

---
4

> it was added by a core version that is supported (as reported by `vkEnumerateInstanceVersion`)

That is slightly incorrect. Many enumerants were added by core version 1.0, and such version cannot be queried by `vkEnumerateInstanceVersion`. There seem to be a Querying Version Support chapter, so I suggest linking to that, and improving that chapter.

---
5

> If the enumerant is used in a function that has a `VkPhysicalDevice` object as its first parameter

This seem incorrect for `vkCreateDevice` (something like #1540).

BTW, I think functions in Vulkan are generally called "commands".

---
6

> it was added by a core version that is supported by that device (as reported by `VkPhysicalDeviceProperties::apiVersion`);

missing "or"

It seems insufficient, and contradictory to the Valid Usage for Newer Core Versions, which says:

> Physical-device-level functionality or behavior added by a new core version of the API **must** not be used unless it is supported by the physical device as determined by `VkPhysicalDeviceProperties::apiVersion` and the specified version of `VkApplicationInfo::apiVersion`.

---
7

> it was added by an instance extension that was enabled for the instance; or

Probably should be explicit which "the instance".

---
8

> it was added by a device extension that is supported by that device.

Should say "physical device".

This seems incompatible with Vulkan 1.0 without `VK_KHR_get_physical_device_properties2`, which only permit use of enumerants added by instance extensions. Not sure if all extensions list these as dependency. Hypothetically, there could be an extension where use of its physical-device level commands is non-essential, therefore it does not list these dependencies.

---
9

> If the enumerant is used in a function that has any other dispatchable object as its first parameter and either:

Brittle in case there is ever an instance-level handle added.

---
10

> it was added by a core version that is supported for the device (as reported by `VkPhysicalDeviceProperties::apiVersion`); or

same as 6

---
11

> it was added by a device extension that was enabled for the device.

Should say which device, similarly to 7.  Since we have not even mentioned any device earlier, there is no "the device".

Tobski:
Fixes #2077

CC @krOoze

### DIFF
--- a/chapters/fundamentals.adoc
+++ b/chapters/fundamentals.adoc
@@ -877,12 +877,11 @@ Specification_, can: be `NULL`.
 [[fundamentals-validusage-enums]]
 ==== Valid Usage for Enumerated Types
 
-Any parameter of an enumerated type must: be a valid enumerant for that
-type.
-Use of an enumerant is valid if the following conditions are true:
+An enumerant value is valid for a particular use if the following conditions
+are true:
 
   * The enumerant is defined as part of the enumerated type.
-  * The enumerant is not a value suffixed with etext:_MAX_ENUM^1^.
+  * The enumerant is not a value suffixed with etext:_MAX_ENUM.
   ** This value exists only to ensure that C `enum` types are 32 bits in
      size and must: not be used by applications.
   * If the enumerant is used in a function that has a slink:VkInstance as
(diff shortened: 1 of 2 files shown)


Tobski:
@oddhack are the CIs failing here also failing on main? I can't see any reason for them to fail just from the commits here 🤔

oddhack:
> @oddhack are the CIs failing here also failing on main? I can't see any reason for them to fail just from the commits here thinking

Yes, there were some changes in Vulkan-Hpp in response to recent XML schema changes that aren't reflected here yet. Hopefully #2088 will fix that although Actions seems to be running terribly slowly at the moment.

oddhack:
@tobski if you rebase on main branch it should pass CI now. The Rust bindings are still not working but they're not a blocker.

chapters/fundamentals.adoc
```diff
@@ -877,12 +877,11 @@ Specification_, can: be `NULL`.
 [[fundamentals-validusage-enums]]
 ==== Valid Usage for Enumerated Types
 
-Any parameter of an enumerated type must: be a valid enumerant for that
-type.
-Use of an enumerant is valid if the following conditions are true:
+An enumerant value is valid for a particular use if the following conditions
+are true:
```
krOoze, line 881: maybe "all the following conditions".

chapters/initialization.adoc
```diff
@@ -233,6 +233,9 @@ include::{generated}/api/protos/vkEnumerateInstanceVersion.adoc[]
     of Vulkan supported by instance-level functionality, encoded as
     described in <<extendingvulkan-coreversions-versionnumbers>>.
 
+If flink:vkGetInstanceProcAddr returns a code:NULL pointer when this
```
krOoze, line 236: I think styleguide uses only ticks for NULL

chapters/fundamentals.adoc
```diff
@@ -897,18 +896,34 @@ endif::VK_VERSION_1_1[]
...
+     equal to the version that added it; or
   ** it was added by an <<extensions, instance extension>> that was enabled
-     for the instance; or
-  ** it was added by a <<extensions, device extension>> that is supported by
-     that device.
-  * If the enumerant is used in a function that has any other dispatchable
-    object as its first parameter and either:
+     for the current instance.
+  * If the enumerant is used in a function that has a dispatchable object
+    created or allocated from a device as its first parameter and either:
```
krOoze, line 920: This is incorrect. Either "dispatchable object created or allocated from a physical device" or "dispatchable object that is device or its children".

chapters/fundamentals.adoc
fluppeteer: We're inconsistent about whether we use "all". I agree it wouldn't hurt!

PR #2094, opened 2023-03-31
Fix typo in sample code

yknishidate:
[VK_EXT_shader_object] @daniel-story
Hi, I see the function `vkCmdCreateShadersEXT` is used in the shader object sample code, but maybe `Cmd` is not needed.


oddhack:
LGTM, thanks!

PR #2118, opened 2023-05-03
Add explanation for subgroup vs dynamically uniform.

### LINKED ISSUE #2112: Clarify the relationship between invocation group uniformity and subgroup uniformity
I am limiting scope of this to draw calls for now -- compute dispatches have some extra spec language I have not fully taken the time to understand yet.

The basic question here is whether subgroups can contain invocations from multiple invocation groups. Or in more informal language,  whether wavefronts can straddle different draw calls.

The only text I can find in the Vulkan specification related to this can be found in the [Subgroup chapter](https://registry.khronos.org/vulkan/site/spec/latest/chapters/shaders.html#shaders-scope-subgroup):

> For [shaders that have defined workgroups](https://registry.khronos.org/vulkan/site/spec/latest/chapters/shaders.html#shaders-scope-workgroup), ...
>
> In other shader stages, each invocation in a subgroup must be in the same [device scope instance](https://registry.khronos.org/vulkan/site/spec/latest/chapters/shaders.html#shaders-scope-device).

This implies that drivers have the ability to fill subgroups with invocations not only between draw calls, but between render passes and even queues. While I find it extremely unlikely that they would exercise their power to do this, the spec does not appear to place any limits here from my reading of it.

However, the SPIR-V specification strongly suggests in a few places that the subgroup might be a subset of the invocation group. First, in the [Uniformity chapter](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_uniformity):

> An \<id\> is defined to be [dynamically uniform](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#DynamicallyUniform) for a dynamic instance of an instruction if all invocations (in an invocation group) that execute the dynamic instance have the same value for that <id>.
> 
> While the names may suggest otherwise, nothing forbids an <id> from being decorated as both Uniform and NonUniform. **Because dynamically uniform is at a larger scope (invocation group) than the default Uniform scope (subgroup)**, it is even possible for the <id> to be uniform at the subgroup scope but not dynamically uniform.

"Larger scope", and the rest of the text implies that the subgroup is a subset of the invocation group here.

And viewing the NonUniform decoration with IHV-tinted glasses makes similar hints here. NonUniform is often used in the form of e.g. HLSL's NonUniformResourceIndex, where the idea here is that you're telling a compiler that a resource index might not be uniform across *both* the subgroup and the invocation group. If it does not have this assurance, the compiler might need to generate a loop through all of the resource indexes to perform loads/stores on each resource individually.

If vendors *can* pack draw calls across subgroups, then NonUniform does not quite help those vendors, because even if something is missing the NonUniform hint and is invocation group uniform, it might still not be subgroup uniform and would still need the loop. So, it can be assumed that since the hint was added for *some* reason, that it needs to provide some additional expressive power, e.g. implying that it should imply both non-dynamically uniform and non-subgroup uniform. Was this considered during the development of NonUniform at all?

If it is decided that subgroups *are* a subset of the invocation group, then we should clear up the Vulkan specification to make that extra clear.

If it is decided that subgroups *are not* a subset of the invocation group, then we should clear up the SPIR-V specification to remove the misleading language.

##

While it's not directly related in this case, there are cases of ISVs shipping code relying on subgroups being a subset of either the invocation group, or some further partition of the invocation group. e.g. in [this article](https://gpuopen.com/learn/porting-detroit-3/), AMD and Quantic Dream suggest using readFirstInvocationARB on the InstanceID to scalarize it across the warp.

Even if we decide subgroups are a subset of the invocation group, I still don't think this code would be correct, as I believe multiple instances are still considered part of the same invocation group *. But it still shows a danger here. ISVs rely on wave scalarization tricks to improve performance, and it would be nice to have some assurance about what tricks are valid going forward here.

It might be "obvious" to some that InstanceID is not subgroup uniform and can't be scalarized, but I don't think that's very well documented here. Is DrawID guaranteed to be subgroup uniform? Are uniforms variables guaranteed to be subgroup uniform? Are we willing to make any such statements about anything yet? Regardless of our answers here, I think documentation could be improved as well about which builtins, if any, are subgroup uniform.

## * Footnote on invocation group definition

This is more of an unrelated footnote, but while looking into this, I did find a confusing spec ambiguity in the [Command chapter](https://registry.khronos.org/vulkan/site/spec/latest/chapters/shaders.html#shaders-scope-command) of the specification. The text says:

> For indirect drawing commands with drawCount greater than one, invocations from separate draws are in separate command scope instances.
>
> For shaders without defined [workgroups](https://registry.khronos.org/vulkan/site/spec/latest/chapters/shaders.html#shaders-scope-workgroup), this set of invocations forms an invocation group as defined in the [SPIR-V specification](https://registry.khronos.org/vulkan/site/spec/latest/chapters/introduction.html#spirv-spec).

Carving out indirect multi-draw *only* (and not also, say, direct multi-draw) seemed a bit confusing as a single edge case and can easily be misread, but that's fine. But my question is whether the text "this set of invocations" refers to the invocations in a single command scope instance, or across all the command scope instances generated.

HansKristian-Work:
Fix #2112.

### DIFF
--- a/chapters/shaders.adoc
+++ b/chapters/shaders.adoc
@@ -2502,11 +2502,11 @@ operations>> have defined subgroups.
 [NOTE]
 .Note
 ====
-In shaders, there are two kinds of uniformity that is of interest to
-applications, subgroup uniform and dynamically uniform.
+In shaders, there are two kinds of uniformity that are of interest to
+applications: subgroup uniform and dynamically uniform.
 
 While dynamically uniform appears to be a stronger kind of uniform
-than subgroups, it is not necessarily the case for graphics pipelines.
+than subgroups, it is not necessarily the case for shader stages without defined workgroups.
 
 For shader stages with defined workgroups, this assumption holds.
 If a value is dynamically uniform, it is by definition also uniform across the subgroup,
--- a/chapters/shaders.adoc
+++ b/chapters/shaders.adoc
@@ -2503,15 +2503,16 @@ operations>> have defined subgroups.
 .Note
 ====
 In shaders, there are two kinds of uniformity that are of primary interest to
-applications:, subgroup uniform and dynamically uniform.
+applications: uniform within an invocation group (a.k.a. dynamically uniform),
+and uniform within a subgroup scope.
 
-While one could make the assumption that being dynamically uniform implies being subgroup uniform,
+While one could make the assumption that being uniform in invocation group implies being uniform in subgroup scope,
 it is not necessarily the case for shader stages without defined workgroups.
 
 For shader stages with defined workgroups however,
-the relationship between dynamically uniform and subgroup uniform is well defined.
-If a value is dynamically uniform, it is by definition also uniform across the subgroup,
-as this is a specific guarantee provided to stages with explicit workgroups.
+the relationship between invocation group and subgroup scope is well defined as a subgroup is a subset of the workgroup, and
+the workgroup is the invocation group.
+If a value is uniform in invocation group, it is by definition also uniform in subgroup scope.
 This is important if writing code like:
 
 [source,c]
@@ -2520,17 +2521,17 @@ uniform texture2D Textures[];
 uint dynamicallyUniformValue = gl_WorkGroupID.x;
 vec4 value = texelFetch(Textures[dynamicallyUniformValue], coord, 0);
 
-// subgroupUniformValue is guaranteed to be uniform across the subgroup.
+// subgroupUniformValue is guaranteed to be uniform within the subgroup.
 // This value also happens to be dynamically uniform.
 vec4 subgroupUniformValue = subgroupBroadcastFirst(dynamicallyUniformValue);
 ~~~~
 
 In shader stages without defined workgroups, this gets complicated.
-Due to scoping rules, there is no guarantee that a subgroup is a subset of the invocation scope,
+Due to scoping rules, there is no guarantee that a subgroup is a subset of the invocation group,
 which in turn defines the scope for dynamically uniform.
-In graphics, the invocation scope is a single draw command, except for
+In graphics, the invocation group is a single draw command, except for
 multi-draw situations, and indirect draws with drawCount > 1,
-where there are multiple invocation scopes, one per code:DrawIndex.
+where there are multiple invocation groups, one per code:DrawIndex.
 
 [source,c]
 ~~~~
@@ -2552,13 +2553,13 @@ uint notActuallyDynamicallyUniformAnymore =
 //              | gl_DrawID = 2 | gl_DrawID = 1 | }
 // Subgroup 1: { 2, 2, 2, 2,      2, 2, 2, 2 }
 
-// Bug. gl_DrawID = 1's invocation scope observes both index 0 and 2.
+// Bug. gl_DrawID = 1's invocation group observes both index 0 and 2.
 vec4 value = texelFetch(Textures[notActuallyDynamicallyUniformAnymore],
                         coord, 0);
 ~~~~
 
 Another problematic scenario is when a shader attempts to help the compiler notice
-that a value is subgroup uniform to potentially improve performance.
+that a value is uniform in subgroup scope to potentially improve performance.
 
 [source,c]
 ~~~~
--- a/chapters/shaders.adoc
+++ b/chapters/shaders.adoc
@@ -2578,7 +2578,7 @@ For implementations where subgroups are packed across draws, the implementation 
 make sure to handle descriptor indexing correctly. From the specification's point of view,
 a dynamically uniform index does not require code:NonUniform decoration,
 and such an implementation will likely either promote descriptor indexing into code:NonUniform on its own,
-or handle non-uniformity natively.
+or handle non-uniformity implicitly.
 ====
 endif::VK_VERSION_1_1[]
 
--- a/chapters/shaders.adoc
+++ b/chapters/shaders.adoc
@@ -2502,7 +2502,7 @@ operations>> have defined subgroups.
 [NOTE]
 .Note
 ====
-In shaders, there are two kinds of uniformity that are of interest to
+In shaders, there are two kinds of uniformity that are of primary interest to
 applications: subgroup uniform and dynamically uniform.
 
 While dynamically uniform appears to be a stronger kind of uniform
@@ -2524,7 +2524,7 @@ vec4 value = texelFetch(Textures[dynamicallyUniformValue], coord, 0);
 vec4 subgroupUniformValue = subgroupBroadcastFirst(dynamicallyUniformValue);
 ~~~~
 
-In fragment shaders and other graphics stages, this gets complicated.
+In shader stages without defined workgroups, this gets complicated.
 Due to scoping rules, there is no guarantee that a subgroup is a subset of the invocation scope,
 which in turn defines the scope for dynamically uniform.
 In graphics, the invocation scope is a single draw command.
--- a/chapters/shaders.adoc
+++ b/chapters/shaders.adoc
@@ -2503,12 +2503,13 @@ operations>> have defined subgroups.
 .Note
 ====
 In shaders, there are two kinds of uniformity that are of primary interest to
-applications: subgroup uniform and dynamically uniform.
+applications:, subgroup uniform and dynamically uniform.
 
-While dynamically uniform appears to imply subgroup uniform,
+While one could make the assumption that being dynamically uniform implies being subgroup uniform,
 it is not necessarily the case for shader stages without defined workgroups.
 
-For shader stages with defined workgroups, this assumption holds.
+For shader stages with defined workgroups however,
+the relationship between dynamically uniform and subgroup uniform is well defined.
 If a value is dynamically uniform, it is by definition also uniform across the subgroup,
 as this is a specific guarantee provided to stages with explicit workgroups.
 This is important if writing code like:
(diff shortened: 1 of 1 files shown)


chapters/shaders.adoc
```diff
@@ -2498,6 +2498,86 @@ In other shader stages, each invocation in a subgroup must: be in the same
 
 Only <<limits-subgroup-supportedStages, shader stages that support subgroup
 operations>> have defined subgroups.
+
+[NOTE]
+.Note
+====
+In shaders, there are two kinds of uniformity that is of interest to
+applications, subgroup uniform and dynamically uniform.
```
nanokatze, line 2506:
```suggestion
In shaders, there are two kinds of uniformity that are of interest to
applications: subgroup uniform and dynamically uniform.
```

chapters/shaders.adoc
```diff
@@ -2498,6 +2498,86 @@ In other shader stages, each invocation in a subgroup must: be in the same
...
+
+While dynamically uniform appears to be a stronger kind of uniform
+than subgroups, it is not necessarily the case for graphics pipelines.
```
nanokatze, line 2509: We have graphics stages with workgroups (TS, MS, tessellation stuff), and non-graphics stages without workgroups (RT)

```suggestion
While dynamically uniform appears to be a stronger kind of uniform
than subgroups, it is not necessarily the case for shader stages without defined workgroups.
```

chapters/shaders.adoc
```diff
@@ -2498,6 +2498,86 @@ In other shader stages, each invocation in a subgroup must: be in the same
...
+uniform texture2D Textures[];
+uint dynamicallyUniformValue = gl_WorkGroupID.x;
+vec4 value = texelFetch(Textures[dynamicallyUniformValue], coord, 0);
+
+// subgroupUniformValue is guaranteed to be uniform across the subgroup.
+// This value also happens to be dynamically uniform.
+vec4 subgroupUniformValue = subgroupBroadcastFirst(dynamicallyUniformValue);
+~~~~
+
+In fragment shaders and other graphics stages, this gets complicated.
```
nanokatze, line 2527: Other graphics stages is once again not correct. I'm not sure what a precise and good wording would be. Perhaps enumerate all stages? This should also include RT

nanokatze (changes requested):

martty:
Would it be possible to instead "forbid" the repacking behavior in graphics? In essence only allow implementations to pack subgroups if the shader cannot observe the difference - I think this could be a more intuitive solution.
If the perf loss is large then an extension could relax this requirement.

chapters/shaders.adoc
oddhack: "stronger" seems poorly defined in this context - is there a more specific meaning? It sounds like "better".

chapters/shaders.adoc
oddhack: Are there other types that are *not* of interest to apps?

chapters/shaders.adoc
nanokatze: > "stronger" seems poorly defined in this context - is there a more specific meaning? It sounds like "better".

"Stronger" here appears to mean "implies", so perhaps something along the lines of "While dynamically uniform appears to imply subgroup uniform" would be better?

oddhack: I was still hoping to improve the "stronger" term - something like @nanokatze suggests would be better.

chapters/shaders.adoc
HansKristian-Work: Reworded.

chapters/shaders.adoc
```diff
@@ -2498,6 +2498,86 @@ In other shader stages, each invocation in a subgroup must: be in the same
...
 Only <<limits-subgroup-supportedStages, shader stages that support subgroup
 operations>> have defined subgroups.
+
+[NOTE]
+.Note
+====
+In shaders, there are two kinds of uniformity that are of primary interest to
+applications: subgroup uniform and dynamically uniform.
+
+While dynamically uniform appears to imply subgroup uniform,
```
janharaldfredriksen-arm, line 2508: nit: "appears to" reads as if the spec implies this. Maybe something like: "A common assumption is that dynamically uniform implies .." or "One could easily make the assumption that..".

chapters/shaders.adoc
```diff
@@ -2498,6 +2498,86 @@ In other shader stages, each invocation in a subgroup must: be in the same
...
+it is not necessarily the case for shader stages without defined workgroups.
+
+For shader stages with defined workgroups, this assumption holds.
```
janharaldfredriksen-arm, line 2511: nit: "this assumption" -> I understand what this refers to, but there's strictly speaking no "assumption" stated above. (My suggestion above would make the assumption explicit.)

chapters/shaders.adoc
```diff
@@ -2498,6 +2498,86 @@ In other shader stages, each invocation in a subgroup must: be in the same
...
+
+// subgroupUniformValue is guaranteed to be uniform across the subgroup.
+// This value also happens to be dynamically uniform.
+vec4 subgroupUniformValue = subgroupBroadcastFirst(dynamicallyUniformValue);
+~~~~
+
+In shader stages without defined workgroups, this gets complicated.
+Due to scoping rules, there is no guarantee that a subgroup is a subset of the invocation scope,
+which in turn defines the scope for dynamically uniform.
+In graphics, the invocation scope is a single draw command.
```
janharaldfredriksen-arm, line 2530: "single draw." ?
(The sentence below is somewhat contradictory otherwise sine vkCmdDrawIndexedIndirect (with drawCount > 1) is technically also a single draw _command_.)

chapters/shaders.adoc
```diff
@@ -2498,6 +2498,86 @@ In other shader stages, each invocation in a subgroup must: be in the same
...
+uniform texture2D Textures[];
+
+// DrawIndex builtin is dynamically uniform
+uint dynamicallyUniformValue = gl_DrawID;
+//              | gl_DrawID = 0 | gl_DrawID = 1 | }
+// Subgroup 0: { 0, 0, 0, 0,      1, 1, 1, 1 }
+//              | DrawID = 2 | DrawID = 1 | }
+// Subgroup 1: { 2, 2, 2, 2,      1, 1, 1, 1 }
+
+uint notActuallyDynamicallyUniformAnymore =
```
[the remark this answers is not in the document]
HansKristian-Work: > t's concerning because this part suggests that you can't use a peeling loop to remove the need for the nonuniform decoration, which I think existing content relies on.

Peeling loops to avoid nonuniformEXT are technically out of spec, since dynamically uniform is the requirement, not subgroup uniform.

chapters/shaders.adoc
HansKristian-Work: Reworded.

chapters/shaders.adoc
HansKristian-Work: Reworded.

chapters/shaders.adoc
```diff
@@ -2498,6 +2498,88 @@ In other shader stages, each invocation in a subgroup must: be in the same
...
+// Hazard: If different draw commands are packed into one subgroup, the uniformIndex is wrong.
+
+DrawData d = UBO.perDrawData[uniformIndex];
+~~~~
+
+For implementations where subgroups are packed across draws, the implementation must
+make sure to handle descriptor indexing correctly. From the specification's point of view,
+a dynamically uniform index does not require code:NonUniform decoration,
+and such an implementation will likely either promote descriptor indexing into code:NonUniform on its own,
+or handle non-uniformity natively.
```
samhsw, line 2581: This point makes a lot of senses to me, however I am hassled between using the word "natively" and "implicitly". In this specific case the implementation will be expected to handle this "implicit" non-uniformity correctly by itself without an "explicit" NonUnifom decoration on the dynamically uniform index. It seems to me that it saying the implementation will be responsible to make the judgement for individual situations before taking the actions correspondingly, but not natively for all the general cases. I may go for a "implicitly" as a nitpicker.

samhsw (approved):

chapters/shaders.adoc
samhsw, line 2581: This point makes a lot of senses to me, however I am hassled between using the word "natively" and "implicitly". In this specific case the implementation will be expected to handle this "implicit" non-uniformity correctly by itself without an "explicit" NonUnifom decoration on the dynamically uniform index. It seems to me that it saying the implementation will be responsible to make the judgement for individual situations before taking the actions correspondingly, but not natively for all the general cases. I may go for a "implicitly" as a nitpicker.

samhsw: Ah, I was meant to click the review with comment.

HansKristian-Work: Reworded to "implicitly" rather than "native".

chapters/shaders.adoc
```diff
@@ -2498,6 +2498,88 @@ In other shader stages, each invocation in a subgroup must: be in the same
 
 Only <<limits-subgroup-supportedStages, shader stages that support subgroup
 operations>> have defined subgroups.
+
+[NOTE]
+.Note
+====
+In shaders, there are two kinds of uniformity that are of primary interest to
+applications:, subgroup uniform and dynamically uniform.
```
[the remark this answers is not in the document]
Tobski: Whoops, edit: we _do_ still use dynamically uniform, but not subgroup uniform.  We also say "dynamically uniform within a <SCOPE> scope instance", so my point mostly still stands, with some caveats 😅.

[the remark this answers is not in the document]
HansKristian-Work: Fixed.

PR #2121, opened 2023-05-08
Add VK_MSFT_layered_driver

jenatali:
This is my first attempt at a Vulkan extension. I chose `MSFT` for the vendor bits, but perhaps `EXT` would be more appropriate if MoltenVK has interest? Not sure if I'm supposed to also go ahead and update XML as part of this proposal or if that would come later.

This is needed to address https://github.com/KhronosGroup/Vulkan-Loader/issues/981. Without this, or something like it, VulkanOn12's physical devices can't be enumerated by LUID, or else they run the risk of being enumerated before a native driver for the same GPU.

### DIFF
--- a/proposals/VK_MSFT_layered_driver.adoc
+++ b/proposals/VK_MSFT_layered_driver.adoc
@@ -1,0 +1,87 @@
+// Copyright 2021-2023 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+= Proposal Template
+:toc: left
+:refpage: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/
+:sectnums:
+
+This document proposes an extension to allow the loader to understand
+driver layering, for improving physical device sorting.
+
+== Problem Statement
+
+The Vulkan loader is able to sort physical devices according to
+platform-specific criteria. For example, on Windows, the loader uses LUIDs
+to put physical devices in the same order as DXGI adapters. However, it is
+possible to have multiple Vulkan drivers that provide support for the same
+physical device, for example, where one is a "native" vendor-provided
+implementation and another is a "layered" implementation on top of a
+different API. Examples of layered implementations would include VulkanOn12
+(aka Dozen), layered on D3D12, and MoltenVK, layered on Metal.
+
+On a system where a physical device has two possible drivers, the sort
+order between them is currently unspecified. An ideal sort order
+should place any native/un-layered drivers sorted-before any layered
+drivers, as it should be expected that native drivers will provide more
+functionality and higher performance, since layering inherently adds
+overhead. But the loader has no way of knowing which driver to prefer.
+
+An additional problem that is not addressed by this specification is the
+case where you have multiple "native" drivers for a single physical device.
+In that case, the sort order remains unspecified, as a correct ordering
+between drivers is non-obvious.
+
+== Solution Space
+
+Options that were considered include:
+* Special-casing well-known layered drivers in the Vulkan loader.
+* Extending the Loader-ICD interface to identify layered drivers.
+* Providing an extension to allow layered drivers to self-identify.
+
+The latter solution is the more general, and also has the benefit of
+allowing applications to understand when they're running on a layered
+driver.
+
+== Proposal
+
+The following properties are exposed by the `VK_MSFT_layered_driver`
+extension:
+[source,c]
+----
+typedef enum VkLayeredDriverUnderlyingApiMSFT {
+    VK_LAYERED_DRIVER_UNDERLYING_API_NONE_MSFT,
+    VK_LAYERED_DRIVER_UNDERLYING_API_D3D12_MSFT,
+} VkLayeredDriverUnderlyingApiMSFT;
+
+typedef struct VkPhysicalDeviceLayeredDriverPropertiesMSFT {
+    VkStructureType sType;
+    const void* pNext;
+    VkLayeredDriverUnderlyingApiMSFT underlyingAPI;
+} VkPhysicalDeviceLayeredDriverPropertiesMSFT;
+----
+
+Layered drivers should implement this extension. The Vulkan loader can then
+be updated to query for the this structure. If the `underlyingAPI` is not
+`NONE`, then the driver should be considered layered. The specific value of
+`underlyingAPI` is simply informational for applications to query if they
+so choose.
+
+== Issues
+
+=== RESOLVED: Is a string the right way to identify an underlying API?
+
+No, an enum is a much better solution. The same conclusion was already
+reached with the `VkDriverId` enum.
+
+== Further Functionality
+
+Additional properties of the layering implementation, such as underlying
+API object pointers, could be exposed, but considering that the nature of
+those will depend on the underlying API, it seems like those should be
+exposed via separate extensions, if at all.
+
+It might make sense to add things like driver version for the underlying
+driver, since the version information exposed through existing properties
+would refer to the version of layered implementation.
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -781,6 +781,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
         <type name="VkComponentTypeKHR" category="enum"/>
         <type category="enum" name="VkScopeNV"                                     alias="VkScopeKHR"/>
         <type category="enum" name="VkComponentTypeNV"                             alias="VkComponentTypeKHR"/>
+        <type name="VkLayeredDriverUnderlyingApiMSFT" category="enum"/>
 
             <comment>WSI extensions</comment>
         <type name="VkColorSpaceKHR" category="enum"/>
@@ -8388,6 +8389,11 @@ typedef void* <name>MTLSharedEvent_id</name>;
             <member optional="true"><type>void</type>*                               <name>pNext</name></member>
             <member limittype="bitmask"><type>VkShaderStageFlags</type>              <name>cooperativeMatrixSupportedStages</name></member>
         </type>
+        <type category="struct" name="VkPhysicalDeviceLayeredDriverPropertiesMSFT" structextends="VkPhysicalDeviceProperties2" returnedonly="true">
+            <member values="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_LAYERED_DRIVER_PROPERTIES_MSFT"><type>VkStructureType</type> <name>sType</name></member>
+            <member optional="true"><type>void</type>*                       <name>pNext</name></member>
+            <member><type>VkLayeredDriverUnderlyingApiMSFT</type>            <name>underlyingAPI</name></member>
+        </type>
     </types>
 
 
@@ -10441,6 +10447,10 @@ typedef void* <name>MTLSharedEvent_id</name>;
         <enum value="8"     name="VK_COMPONENT_TYPE_UINT16_KHR"/>
         <enum value="9"     name="VK_COMPONENT_TYPE_UINT32_KHR"/>
         <enum value="10"    name="VK_COMPONENT_TYPE_UINT64_KHR"/>
+    </enums>
+    <enums name="VkLayeredDriverUnderlyingApiMSFT" type="enum">
+        <enum value="0" name="VK_LAYERED_DRIVER_UNDERLYING_API_NONE_MSFT"/>
+        <enum value="1" name="VK_LAYERED_DRIVER_UNDERLYING_API_D3D12_MSFT"/>
     </enums>
 
     <commands comment="Vulkan command definitions">
@@ -22521,10 +22531,13 @@ typedef void* <name>MTLSharedEvent_id</name>;
                 <command name="vkGetScreenBufferPropertiesQNX"/>
             </require>
         </extension>
-        <extension name="VK_MSFT_extension_531" number="531" author="MSFT" contact="Jesse Natalie @jenatali" supported="disabled">
+        <extension name="VK_MSFT_layered_driver" number="531" type="device" depends="VK_KHR_get_physical_device_properties2" author="MSFT" contact="Jesse Natalie @jenatali" supported="vulkan">
             <require>
-                <enum value="0"                                         name="VK_MSFT_EXTENSION_531_SPEC_VERSION"/>
-                <enum value="&quot;VK_MSFT_extension_531&quot;"         name="VK_MSFT_EXTENSION_531_EXTENSION_NAME"/>
+                <enum value="1"                                         name="VK_MSFT_LAYERED_DRIVER_SPEC_VERSION"/>
+                <enum value="&quot;VK_MSFT_layered_driver&quot;"        name="VK_MSFT_LAYERED_DRIVER_EXTENSION_NAME"/>
+                <enum offset="0" extends="VkStructureType"              name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_LAYERED_DRIVER_PROPERTIES_MSFT"/>
+                <type name="VkLayeredDriverUnderlyingApiMSFT"/>
+                <type name="VkPhysicalDeviceLayeredDriverPropertiesMSFT"/>
             </require>
         </extension>
         <extension name="VK_KHR_extension_532" number="532" author="KHR" contact="Tobias Hector @tobias" supported="disabled">
--- a/appendices/VK_MSFT_layered_driver.adoc
+++ b/appendices/VK_MSFT_layered_driver.adoc
@@ -1,0 +1,31 @@
+// Copyright 2023 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_MSFT_layered_driver.adoc[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2023-06-21
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Jesse Natalie, Microsoft
+
+=== Description
+
+This extension adds new physical device properties to allow applications
+and the Vulkan ICD loader to understand when a physical device is
+implemented as a layered driver on top of another underlying API.
+
+include::{generated}/interfaces/VK_MSFT_layered_driver.adoc[]
+
+=== Examples
+
+None.
+
+=== Version History
+
+  * Revision 1, 2023-06-21 (Jesse Natalie)
+  ** Initial revision
--- a/chapters/devsandqueues.adoc
+++ b/chapters/devsandqueues.adoc
@@ -1135,6 +1135,36 @@ include::{generated}/validity/structs/VkPhysicalDeviceShaderTileImagePropertiesE
 --
 endif::VK_EXT_shader_tile_image[]
 
+ifdef::VK_MSFT_layered_driver[]
+[open,refpage='VkPhysicalDeviceLayeredDriverPropertiesMSFT',desc='Structure containing information about driver layering for a physical device',type='structs']
+--
+The sname:VkPhysicalDeviceLayeredDriverPropertiesMSFT structure is defined
+as:
+
+include::{generated}/api/structs/VkPhysicalDeviceLayeredDriverPropertiesMSFT.adoc[]
+
+  * pname:sType is the type of this structure.
+  * pname:pNext is `NULL` or a pointer to a structure extending this
+    structure.
+  * pname:underlyingAPI is a elink:VkLayeredDriverUnderlyingApiMSFT value
+    indicating which underlying API is used to implement the layered driver,
+    or ename:VK_LAYERED_DRIVER_UNDERLYING_API_NONE_MSFT if the driver is not
+    layered.
+
+These are properties of the driver layering information of a physical device.
+
+include::{generated}/validity/structs/VkPhysicalDeviceLayeredDriverPropertiesMSFT.adoc[]
+--
+
+[open,refpage='VkLayeredDriverUnderlyingApiMSFT',desc='Layered driver underlying APIs',type='enums']
+--
+Underlying APIs which may: be returned in
+slink:VkPhysicalDeviceLayeredDriverPropertiesMSFT::pname:underlyingAPI are:
+
+include::{generated}/api/enums/VkLayeredDriverUnderlyingApiMSFT.adoc[]
+--
+endif::VK_MSFT_layered_driver[]
+
 [open,refpage='vkGetPhysicalDeviceQueueFamilyProperties',desc='Reports properties of the queues of the specified physical device',type='protos']
 --
 To query properties of queues available on a physical device, call:


oddhack:
Generally we defer the XML / spec work until the proposal is agreed on by all the stakeholders - apparently that includes Microsoft and perhaps MoltenVK, but would this touch on LunarG and the SDK work they do as well? We are having the F2F meeting this week and I'll try to make sure  this gets brought up to people's attention.

jenatali:
> would this touch on LunarG and the SDK work they do as well?

I had planned to submit the changes to the `Vulkan-Loader` repo to make use of this extension once there was consensus that it was a good idea. That *should* be the only impact to LunarG as far as I'm aware.

oddhack:
> > would this touch on LunarG and the SDK work they do as well?
> 
> I had planned to submit the changes to the `Vulkan-Loader` repo to make use of this extension once there was consensus that it was a good idea. That _should_ be the only impact to LunarG as far as I'm aware.

Thanks. Is there anyone else in particular that should feed into consensus? I put this on the F2F agenda for this evening (Osaka physical meeting) to raise awareness, but not sure who in particular you would like input from.

jenatali:
Good question. I'm honestly not sure. I/we are new at this so I'm not really sure what to expect or who needs to be aware of it.

versalinyaa:
jenatali, thanks for the proposal. We just now briefly discussed it at the Khronos F2F. I will put it on the agenda for a Khronos-internal meeting on Wed 24 May, where we can get realtime feedback from Vulkan loader dev.

But please also continue the discussion here in GitHub. The public and internal discussions can happen in parallel, and I will cross-post updates as needed.

versalinyaa:
@charles-lunarg Do you have any feedback for this proposal? It will also be on the SI agenda for Wed 24 May.

pdaniell-nv:
> @charles-lunarg Do you have any feedback for this proposal? It will also be on the SI agenda for Wed 24 May.

It looks like @charles-lunarg already discussed this in https://github.com/KhronosGroup/Vulkan-Loader/issues/981 and recommended doing this extension. FWIW I have no objection to this extension and think it's the right solution.

charles-lunarg:
I confirm, this extension matches the discussion. Happy to put it up to further discussion/review in the SI meeting.

oddhack:
@jenatali since this is your first extension, be aware you need to reserve an extension number in main branch before creating the PR with the actual extension XML / spec language. Something like #2127 but using the next free extension number at the time you propose the PR, and the MSFT (or EXT) author ID. This is so enum values remain stable during development of the extension, since each extension gets its own numeric range to play in.

jenatali:
@oddhack Thanks for pointing that out. I've created #2129.

krOoze:
If we suspect those drivers are inferior, perhaps they should not be exposed by default to the users. Add something like a `ENUMERATE_REDUNDANT_DRIVERS` flag similar to portability flag?

On point of nomenclature, is "layered" the term we want to go with? It is somewhat vague, and also has overloaded meaning with Vulkan Layers.

jenatali:
> If we suspect those drivers are redundant and inferior in every way, perhaps they should not be exposed by default to the users.

As it stands today, I don't expect a layered driver to be able to match a native driver for performance or functionality, though that may change in the future (e.g. I've heard that OpenGLOn12 can surpass some vendor's native GL drivers for performance, and D3D11On12 can surpass native D3D11 drivers for some hardware). That said, there's 2 cases where it's interesting to expose by default:
* Devices where there is no native driver. Windows ARM devices do not have a Qualcomm Vulkan driver, and the D3D software rasterizer can also benefit from exposing Vulkan support without having a full native Vulkan frontend.
* There may be a select subset of extensions or functionality that could be exposed from a layered driver ahead of a native implementation, or interop scenarios where choosing a layered implementation makes sense.

All that said, the specific loader policy is a bit outside the scope of this extension definition. The purpose here is to enable the loader to make *any* kind of choice for which drivers to expose for a given device and how to sort them.

> On point of nomenclature, is "layered" the term we want to go with? It is somewhat vague, and also has overloaded meaning with Vulkan Layers.

I'm building on top of the nomenclature that Khronos has used in the past in official documentation. For example, in the [Khronos conformance process](https://www.khronos.org/files/conformance_procedures.pdf) documents, section A9 part (ii) has specific instructions for "layered implementations." That said, I have no specific attachment to the term.

krOoze:
> As it stands today, I don't expect a layered driver to be able to match a native driver for performance or functionality, though that may change in the future

Ok, but in that case, what does the `layered` bool even tells the application. Makes the whole information irrelevant to anything.

> Devices where there is no native driver. 

Certainly, I meant where there are duplicate drivers, loader could eliminate the inferior ones (non-conformant\portability, and "layered" ones).

> There may be a select subset of extensions or functionality that could be exposed from a layered driver

That sounds like something useful to a developer, and not an end user, and as such should be protected away behind a flag?

> All that said, the specific loader policy is a bit outside the scope of this extension definition. 

I try to think how it fits in the overal picture. The question should always be "why do we need to frankenstein this onto the Vulkan API".

This proposes an extension for a certain **situation**. Should we contemplate the extension, without first contemplating whether we even should be getting ourself into that **situation**? Before we contemplate sorting, **should there even be duplicate inferior drivers included there to be sorted in the first place**?

 I mean what does an end user even get out of this, except more potential for things to explode? I mean, old apps won't even be aware of existence of this proposed extension. And as you seem to imply, the information that the driver is "layered"  can mean anything, and so doesn't tell the app much actionable information in the first place that can be relied upon for anything.

jenatali:
Just to be clear, the goal of this extension is for the implementation to provide information to the *loader* about what kind of driver it is. Providing this information to applications is much, much less interesting. If the loader then uses that info to hide the devices reported by that driver when there's already a (non-layered/native) driver that enumerated the same physical device, that sounds fine. Personally I'd like a way to opt out of that, but I'm fine with an environment variable.

If there's an alternative mechanism to provide this information to the loader beyond extending the information reported from the driver, I'm all ears, but this was the suggested mechanism from the loader maintainers.

krOoze:
>  the goal of this extension is for the implementation to provide information to the loader

Interesting. That would imply extending [https://github.com/KhronosGroup/Vulkan-Loader/blob/main/docs/LoaderDriverInterface.md](Loader-ICD interface), rather than Vulkan API, no?

jenatali:
An extension seemed like the right fit because then the information *can* be queried by applications, if they'd have a use for it, similar to the portability subset, but yes a viable alternative would be to revise the Loader-ICD interface.

I strongly disagree about the loader giving information to the driver to put policy about what should be enumerated in the driver. The loader should be the arbitrator of policy.

krOoze:
IDK, Loader is the one component that has least amount of information and context to decide anything. Its purpose is to mediate between Apps, Drivers, and Layers, nothing more. This sounds like something two drivers want to talk through between themselves. It feels the Emulator should be the one to decide whether competing with a native driver is its core mission, or not.

r-potter:
IMO this extension is the right way to do this. Applications are the ultimate arbiter of what implementations they want to run on top of. All this extension is doing is:

1.) Providing the loader enough information to make a best effort at a sensible default ordering
2.) Providing applications some information to distinguish between two implementations more explicitly than trying to parse driver name strings

Existing examples of API layering, like ANGLE, suggest that describing the layered implementation as "inferior" is likely to be overly simplistic. It's entirely likely that a layered implementation will outperform a native driver in some situations, and underperform it in others. It's also entirely plausible for a layered implementation to end up exposing features or extensions that the native driver didn't, which may also make the layered implementation the preferred choice in some situations. 

Sorting native drivers first is the best default behaviour we can expect to come up with. If an application developer discovers through benchmarking that actually the layered implementation is better for their specific application, then with this extension they still get the option to it make use of that. If we start hiding (layered) implementations then we we lose that option.

krOoze:
@r-potter  I am not so sure. Drivers want to offload responsibilities to the Loader (which knows nothing) and exports it to the App which will do nothing else than try and fail on flimsy information, and in the end export the responsibilities to the unsuspecting end-users.

1) Best effort is not good enough. Separation of concerns so each party only does what he is most competent at is the way. The Emulator driver is the one who knows what it wants to achieve. E.g. MoltenVK driver wants to achieve bringing Vulkan to Apple stuff. If a native driver shows up, that precondition is invalidated. We can either communicate ever increasing information to the Loader on all the nuanced way whether device should show up, and at what position. OR, we can let driver decide on things that increasingly start looking more and more like driver responsibility to me.

2) Old apps would not have access to those informations, therefore might explode. Additionally, not really being apps concern in the first place, it does not have much to decide on (as you said, describing the X implementation as "inferior" based on Y flag is likely to be overly simplistic, and apps can only do deterministic choices, not some kind of "moral judgements"). It is the "layered implementation" choice\mission whether it wants to outperform a native driver, or whether its mission is to offer a driver where there was none before. What seems to be the case, we want to strip out this context the Driver knows about itself, and dump this responsibility on someone else down the road.

charles-lunarg:
@krOoze What are you suggesting we do then?

My view of the problem is that "layered implementations exist, but there may be native drivers available. The loader's response to this situation is to simply order the list of VkPhysicalDevices such that the native one is ahead of the layered one." Do you intent to suggest an alternate solution to 'there may exist drivers that are layered ontop of native ones while native drivers are also present'?

krOoze:
@charles-lunarg Think, measure, and only then act.

MarkY-LunarG:
I personally think that an enum for various APIs would be better than a string.  If we have strings, there may be different variations for the same underlying API if the Vulkan layer support comes from different developers.

nanokatze:
Has a case of two "native" drivers been considered/in scope for this ext? An example of this could be RADV and AMDVLK

versalinyaa:
> I personally think that an enum for various APIs would be better than a string. If we have strings, there may be different variations for the same underlying API if the Vulkan layer support comes from different developers.

I agree. We already have a similar enum, `VkDriverId`, that (a) eliminates the inherent variance of vendor-provided strings and (b) is a sloppy ad-hoc list that anyone is permitted to add to. If this extension also uses an enum, I expect it to be managed similarly to `VkDriverId`.

jenatali:
> Has a case of two "native" drivers been considered/in scope for this ext? An example of this could be RADV and AMDVLK

Already called out as not in scope:

> An additional problem that is not addressed by this specification is the case where you have multiple "native" drivers for a single physical device. In that case, the sort order remains unspecified, as a correct ordering between drivers is non-obvious.

Thanks for the discussion everyone. I've updated the draft to switch to using an enum instead of a string, and to capture a couple of the points raised here.

krOoze:
- The enum kinda obviates the bool. One of the enumerants could be `VK_LAYERED_DRIVER_UNDERLYING_API_NONE_MSFT?
- Not sure I like `VK_LAYERED_DRIVER_UNDERLYING_API_UNKNOWN_MSFT`. The emulator should know the underlying API, and should be able to make the enumerant specified by the point it is reaching Vulkan conformance (and until it does it shouldn't matter).
- It's _only_ a proposal (hence many APIs missing there), but it is somewhat inconsistent the DX is versioned, and the Metal is not.
    - If we are trying to maintain this information is useful\actionable, it should be versioned. E.g. OpenGL 2 would be more yikes than OpenGL 4. But that also leaves out the question open what should be the granularity of those version enumerants, or if it needs outright a separate mechanism to query the capabilities of the underlying API.
        - When I think about it, all of this is basically a very imperfect proxy for determining which features are emulated (i.e. not performant as much as they should be). It also ignores the purpose of the layered implementation. It may be to compete with (or provide the only) Vulkan driver, but it may be for developers to emulate as many features while not caring about performance. The `isLayered` bool simply does not tell the whole story by omitting such important facts, and in so doing might be misleading and unactionable in of itself.

jenatali:
Agreed about `isLayered` being not useful with an enum that could simply have a `NONE` entry. I've pushed that change, and scoped the enum to just `NONE` and the only API that I have any personal interest in at the moment. I don't have enough knowledge about Metal or any other layered implementations (if there are any) to know what the right granularity is for versioning there.

> Offering multiple same physical devices feels like a soft API break.

Yes, I agree. Personally, I feel the CL API did this better, where enumeration has 2 levels, one for a platform, i.e. a driver, and then from the platform you can query for devices. Having the same device enumerated by 2 platforms makes more sense than having the same device enumerated twice in the same flat list. I mentioned this in a one-off Khronos meeting in July 2020, which at the time was largely focused around loader behaviors for CLOn12, and Vulkan experts suggested that the same GPU would be exposed as 2 physical devices and there's not really another way around that.

> IMO they should perhaps be unique, and alternatives only offered through pNext or something for the rare case it is sensible for an app to override this.

I'd be okay with a `pNext` style extension. That still seems like something that should be implemented by the loader, which can de-duplicate physical devices based on unique identifiers (PCI paths on Linux, LUIDs on Windows), select a primary one, and relegate others to `pNext`. That sounds like a separate extension though, one which would still depend on this one to help inform which implementation should be the primary one for a given adapter.

> I need to reiterate this whole proposal hinges on the premise that physical devices are ordered, which they are not per specification.

The Loader-ICD interface version 6 (for Windows) was explicitly introduced to begin ordering physical devices. If ordering wasn't intended, then why was this done? Perhaps the fact that the loader does ordering is outside of the core Vulkan specification and only a "quirk" of the official loader, but Windows in general has a well-defined sort order for GPUs, which can be influenced by user interaction with the OS or driver control panels.

>  therefore there should be a mechanism present for them to self-eliminate themselves at runtime

My personal opinion here is that this kind of policy belongs in the hands of a component that has full knowledge of the state of the system and ecosystem, i.e. the Vulkan loader, rather than in an individual driver. In the case of RADV/AMDVLK, would you expect one of them to hide themselves? Which one? What happens if both of them try to hide themselves?

proposals/VK_MSFT_layered_driver.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+=== RESOLVED: Is a string the right way to identify an underlying API?
+
+No, an enum is a much better solution. The same conclusion was already reached with the `VkDriverId` enum.
+
+== Further Functionality
+
+Additional properties of the layering implementation, such as underlying API object pointers, could be exposed, but considering that the nature of those will depend on the underlying API, it seems like those should be exposed via separate extensions, if at all.
+
+It might make sense to add things like driver version for the underlying driver, since the version information exposed through existing properties would refer to the version of layered implementation.
```
versalinyaa, line 58: I agree, it makes sense to expose properties of the underlying driver, such as version and vendor. But the set of the significant underlying driver info would probably differ for different underlying apis, and that info is likely to be structured differently too. Therefore, I think the exposure of underlying driver details should be delegated to per-api extensions (D3D12, Metal, etc).

proposals/VK_MSFT_layered_driver.adoc
```diff
@@ -0,0 +1,58 @@
...
+typedef enum VkLayeredDriverUnderlyingApiMSFT {
+    VK_LAYERED_DRIVER_UNDERLYING_API_NONE_MSFT,
+    VK_LAYERED_DRIVER_UNDERLYING_API_D3D12_MSFT,
+} VkLayeredDriverUnderlyingApiMSFT;
+
+typedef struct VkPhysicalDeviceLayeredDriverPropertiesMSFT {
+    VkStructureType sType;
+    const void* pNext;
+    VkLayeredDriverUnderlyingApiMSFT underlyingAPI;
+} VkPhysicalDeviceLayeredDriverPropertiesMSFT;
```
versalinyaa, line 43: Tiny nitpick. The field members must be aligned. It's the required style everywhere in Vulkan, even in proposal docs. Sorry, I dislike it myself, it's annoying.

oddhack: I'm not sure why you say that? The XML -> C header generator script does do this, but it's not mandated in the style guide, or relevant to proposals documents, unless I've forgotten something.

jenatali:
Hi folks, what are next steps here? I recognize that not everybody who's been involved in the discussion here agrees that this is the right approach. It seems that the working group needs to reach a consensus on whether to move forward, request changes, or reject / suggest an alternate approach. I believe I've addressed outstanding comments on the substance of the proposal, so from my perspective, I'm just waiting for a resolution or additional feedback.

Is there any kind of timeline I can expect for when the working group would be able to provide a resolution?

proposals/VK_MSFT_layered_driver.adoc
versalinyaa: Ok, I retract. It's just that I observed the same style in every proposal doc I've read so far.

versalinyaa:
> Hi folks, what are next steps here? [...]
> 
> Is there any kind of timeline I can expect for when the working group would be able to provide a resolution?

This PR looks good to me. I'll raise awareness of it again in the relevant committee today to check if anyone has significant suggestions, but I don't expect any. I suggest proceeding with the next steps.

Next Steps:
- [Jesse] Update the PR with the following. I suggest following `VK_KHR_driver_properties` as an example.
    - Add the API to vk.xml.
    - Add a new file in the `appendices` directory.
    - Add spec language to `chapters/devsandqueues.adoc`.
- [Lina] In parallel, starting now, I will start the Khronos-internal release process checklist.
- [Lina] The Vulkan System Integration committee will give the final PR a review on a Wednesday meeting and approve it.
- Spec gets merged and published approx 1 to 2 weeks later.

jenatali:
After a few iterations of fighting with CI (I didn't try to set up a local spec build) I have something that mostly works, though it does depend on https://github.com/ash-rs/ash/pull/762 before the rust bindings will work.

jenatali:
> Re-kicked the CI because it was designed to pick up the `master` branch of Ash.

Thanks, looks like that's worked and CI is happy.

I downloaded the artifacts as well and I'm content with the results.

MarijnS95:
@jenatali I almost agree, except that it doesn't extend any structure so you're not allowed to push it anywhere. Perhaps the struct is missing `structextends="VkPhysicalDeviceFeatures2,VkDeviceCreateInfo"`?

jenatali:
Good catch.

xml/vk.xml
```diff
@@ -8362,6 +8363,11 @@ typedef void* <name>MTLSharedEvent_id</name>;
             <member optional="true"><type>void</type>*                       <name>pNext</name></member>
             <member><type>VkBool32</type>                                    <name>screenBufferImport</name></member>
         </type>
+        <type category="struct" name="VkPhysicalDeviceLayeredDriverPropertiesMSFT" structextends="VkPhysicalDeviceProperties2" returnedonly="true">>
```
MarijnS95, line 8366:
```suggestion
        <type category="struct" name="VkPhysicalDeviceLayeredDriverPropertiesMSFT" structextends="VkPhysicalDeviceProperties2" returnedonly="true">
```

proposals/VK_MSFT_layered_driver.adoc
```diff
@@ -0,0 +1,58 @@
...
+= Proposal Template
+:toc: left
+:refpage: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/
+:sectnums:
+
+This document proposes an extension to allow the loader to understand driver layering, for improving physical device sorting.
+
+== Problem Statement
+
+The Vulkan loader is able to sort physical devices according to platform-specific criteria. For example, on Windows, the loader uses LUIDs to put physical devices in the same order as DXGI adapters. However, it is possible to have multiple Vulkan drivers that provide support for the same physical device, where one is a "native" vendor-provided implementation and another is a "layered" implementation on top of a different API. Examples of layered implementations would include VulkanOn12 (aka Dozen), layered on D3D12, and MoltenVK, layered on Metal.
```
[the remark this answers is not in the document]
jenatali: I don't get the point, but I've done it.

jenatali:
Thanks, I've addressed those comments, and rebased and resolved the conflicts as well.

versalinyaa:
@jenatali Since nothing in this extension is related to Microsoft, and is equally useful on non-Microsoft platforms, would you be interested in changing it to EXT? Several Khronos members have expressed an interest in the rename.

Changing it to EXT would add some extra overhead, however.
* You would also need to submit CLs to the CTS. For this extension, only light testing would be required.
* A delay of approx 8 weeks between approval and publication, due to additional Khronos-internal process. This process will be automatic from your perspective.

It is your call to make. Keeping it as MSFT is a valid choice if the additional delay is problematic for you.

jenatali:
I'm open to it. The only real complicating factor is that I'm about to take parental leave, so I'm not sure if/when I'd be able to add the relevant CTS tests. Though if the delay lines up with the period while I'm out, that'd be convenient, since I'd be able to start leveraging it once I'm back.

proposals/VK_MSFT_layered_driver.adoc
```diff
@@ -0,0 +1,87 @@
...
+} VkLayeredDriverUnderlyingApiMSFT;
+
+typedef struct VkPhysicalDeviceLayeredDriverPropertiesMSFT {
+    VkStructureType sType;
+    const void* pNext;
+    VkLayeredDriverUnderlyingApiMSFT underlyingAPI;
+} VkPhysicalDeviceLayeredDriverPropertiesMSFT;
+----
+
+Layered drivers should implement this extension. The Vulkan loader can then
```
ianelliottus, line 65: Should this "should" be a "should:"?

r-potter: I assume no, because proposal docs are non-normative?

jenatali: That was my expectation as well.

oddhack:
> > A delay of approx 8 weeks between approval and publication, due to additional Khronos-internal process. This process will be automatic from your perspective.
> 
> @versalinyaa Was this time estimate intended to cover the new EXT ratification process? My intuition is that probably isn't applicable here.

Agreed. As long as there are multiple vendors intending to ship an EXT it can be published. Ratification need not happen first (or at all, though at this point we'd like to ratify new EXTs).

versalinyaa:
@jenatali If you're not on parental leave yet, and are available to answer questions...

Do you want us to merge this branch as-is, with the MSFT prefix? If yes, I can drive that while you're on parental leave.

If you are interested in renaming it to EXT, then there should preferably be a commitment from at least one additional vendor to implement. MoltenVK seems like the obvious choice.

jenatali:
My gut says to merge as-is. I haven't heard interest from another vendor, though I haven't tried to seek it out yet.

oddhack:
> My gut says to merge as-is. I haven't heard interest from another vendor, though I haven't tried to seek it out yet.

@versalinyaa maybe it's worth your pinging the WG mailing list as a reminder that this potentially could be EXT (and ties into the maintenance6 agenda)? If no interest then you drive publication as a vendor extension. If there is interest, would @jenatali be willing to let us promote the proposal to EXT status prior to publication, to move this along?

oddhack:
@jentali can you resolve the conflicts? I don't know if you're still away from work.

jenatali:
I'll be back in November.

oddhack:
@jantali I cannot push to your fork but I can fix the trivial conflicts in my local clone and merge that way. Given discussion above I'm going to do that in order to get this published, unless you or @versalinyaa objects by Thursday evening.

jenatali:
> @jenatali I cannot push to your fork but I can fix the trivial conflicts in my local clone and merge that way. Given discussion above I'm going to do that in order to get this published, unless you or @versalinyaa objects by Thursday evening.
> 
> (Also I appear unable to type your handle correctly the first time, again).

Works for me. This PR is configured to allow edits by project maintainers though so it should be possible to push.

PR #2126, opened 2023-05-18
Add empty extension 530.

mgorchak-blackberry:
Empty extension number 530.


mgorchak-blackberry:
It failed validation because of mixed names in the extension. Just submitted a new one with name "530" everywhere.

PR #2127, opened 2023-05-18
Add new empty extension number 530.

mgorchak-blackberry:


xml/vk.xml
```diff
@@ -22172,6 +22172,13 @@ typedef void* <name>MTLSharedEvent_id</name>;
                 <enum value="&quot;VK_KHR_extension_529&quot;"          name="VK_KHR_EXTENSION_529_EXTENSION_NAME"/>
             </require>
         </extension>
+        <extension name="VK_KHR_extension_530" number="530" author="QNX" contact="Mike Gorchak @mgorchak-blackberry" supported="disabled">
```
oddhack, line 22175: Actually, my bad - should change all the "KHR" in the PR to "QNX" and add the "_QNX" suffix to the flag bit reservation. But structurally it is good.

mgorchak-blackberry: Will do. Cancelling this request now.

krOoze: You realize you can just update it. Creating new tickers is super spammy...

PR #2152, opened 2023-06-22
Clarify mapping between wait semaphores and fences for swapchain present

Rua:
The spec for `VkSwapchainPresentFenceInfoEXT` currently says:
> The application **can** destroy the wait semaphores associated with a given presentation operation when the associated fence is signaled, and **can** destroy the swapchain when the fences associated with all past presentation requests have signaled.

You can specify more than one swapchain, and each swapchain has its own fence, so there is not one single "associated fence". The present operation does not start until the wait semaphores have all been signalled, therefore the signalling of any fence must happen after the wait operation of all the semaphores. This PR clarifies that point.

### DIFF
--- a/chapters/VK_EXT_swapchain_maintenance1/SwapchainPresentFenceInfo.adoc
+++ b/chapters/VK_EXT_swapchain_maintenance1/SwapchainPresentFenceInfo.adoc
@@ -40,9 +40,9 @@ fence must: not signal until all such semaphore payloads have been reset by
 the presentation engine.
 
 The application can: destroy the wait semaphores associated with a given
-presentation operation when the associated fence is signaled, and can:
-destroy the swapchain when the fences associated with all past presentation
-requests have signaled.
+presentation operation when at least one of the associated fences is signaled,
+and can: destroy the swapchain when the fences associated with all past
+presentation requests have signaled.
 
 Fences associated with presentations to the same swapchain on the same
 slink:VkQueue must: be signaled in the same order as the present operations.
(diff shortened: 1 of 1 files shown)


versalinyaa:
The proposed change looks correct to me. If at least one fence has signalled, then all wait semaphores have signalled, IIUC the `VkPresentInfoKHR` section correctly.

@cubanismo @ShabbyX What do you think?

chapters/VK_EXT_swapchain_maintenance1/SwapchainPresentFenceInfo.adoc
```diff
@@ -40,9 +40,9 @@ fence must: not signal until all such semaphore payloads have been reset by
 the presentation engine.
 
 The application can: destroy the wait semaphores associated with a given
-presentation operation when the associated fence is signaled, and can:
-destroy the swapchain when the fences associated with all past presentation
-requests have signaled.
+presentation operation when at least one of the associated fences is signaled,
+and can: destroy the swapchain when the fences associated with all past
+presentation requests have signaled.
```
cubanismo, line 45: Given the ambiguity pointed out by the language changed here, perhaps we should be clearer about the second clause as well. It's all fences associated with all past presentation requests referring to that swapchain.

cubanismo:
Agreed, left a suggestion inline, but this is strictly an improvement as-is.

chapters/VK_EXT_swapchain_maintenance1/SwapchainPresentFenceInfo.adoc
Rua: I've included that as well now.

cubanismo (commented):
Looks good to me now.

stonesthrow (approved):
Reads OK

Issue #2148, opened 2023-06-19
Feature request: Add an "occlusion extents" shader function to enable occlusion culling from within a shader


myaaaaaaaaa:
### Mechanism

This "occlusion extents" function would accept a single point[^box] as input, from which a box would be projected against the depth buffer:
[^box]: Simplified for ease of depiction. A signature like `occlusionExtents(inout vec3 bb_min, inout vec3 bb_max)`, which accepts a bounding box rather than just a point, should have even more utility.

![occbox2](https://github.com/KhronosGroup/Vulkan-Docs/assets/103326468/e7ca8831-cb69-4689-bf49-72c33f6dbbc9)
<sup>LEFT: depth buffer | RIGHT: view from another angle | ORANGE DOT: input point | FLASHING: output box</sup>

Assuming this works as intended, shaders can then use this "occlusion extents box" to test multiple bounding boxes and determine whether their corresponding objects/meshlets should be culled:

![occmeshlet2](https://github.com/KhronosGroup/Vulkan-Docs/assets/103326468/29226391-8be1-4420-bd01-70c39a3fd057)
<sup>LEFT: depth buffer | RIGHT: view from another angle | TOP: a sample model divided into meshlets | BOTTOM: the bounding boxes of the meshlets | FLASHING: meshlets being culled by the "occlusion extents box" | GREEN LINE: forgot to hide the Y axis in blender, sorry</sup>

Implementations should be free to return any "occlusion extents box", as long as rasterizing it would cause all of its fragments to be discarded via depth, frustum, stencil, or scissor tests. The input merely acts as a suggestion.

### Rationale

Assuming this works as intended, this would provide the convenience of a single shader function that can quickly perform depth, frustum, stencil, and scissor tests for multiple objects all at once.

Since it provides more information than the simple pass/fail result of a standard bounding box test, culling can still be performed for sub-objects even if the main object passes, making this more friendly to wider BVH trees.

This is obviously primarily meant for mesh/task shaders (multiple "occlusion extents boxes" can easily fit into the payload to provide the desired level of occlusion granularity), but it also has benefits for tessellation shaders (which can now cull patches before they are tessellated into potentially hundreds of triangles) and compute pre-passes.

### Engine vs Driver

The standard approach for GPU-based occlusion culling is to do it from within the engine by manually copying the depth buffer, building a mipmap chain, and culling against it.

[...]

PR #2163, opened 2023-07-05
Add a proposal for an "occlusion extents" function

myaaaaaaaaa:
#2148 converted into a proposal as recommended by @TomOlson . Apologies in advance for any awkwardness.

Includes potential methods to emulate this function on older hardware.

See #2148 for some GIFs that may make this feature easier to visualize.


TomOlson:
@myaaaaaaaaa Thanks very much! The WG has been discussing next steps - to be honest, it is unusual for us to get a detailed proposal for a new hardware feature that looks like it might be a genuinely good idea, so we did not have an established process for handling it.

We have decided to treat this proposal the same way we treat internal proposals from working group members. That is, the PR will stay open and all members (and in this case, GitHub visitors as well) will be invited to discuss and investigate. This may result in discussion on this PR, the parent issue (#2148), or Khronos internal issues.  If one or more GPU vendors decide to implement, they will propose an extension specification, and when that is accepted into the spec, this proposal document (or a later version of it) will be merged along with the spec language. You will of course get credit as original contributor on the extension, if you can give us a name to list you by.

I have to tell you that the time frame for this is long even if the idea works brilliantly.  The lead time to prove that a new feature is useful, integrate it into the GPU architecture, and build and test the silicon implementation can easily be 4-5 years. So this is unlikely to land in hardware earlier than 2028. It is frustrating, but it is reality.

If by any chance you will be at SIGGRAPH next week, I hope you will come and meet us at the "Vulkan, Forging Ahead" BOF on Wednesday.

myaaaaaaaaa:
@TomOlson Thank you very much for the detailed update! I'm unable to attend SIGGRAPH, but I'd be happy to discuss anything you'd like via email. You can reach me at knthzh (at) gmail (dot) com

PR #2178, opened 2023-07-26
Fix incorrect link to GLSL spec.

PanGao-h:
The actual file name is GLSL_EXT_shader_tile_image.txt instead of GL_EXT_shader_tile_image.txt. This MR fixes the broken link by using the correct name.

Comparing to changing the names of the file, it's perhaps better to change the link name here as suggested in https://github.com/KhronosGroup/GLSL/pull/211


janharaldfredriksen-arm:
Thanks. LGTM.

PR #2189, opened 2023-08-08
Fix missing vkQueue* Command Properties table

krOoze:


oddhack:
Could you explain what this is doing, preferably by reference to a diff listing of the changed markup, to help us evaluate this? Thanks.

PR #2199, opened 2023-08-21
Specify that VUID-RuntimeSpirv-x-06432 applies only to compute shaders

Rua:
Both VUID-RuntimeSpirv-TaskEXT-07294 and VUID-RuntimeSpirv-MeshEXT-07298 specify that the requirement only applies to task and mesh shaders, respectively, but VUID-RuntimeSpirv-x-06432 does not specify that it applies only to compute shaders. Since `maxComputeWorkGroupInvocations` is defined elsewhere as
> the maximum total number of **compute** shader invocations in a single local workgroup

this seems like an omission. This PR adds it, using the first two VUIDs as a template.

I've also given the same treatment to VUID-RuntimeSpirv-x-06429, VUID-RuntimeSpirv-x-06430 and VUID-RuntimeSpirv-x-06431, regarding the `maxComputeWorkGroupSize` limit.

### DIFF
--- a/appendices/spirvenv.adoc
+++ b/appendices/spirvenv.adoc
@@ -1752,8 +1752,9 @@ endif::VK_NV_ray_tracing_invocation_reorder[]
     than or equal to
     slink:VkPhysicalDeviceLimits::pname:maxComputeWorkGroupSize[2]
   * [[VUID-{refpage}-x-06432]]
-    The product of pname:x size, pname:y size, and pname:z size in
-    code:LocalSize or code:LocalSizeId must: be less than or equal to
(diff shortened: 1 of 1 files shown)


oddhack:
@spencer-lunarg please take a look.

spencer-lunarg (approved):
LGTM - This is just making the VU wording consistent with the Mesh/Task shaders VUs

oddhack:
> LGTM - This is just making the VU wording consistent with the Mesh/Task shaders VUs

OK. I think we can just sign it off for the next spec update, then.

PR #2205, opened 2023-08-26
Explicitly state that VkDeviceOrHostAddressConstKHR is also ignored

Rua:
The spec only mentions that `vkGetAccelerationStructureBuildSizesKHR` ignores `VkDeviceOrHostAddressKHR` values, but there are also `VkDeviceOrHostAddressConstKHR` values. `VkAccelerationStructureGeometryTrianglesDataKHR::transformData`, which is mentioned explicitly in the second half of the sentence as an exception, is a `VkDeviceOrHostAddressConstKHR` value. So I assume that both types are really meant to be ignored by this function. This PR clarifies that.


dgkoch (approved):
LGTM. 

Approved by the Ray Tracing TSG on 8/31

PR #2209, opened 2023-09-03
Shorten VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT descriptions

krOoze:


spencer-lunarg:
so while this is "correct" I am curious what it solves.

It adds an indirect lookup for scripts, which adds complexity and not sure what we gain

If a new STAGE was added, I would hope it was consciously added to where it is used and not have to do the indirect lookup if another current stage is a logical-or of it or not

krOoze:
1) As it is, it currently covers almost whole page. That impacts scrollability, readability of this flag soup, and everything.
2) It is less brittle. If bits are added to the subset bit, then they are automatically included in the superset, rather than having to be manually copy pasted there. (DRY principle)

spencer-lunarg:
This was discussed again, the main 2 things are

1. We already are normally explicit in nature with the spec. We use to do this indirect lookup before Sync2 and decide to expand it when this extension was add (lots of confusing what was all in `VK_PIPELINE_STAGE_ALL_TRANSFER_BIT`)
2. I agree it might be a _tad_ less brittle, but considering the rate new graphic stage flags are added, I would hope this type of mistake would be caught in validation layer logic generating this

krOoze:
Not sure what (1) responds to. Nobody was talking about explicitness. Problem is reading the docs and unnecessary verbosity. This one singular Bit covers one whole page (if not two in PDF), and it is not strictly necessary for it to do so.

(2) is slightly naive, given [historical experience](https://github.com/KhronosGroup/Vulkan-Docs/issues/2211) with this class of human errors.

PR #2207, opened 2023-08-29
Create Roadmap Feedback issue template

marty-johnson59:

PR #2215, opened 2023-09-07
Update issue templates

KhronosWebservices:
Marty requested I setup a roadmap tempalte for Vulkan-Docs as adding directly to .github/ISSUE_TEMPLATES was not working correctly.

### DIFF
--- a/.github/ISSUE_TEMPLATE/roadmap-feedback.md
+++ b/.github/ISSUE_TEMPLATE/roadmap-feedback.md
@@ -1,0 +1,32 @@
+---
+name: Roadmap Feedback
+about: Provide feedback to our future roadmaps
+title: "[Roadmap Feedback] "
+labels: ''
+assignees: ''
+
+---
+
+<!--
+Copyright 2018-2023 The Khronos Group Inc.
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+<!-- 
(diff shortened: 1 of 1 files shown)


oddhack:
> Hi @oddhack, looks like new templates need to be configured via settings - which I don't have permissions for. @KhronosWebservices is helping out here - need this one merged for the templates to show up (hopefully).

Isn't this #2207 with the filename changed? Do we want two of them, or is there something specific about the filename that makes it interact properly with the new issues UI?

oddhack:
N.b. I looked at the Settings and all it lets you do is edit an existing template, or create a new one in the web UI. I am guessing that the web UI ends up creating a filename like this instead of the one in #2207 (with the caps / whitespace in the filename) so accepting this and removing the #2207 file would do what is needed.

oddhack:
> I don't see a config .yml for this anywhere.

Are you referring to https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser ?

outofcontrol:
ISTM Marty was wondering if the config.yml was required to get the first template working. Which I don't believe it is with this PR.

marty-johnson59:
Yah, that was just me guessing... But James' PR should be all we need (hopefully)

oddhack:
> Yah, that was just me guessing... But James' PR should be all we need (hopefully)

I accepted this and removed the "Roadmap Feedback.md" copy. Opening a new issue now brings up a menu with the two templates (or a blank issue) so hopefully that resolves it.

krOoze:
There is a risk the Roadmap template will be used indiscriminately for any and every suggestion.

Everyone would obviously feel their idea is the best and should go on a Roadmap (and by extension onto support of virtually every device). Then again, there's probably nothing wrong with that. I propose changing it to generic "Feature suggestion" template.

(Which would also be clearer template name IMO. "Roadmap Feedback" sounds kinda too mysterious, and not immediately clear what it is for.)

PR #2219, opened 2023-09-11
Express "ALL_*" pipeline stages as including _valid_ bits

### LINKED ISSUE #2187: clarification on the meaning of `VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT`
§ 7.1.2 currently defines `VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT` as 
> [...] specifies the execution of all graphics pipeline stages, and is equivalent to the logical OR of [list of bits]

At the same time, `VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT` is defined as (emphasis mine)
> [...] specifies all operations performed by all commands **supported on the queue** it is used with.

While `ALL_COMMANDS` explicitely mentions that only supported features are included, `ALL_GRAPHICS` doesn't; furthermore, it is specified as being equivalent to a logical OR of all the specific bits. Since a stage mask must not contain unsupported features, this could be interpreted as that `ALL_GRAPHICS` cannot be used unless all specific features are supported by the device.

The validation layers do interpret `ALL_GRAPHICS` to only include supported features, which is most likely what is intended.

If that is the case, the wording should be changed to make clear that it's the logical OR of *all supported* bits.

Tobski:
Fix #2187

### DIFF
--- a/chapters/synchronization.adoc
+++ b/chapters/synchronization.adoc
@@ -440,10 +440,11 @@ ifdef::VK_KHR_ray_tracing_maintenance1[]
 endif::VK_KHR_ray_tracing_maintenance1[]
 endif::VK_KHR_acceleration_structure,VK_NV_ray_tracing[]
   * ename:VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT specifies the execution of
-    all graphics pipeline stages, and is equivalent to the logical OR of:
+    all graphics pipeline stages, that would be valid in the context that
+    this bitflag is provided, from the following list of stages:
 include::{generated}/sync/flagDefinitions/VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT.adoc[]
-  * ename:VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT specifies all operations
-    performed by all commands supported on the queue it is used with.
+  * ename:VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT specifies all pipeline stages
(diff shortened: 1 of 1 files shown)


krOoze (commented):
Following shorthand Bits should potentially receive analogous treatment: 
- `VK_PIPELINE_STAGE_2_PRE_RASTERIZATION_SHADERS_BIT`
- `VK_PIPELINE_STAGE_2_ALL_TRANSFER_BIT`
- `VK_PIPELINE_STAGE_2_VERTEX_INPUT_BIT`
- `VK_ACCESS_MEMORY_*_BIT`
- `VK_ACCESS_2_SHADER_READ_BIT`
- `VK_ACCESS_2_SHADER_WRITE_BIT`

chapters/synchronization.adoc
```diff
@@ -440,10 +440,11 @@ ifdef::VK_KHR_ray_tracing_maintenance1[]
 endif::VK_KHR_ray_tracing_maintenance1[]
 endif::VK_KHR_acceleration_structure,VK_NV_ray_tracing[]
   * ename:VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT specifies the execution of
-    all graphics pipeline stages, and is equivalent to the logical OR of:
+    all graphics pipeline stages, that would be valid in the context that
```
krOoze, line 443: "graphics pipeline stages" is somewhat misleading, since it include Bits frome e.g. subpass shading pipeline.

PR #2247, opened 2023-10-11
Don't use CURDIR to find spec files

ShabbyX:
... or rather have it default to CURDIR but be overridable.


ShabbyX:
@oddhack context is I'm adding some "integration" tests that build a test spec and make sure the output is as expected. This is because otherwise our asciidoc extensions are not tested at all.

PR #2248, opened 2023-10-11
Add a spec build test

ShabbyX:
This change adds a test spec filled mostly with Lorem Ipsum that can be
built in various versions and extensions.  The build results are checked
against expectations to ensure the spec output is as expected.

This is primarily a test of the asciidoc extensions that have been
implemented in this repository whose functionality is not automatically
tested.


ShabbyX:
The first commit is this PR: https://github.com/KhronosGroup/Vulkan-Docs/pull/2247

I ended up doing this as part of codified VU tests because everything can be unit tested there except the integration with asciidoc. This is the setup for that, but it's also a good thing on its own to make sure we don't accidentally break things when touching the asciidoc extensions.

ShabbyX:
@oddhack, I didn't find out how to make the expectations directory exempt from the license check, any ideas how to deal with that failure?

oddhack:
> @oddhack, I didn't find out how to make the expectations directory exempt from the license check, any ideas how to deal with that failure?

REUSE requires that every file has a license, but the license does not have to be *in* the file, if that is the problem. You can add a directory/*html wildcard to .reuse/dep5 (probably just adding to the first clause, which places the specified files under CC-BY-4.0 like most repo content).

oddhack:
I don't know how stable asciidoctor HTML output is across updates. Probably quite stable, but if we do this we should be prepared for everything failing simultaneously at some time in the future when the Docker image gets rebuilt with a newer version. I expect it would not be more complicated than rebuilding the HTML reference output, though?

oddhack:
> That reminds me, I haven't added this to CI yet. I suggest doing that in a follow up dedicated commit that can be reverted or at least easy to look over in case there is a need for revert/disable.

Dedicated commit in this PR, or a separate PR? Once the REUSE issue is fixed this could go in as is for the Friday update, but the internal gitlab CI is really more important since that's where most changes happen. So merging this and doing CI updates once gitlab has the changes makes sense to me.

ShabbyX:
> > That reminds me, I haven't added this to CI yet. I suggest doing that in a follow up dedicated commit that can be reverted or at least easy to look over in case there is a need for revert/disable.
> 
> Dedicated commit in this PR, or a separate PR? Once the REUSE issue is fixed this could go in as is for the Friday update, but the internal gitlab CI is really more important since that's where most changes happen. So merging this and doing CI updates once gitlab has the changes makes sense to me.

Separate PR, definitely. We can get a chance to verify the output is stable across timezones, versions, commit hash etc (and fix it if it isn't).

oddhack (approved):
LGTM once REUSE issues are fixed.

oddhack:
> cd build_tests && ./update-expectations

Probably worth documenting the update process so when a output format change actually happens someday, whomever deals with it will get a running start. If the testBuild failure message advised to that effect that seems useful.

ShabbyX:
> You can add a directory/*html wildcard to .reuse/dep5 (probably just adding to the first clause, which places the specified files under CC-BY-4.0 like most repo content).

Done, thanks!

> Probably worth documenting the update process so when a output format change actually happens someday, whomever deals with it will get a running start. If the testBuild failure message advised to that effect that seems useful.

Done, the script now outputs instructions in case of failure. There is a new README.adoc (and I added a similar one for reflow-tests too).

---

Also noticed a bit of instability due to a footer asciidoctor inserts that has the date in it. I removed it with `:last-update-label!:`.

PR #2252, opened 2023-10-17
Add/correct a couple of optional attributes

expipiplus1:
To bring them both in line with other dual-use count parameters, as found in other commands or structs.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -13748,7 +13748,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
             <param><type>VkDevice</type> <name>device</name></param>
             <param><type>VkFaultQueryBehavior</type> <name>faultQueryBehavior</name></param>
             <param><type>VkBool32</type>* <name>pUnrecordedFaults</name></param>
-            <param><type>uint32_t</type>* <name>pFaultCount</name></param>
+            <param optional="false,true"><type>uint32_t</type>* <name>pFaultCount</name></param>
             <param optional="true" len="pFaultCount"><type>VkFaultData</type>* <name>pFaults</name></param>
         </command>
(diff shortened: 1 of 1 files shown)


xml/vk.xml
```diff
@@ -8776,7 +8776,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
         <type category="struct" name="VkLatencySurfaceCapabilitiesNV" structextends="VkSurfaceCapabilities2KHR">
             <member values="VK_STRUCTURE_TYPE_LATENCY_SURFACE_CAPABILITIES_NV"><type>VkStructureType</type> <name>sType</name></member>
             <member optional="true">const <type>void</type>*                              <name>pNext</name></member>
-            <member optional="false,true"><type>uint32_t</type>                           <name>presentModeCount</name></member>
+            <member optional="true"><type>uint32_t</type>                                 <name>presentModeCount</name></member>
             <member optional="true" len="presentModeCount"><type>VkPresentModeKHR</type>* <name>pPresentModes</name></member>
```
krOoze, line 8780: These feel like they should not have `optional` attributes at all, but should be marked `returnedonly`.

chansen-nv: The user can write to `presentModeCount` to get a reduced subset of optimized present modes so `returnedonly` doesn't seem to apply here:
```
If pname:pPresentModes is `NULL`, then the number of present modes that are
optimized for use with low latency mode returned in pname:presentModeCount.
Otherwise, pname:presentModeCount must be set by the user to the number of
elements in the pname:pPresentModes array, and on return the variable is
overwritten with the number of values actually written to
pname:pPresentModes.
If the value of pname:presentModeCount is less than the number of optimized
present modes, at most pname:presentModeCount values will be written to
pname:pPresentModes.
```

Issue #2262, opened 2023-10-21
Conflicting documation for VkCudaLaunchInfoNV::pParams


expipiplus1:
Distinct from https://github.com/KhronosGroup/Vulkan-Docs/issues/2261

The spec states both: 

- pParams is a pointer to an array of paramCount pointers, corresponding to the arguments of function
- If paramCount is not 0, and pParams is not NULL, pParams must be a valid pointer to an array of paramCount bytes

This is caused by the `optional=true` on `pParams`, which indicates that despite a non-zero `paramCount` value, `pParams` may still be null.

(this also applies to `pExtras`)

tlorach:
removing  optional=true for pParam sounds the right choice. 
Now, I it should be valid that paramCount = 0 and pParams = NULL ... in other words, pParam can only be NULL if paramCount = 0

PR #2263, opened 2023-10-21
Remove incorrect optional attribute on VkCudaLaunchInfoNV::pParams and pExtras

expipiplus1:
Fixes https://github.com/KhronosGroup/Vulkan-Docs/issues/2262

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -7627,9 +7627,9 @@ typedef void* <name>MTLSharedEvent_id</name>;
             <member><type>uint32_t</type>               <name>blockDimZ</name></member>
             <member><type>uint32_t</type>               <name>sharedMemBytes</name></member>
             <member optional="true"><type>size_t</type>                 <name>paramCount</name></member>
-            <member optional="true" len="paramCount">const <type>void</type>* const *    <name>pParams</name></member>
+            <member len="paramCount">const <type>void</type>* const *   <name>pParams</name></member>
             <member optional="true"><type>size_t</type>                 <name>extraCount</name></member>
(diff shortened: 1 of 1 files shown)


xml/vk.xml
```diff
@@ -7627,9 +7627,9 @@ typedef void* <name>MTLSharedEvent_id</name>;
             <member><type>uint32_t</type>               <name>blockDimZ</name></member>
             <member><type>uint32_t</type>               <name>sharedMemBytes</name></member>
             <member optional="true"><type>size_t</type>                 <name>paramCount</name></member>
-            <member optional="true" len="paramCount">const <type>void</type>* const *    <name>pParams</name></member>
+            <member len="paramCount">const <type>void</type>* const *   <name>pParams</name></member>
```
Snektron, line 7630: Now there is no information regarding the inner pointer. It is unclear to me whether these are allowed to be null or not, but in any case, I would expect either `optional="false,true"` or `optional="false,false"` here.

expipiplus1: yes, that would be better. Not having an `optional` attribute is definitely the exception. According to the referenced documentation here https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXEC.html#group__CUDA__EXEC_1gb8f3dc3031b40da29d5f9a7139e52e15 it seems as though this ought to be `false,false`.

tlorach: removing  optional=true for pParam sounds like the right choice.

tlorach: Note: I have the feeling the same `optional="true"` should be removed from  
`<member optional="true" len="extraCount">const <type>void</type>* const *    <name>pExtras</name></member>`

PR #2264, opened 2023-10-23
Add VK_KHR_timeline_semaphore dependency for VK_NV_low_latency2.

ishitatsuyuki:
vkLatencySleepNV requires the use of a timeline semaphore, therefore this extension naturally depends on the timeline semaphore extension or Vulkan 1.2 where it is promoted to Core.


krOoze:
Why can't it work with binary semaphores as well though?

ishitatsuyuki:
My guess is that binary semaphores cannot be signaled from CPU code, but Reflex most likely is a tech running on CPU. It probably doesn't matter if it's an internal driver semaphore though.

krOoze:
Right. A Fence then...

pdaniell-nv (approved):

PR #2286, opened 2023-12-28
Add epub generation

raSTARgfx:
Hey all,

I really wanted the spec for my kindle scribe so I poked a bit at getting asciidoc to generate epub. Not really sure if this is something you are all interested in but I figured the PR wouldn't hurt.

Here is a sample of the output generated by `./makeSpec -spec all epub`
[vkspec.zip](https://github.com/KhronosGroup/Vulkan-Docs/files/13789917/vkspec.zip)

Let me know if there is anything I can do to improve this. I mostly followed what was already there but I'm not an expert on asciidoc so I left out the bits I didn't know much about. This requires that https://docs.asciidoctor.org/epub3-converter/latest/ be installed but I didn't add any checks for that or anything as that seems to also be the case for pdf.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,7 +243,12 @@ ADOCPDFOPTS  = $(ADOCPDFEXTS) -a mathematical-format=svg \
 	       -a pdf-fontsdir=$(CONFIGS)/fonts,GEM_FONTS_DIR \
 	       -a pdf-stylesdir=$(CONFIGS)/themes -a pdf-style=pdf
 
-ADOCEPUBOPTS = -r asciidoctor-epub3
+# EPUB target-specific Asciidoctor options
+ADOCEPUBOPTS = -r asciidoctor-epub3 \
+	       -r asciidoctor-mathematical \
+	       -r $(CONFIGS)/asciidoctor-mathematical-ext.rb \
+	       -a mathematical-format=svg \
+	       -a imagesoutdir=$(PDFMATHDIR)
 
 # Valid usage-specific Asciidoctor extensions and options
 ADOCVUEXTS = -r $(CONFIGS)/vu-to-json.rb -r $(CONFIGS)/quiet-include-failure.rb
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,7 @@ OUTDIR	  = $(GENERATED)/out
 HTMLDIR   = $(OUTDIR)/html
 VUDIR	  = $(OUTDIR)/validation
 PDFDIR	  = $(OUTDIR)/pdf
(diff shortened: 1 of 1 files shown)


oddhack:
How does this deal with latexmath equations (there are a lot in the texturing / image operations chapter for reference)? ISTR looking at this a few years ago at which time there were some issues.

raSTARgfx:
Thank you for pointing me in that direction. I wasn't aware of it so it was missed.

With some tinkering I was able to get it to look correct (at least to my eye).

There is something weird about generating the svg's locally where it generates a bunch of mojibake. Not sure if it's because I'm missing a font or something. Quickly installing the requirements to the docker container:

```
gem install asciidoctor-epub3 rouge
./makeSpec -clean -spec core epub
```

2023-12-28 23_52_26-Vulkan® 1 3 274 - A Specification  EPUB  — E-book viewer

Compared to the PDF
2023-12-29 00_08_50-Vulkan® 1 3 274 - A Specification - Foxit PDF Reader

The only other issue I could spot was some issues with scaling:
2023-12-29 00_24_26-Vulkan® 1 3 274 - A Specification  EPUB  — E-book viewer

oddhack:
This is clearly potentially useful for some developers, but at the same time, it's not something we ourselves want to be responsible for - we are publishing way too many Vulkan spec artifacts already.

My inclination is to put the new target into the Makefile, but make clear that this is unsupported by us and community-contributed, such that people would need to build their own EPUBs from the repository. Would that be OK for you?

oddhack:
Also, we'd like to be able to point people to you if there are issues. Unfortunately we apparently cannot file issues on a repo fork, but if you open a tracking issue here we can hold it open indefinitely and point people to it.

raSTARgfx:
> My inclination is to put the new target into the Makefile, but make clear that this is unsupported by us and community-contributed, such that people would need to build their own EPUBs from the repository.

That sounds good to me.

Let me know if there is anything I can do with this PR to facilitate making that clear.

> Also, we'd like to be able to point people to you if there are issues.

Yup that is fine.

> you open a tracking issue here we can hold it open indefinitely and point people to it.

Sure thing I'll do that once everything is finalized here and link back to this discussion.

oddhack:
> > you open a tracking issue here we can hold it open indefinitely and point people to it.
> 
> Sure thing I'll do that once everything is finalized here and link back to this discussion.

This has been merged and published in the 1.3.276 spec update. If you would like to create a tracking issue for epub related issues, please go ahead, and I will link to that from the BUILD.adoc and a comment in the Makefile for the next update. Thanks!

PR #2328, opened 2024-03-15
fix VK_ARM_render_pass_striped dependency

star-e:
Current VK_ARM_render_pass_striped depends on [VK_KHR_get_physical_device_properties2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_get_physical_device_properties2.html) **OR** [VK_KHR_synchronization2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_synchronization2.html)

I think it might be a typo, and it should be **AND**.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -22639,7 +22639,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
                 <enum value="&quot;VK_ARM_extension_424&quot;"          name="VK_ARM_EXTENSION_424_EXTENSION_NAME"/>
             </require>
         </extension>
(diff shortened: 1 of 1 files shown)


janharaldfredriksen-arm:
Hi @star-e,
You are correct. This was found internally as well and is being fixed as part of another change. Those updates should land in the next spec update (probably next week) and should also address this issue.
Let's keep this open until the next spec update is out and we can confirm the fix.

(For Khronos-internal reference, this is fixed in gitlab MR 6536.)

star-e:
Thanks for the information! I'm looking forward to the next update.

oddhack:
This has been fixed in the published spec - closing. Please reopen if you disagree.

PR #2331, opened 2024-03-20
Replace <feature number> attribute with <feature depends>

### LINKED ISSUE #2327: Replace 'number' attributes in feature tags with 'depends' attributes
Spun off #2326. The goal would be to say e.g.

> \<feature name="VK_VERSION_1_1" depends="VK_VERSION_1_0">  
> \<feature name="VK_VERSION_1_2" depends="VK_VERSION_1_1">

and make use of that (transitive) dependency relationship in the generator scripts to produce the same output ordering as we see today. It would also simplify the 'makeSpec' driver script a bit as it would not need to know all the other features that go into e.g. VK_VERSION_1_3 or VKSC_VERSION_1_0.

Downsides, there do appear to be a couple of uses of this in Vulkan-ValidationLayers and VK-GL-CTS which we have not scoped. Language binding generators like [Vulkan-Hpp](https://github.com/KhronosGroup/Vulkan-Hpp/issues/1825) and [Ash](https://github.com/ash-rs/ash/issues/873) might need to make minor changes, as well.

A reasonable path is to add the depends attributes and only remove 'number' in the future when we're more confident it will not impact downstreams.

oddhack:
Initially this PR only updates the XML schema and vk.xml. For the moment, the 'number' attribute is simply deprecated, but still required, until we can verify downstream repos are not affected.

The generator scripts will need to be modified to do a topological sort on the dependencies, and downstream repositories such as validation layers and Vulkan-Hpp will need to be updated to remove or replace their usage of the old attribute.

Closes #2327

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -15917,7 +15917,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
             <command name="vkCmdExecuteCommands"/>
         </require>
     </feature>
-    <feature api="vulkan,vulkansc" name="VK_VERSION_1_1" depends="VK_VERSION_1_1" comment="Vulkan 1.1 core API interface definitions.">
+    <feature api="vulkan,vulkansc" name="VK_VERSION_1_1" depends="VK_VERSION_1_0" comment="Vulkan 1.1 core API interface definitions.">
         <require>
             <type name="VK_API_VERSION_1_1"/>
         </require>
@@ -16249,7 +16249,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
             <type name="VkPhysicalDeviceShaderDrawParametersFeatures"/>
         </require>
     </feature>
-    <feature api="vulkan,vulkansc" name="VK_VERSION_1_2" depends="VK_VERSION_1_2" comment="Vulkan 1.2 core API interface definitions.">
+    <feature api="vulkan,vulkansc" name="VK_VERSION_1_2" depends="VK_VERSION_1_1" comment="Vulkan 1.2 core API interface definitions.">
         <require>
             <type name="VK_API_VERSION_1_2"/>
         </require>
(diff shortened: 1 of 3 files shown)


oddhack:
Probably the ability to include extension names in the feature dependency expression should be removed, since core versions are unlikely to be dependent on extensions.

xml/vk.xml
```diff
@@ -16684,7 +16684,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
         </require>
     </feature>
 
-    <feature api="vulkansc" name="VKSC_VERSION_1_0" number="1.0" comment="Vulkan SC core API interface definitions">
+    <feature api="vulkansc" name="VKSC_VERSION_1_0" depends="VK_VERSION_1_2" comment="Vulkan SC core API interface definitions">
```
dgkoch, line 16687: should this be: 
```suggestion
    <feature api="vulkansc" name="VKSC_VERSION_1_0" depends="VKSC_VERSION_1_0" comment="Vulkan SC core API interface definitions">
```
?

oddhack: Definitely would not have it dependent on itself, that's at best a no-op and at worst an infinite loop. The 1.2 dependency is because SC 1.0 is defined as a derivative of Vulkan 1.2.

There are only two (intended) usages for this:

* Sort the features when emitting headers, in dependency order rather than 'number' order.
* Allow the build scripts to automatically incorporate the dependent versions rather than having to explicitly list them out. E.g. right now when you ask for 'makeSpec -version sc1.0' there's a static array of other things to automatically turn on:

>         'sc1.0' : [ 'VK_VERSION_1_0', 'VK_VERSION_1_1', 'VK_VERSION_1_2', 'VKSC_VERSION_1_0' ],

dgkoch: I'm confused then, because VK_VERSION_1_1 .. VK_VERSION_1_3 depend on themselves in this MR. Are they supposed to depend on the previous version instead?

oddhack: Woops, yes. Thanks for catching that.

oddhack:
It will take a while to remove downstream use of 'number', so changed this to deprecate it rather than removing it yet. This should be good to go now as it's a purely additive change to the schema.

dgkoch (approved):

PR #2340, opened 2024-03-27
Fix how rowExtent and sliceExtent are calculated for copy operations

gregschlom:
`bufferRowLength` and `bufferImageHeight` are specified in texels, not bytes, therefore it is necessary to multiply them by the block size to get the correct row extent and slice extent in bytes.


pdaniell-nv (approved):

r-potter:
Intersects with internal MR: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6588

gregschlom:
Cool, we can close this, looks like Tobias' PR covers this change

PR #2352, opened 2024-04-14
Resolve warnings in Actions UI by upgrading all actions from @v3 to @v4

oddhack:
The most problematic case is paired download/upload-artifact, which requires unique artifact names. Since we don't run this under a test matrix I don't think it should be a problem.


oddhack:
This ran CI successfully and generated identical artifacts (up to noise-level stuff like date tags on outputs) as main, so merging now to get those warnings off the table.

PR #2355, opened 2024-04-17
Try to fix a typo and additional remark physical-device-level function validusage for Newer Core Versions

FuXiii:
## First
image
The url point to [Extending Physical Device From Device Extensions](https://docs.vulkan.org/spec/latest/chapters/initialization.html#initialization-phys-dev-extensions) but it write `Extending Physical Device Core Functionality` maybe this is a typo.

## Second
image
The `Physical-device-level functionality valid Usage` describe maybe need add [Extending Physical Device Core Functionality](https://docs.vulkan.org/spec/latest/chapters/initialization.html#_extending_physical_device_core_functionality) link as a supplement.

The `Extending Physical Device Core Functionality`:
---
image

### DIFF
--- a/chapters/fundamentals.adoc
+++ b/chapters/fundamentals.adoc
@@ -1255,7 +1255,7 @@ ifdef::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
 Physical-device-level functionality or behavior added by a <<extensions,
 device extension>> to the API must: not be used unless the conditions
 described in <<initialization-phys-dev-extensions, Extending Physical Device
-Core Functionality>> are met.
+From Device Extensions>> are met.
 endif::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
 
 Device-level functionality added by a <<extensions, device extension>> that
@@ -1281,6 +1281,9 @@ core version>> of the API must: not be used unless it is supported by the
 physical device as determined by
 slink:VkPhysicalDeviceProperties::pname:apiVersion and the specified version
 of slink:VkApplicationInfo::pname:apiVersion.
+Or the conditions described in 
+<<_extending_physical_device_core_functionality, Extending Physical Device
+Core Functionality>> are met.
 
 Device-level functionality or behavior added by a <<versions, new core
 version>> of the API must: not be used unless it is supported by the device


chapters/fundamentals.adoc
```diff
@@ -1281,6 +1281,9 @@ core version>> of the API must: not be used unless it is supported by the
 physical device as determined by
 slink:VkPhysicalDeviceProperties::pname:apiVersion and the specified version
 of slink:VkApplicationInfo::pname:apiVersion.
+Or the conditions described in 
+<<_extending_physical_device_core_functionality, Extending Physical Device
+Core Functionality>> are met.
```
oddhack, line 1286: We reviewed this and think the new language is redundant with the paragraph as it stands - if we've missed something and that's not true, please say more. Otherwise, we suggest removing this and just leaving the typo fix in the first change, which is clearly correct.

FuXiii: > Physical-device-level functionality or behavior added by a [new core version](https://docs.vulkan.org/spec/latest/appendices/versions.html#versions) of the API `must not` be used unless it is supported by the physical device as determined by [VkPhysicalDeviceProperties](https://docs.vulkan.org/spec/latest/chapters/devsandqueues.html#VkPhysicalDeviceProperties)::`apiVersion` and the specified version of [VkApplicationInfo](https://docs.vulkan.org/spec/latest/chapters/initialization.html#VkApplicationInfo)::`apiVersion`.

Pseudocode may looks like:
```C++
vk_pfn_SomeOnePhysicalDeviceFunction* sopdf = ...;

//If we have a function can get the Vulkan function release version
uint32_t sopdf_release_version = GetPhysicalDeviceFunctionReleaseVulkanVersion(sopdf);

if(VkPhysicalDeviceProperties::apiVersion >= sopdf_release_version && VkApplicationInfo::apiVersion >= sopdf_release_version)
{
    sopdf(...);//Can just use it
}
```
---

> New core physical-device-level functionality can be used when the physical-device version is greater than or equal to the version of Vulkan that added the new functionality. The Vulkan version supported by a physical device can be obtained by calling [vkGetPhysicalDeviceProperties](https://docs.vulkan.org/spec/latest/chapters/devsandqueues.html#vkGetPhysicalDeviceProperties).

Pseudocode may looks like:
```C++
vk_pfn_SomeOnePhysicalDeviceFunction* sopdf = ...;

//If we have a function can get the Vulkan function release version
uint32_t sopdf_release_version = GetPhysicalDeviceFunctionReleaseVulkanVersion(sopdf);

if(VkPhysicalDeviceProperties::apiVersion >= sopdf_release_version)
{
    sopdf(...);//Can just use it
}
```
According to the [Extending Physical Device Core Functionality](https://docs.vulkan.org/spec/latest/chapters/initialization.html#_extending_physical_device_core_functionality) standards, there seems to be ambiguity with [Valid Usage for Newer Core Versions](https://docs.vulkan.org/spec/latest/chapters/fundamentals.html#fundamentals-validusage-versions)


* `Ambiguity` : miss conditional judgment on `VkApplicationInfo::apiVersion`

FuXiii: Did I misunderstand something about vulkan spec? Please give me some hint.

oddhack:
@r-potter will revisit this and comment.

chapters/fundamentals.adoc
```diff
@@ -1255,7 +1255,7 @@ ifdef::VK_VERSION_1_1,VK_KHR_get_physical_device_properties2[]
 Physical-device-level functionality or behavior added by a <<extensions,
 device extension>> to the API must: not be used unless the conditions
 described in <<initialization-phys-dev-extensions, Extending Physical Device
-Core Functionality>> are met.
+From Device Extensions>> are met.
```
r-potter, line 1258: I agree that this one is a simple typo and we should accept this part of the change.

chapters/fundamentals.adoc
r-potter: There is an ambiguity here, but I don't think the proposed test resolves it.

* Currently one section says you must check both `VkPhysicalDeviceProperties::apiVersion` and `VkApplicationInfo::apiVersion`, while the other only requires checking `VkPhysicalDeviceProperties::apiVersion`.
* Rather than add a cross-reference to the second section, we should simply make then require the same thing.
* My recommendation would be that we add the `VkApplicationInfo::apiVersion` requirement to "Extending Physical Device Core Functionality", but the working group should discuss this in the next maintenance call.

FuXiii: OK. Thank you. I will keep this `PR` until the conclusion is reached at the next maintenance call. :D

oddhack: @fuxii @r-potter did a (different) spec update for the current Vulkan spec affecting the "Extending Physical Device Core Functionality" section (as well as incorporating the typo below). Please review and see if the current language reflects your concerns such that we could close this PR. Thanks!

FuXiii: Create a new PR or Issue for `Extending Physical Device Core Functionality` ?

oddhack: Please do create a new issue.

FuXiii: OK, I had created a Issue [Physical-Device-Level function valid usage for Newer Core Versions](https://github.com/KhronosGroup/Vulkan-Docs/issues/2384) , and I  will close this PR, and thanks for you help. @oddhack @r-potter

PR #2366, opened 2024-05-08
Add a "constants" type= attribute to <enums> tags

### LINKED ISSUE #2359: Add type="constants" to the "API Constants" "enums" in vk.xml
In vk.xml "enums" is almost an internally tagged enum with the key "type", but the "API Constants" enum doesn't have a type, and it is the only place where there is no type, I suggest adding it and make changes to the schema that make type a mandatory key, this is a minor change to remove that specialisation for it when you write a binding generator. I had some problems with this because I couldn't just codegen the parser here.

Here is the line:
`<enums name="API Constants" comment="Vulkan hardcoded constants - not an enumerated type, part of the header boilerplate">`
I suggest adding `type="constants"` there.

In the schema "?" should be removed: `attribute type { text } ? ,`

I also noticed that "name" is optional too in the schema, even though it is always provided, maybe it can be made mandatory too? Though I don't care about it.

oddhack:
For more consistency with handling other enums type= attributes.

This is purely additive to the schema but it is conceivable that downstream scripts would look for this attribute *not* being present to indicate compile-time constants, so this should be checked against CTS / VVL / Vulkan-Hpp binding / Rust binding usage.

Closes #2359

### DIFF
--- a/registry.adoc
+++ b/registry.adoc
@@ -806,9 +806,9 @@ was based on.
     group.
     If present, this attribute should match the attr:name attribute of a
     corresponding tag:type tag.
-  * attr:type - optional.
+  * attr:type - required.
     String describing the data type of the values of this group of enums.
-    At present the only accepted categories are `enum` and `bitmask`, as
+    The accepted categories are `bitmask`, `constants`, and `enum`, as
     described below.
   * attr:comment - optional.
     Arbitrary string (unused).
@@ -849,7 +849,7 @@ this results in
 [[tag-enum]]
 = Enumerants (tag:enum Tag)
 
-Each tag:enum tag defines a single Vulkan (or other API) token.
+Each tag:enum tag defines a single API token.
 
 == Attributes of tag:enum Tags
 
@@ -2404,10 +2404,10 @@ These correspond to the definitions found in the first tag:enums section of
 
 [source,xml]
 --------------------------------------
-<enums name="API Constants"
+<enums name="API Constants" type="constants"
        comment="Vulkan hardcoded constants - not an enumerated type, part of the header boilerplate">
-    <enum value="256"        name="VK_MAX_PHYSICAL_DEVICE_NAME"/>
-    <enum value="256"        name="VK_MAX_EXTENSION_NAME"/>
+    <enum type="uint32_t" value="256"   name="VK_MAX_PHYSICAL_DEVICE_NAME"/>
+    <enum type="uint32_t" value="256"   name="VK_MAX_EXTENSION_NAME"/>
     ...
 --------------------------------------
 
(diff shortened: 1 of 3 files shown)


oddhack:
@MarijnS95 this changes seems to have triggered something bad in ash-rs - could you take a look at it? This PR is an experiment to see if the request in #2359 can be implemented without annoying downstream consequences.

spencer-lunarg:
@oddhack confirmed that VVL is ok with this

oddhack:
> @oddhack confirmed that VVL is ok with this

Thanks! Do you know who I could ask to check CTS?

spencer-lunarg:
> Thanks! Do you know who I could ask to check CTS?

@rg3igalia

MarijnS95:
@oddhack it looks like it is skipping generation of all the constants. Unfortunately I just left the country and will only be back on Monday to investigate and address this, if that's not too long.

oddhack:
> @oddhack it looks like it is skipping generation of all the constants. Unfortunately I just left the country and will only be back on Monday to investigate and address this, if that's not too long.

Thanks. That's fine - this is not urgent.

oddhack:
@rg3igalia would it be possible for you to check this branch to see if the XML change (adding a type="constants" to the API constants \<enums> tag) causes any issues for CTS?

rg3igalia:
> @rg3igalia would it be possible for you to check this branch to see if the XML change (adding a type="constants" to the API constants <enums> tag) causes any issues for CTS?

I just checked and CTS code generation continues to work normally. The generated code also compiles fine.

MarijnS95:
@oddhack the `ash` change needed to get this working is in https://github.com/krolli/vk-parse/pull/32.

oddhack:
I believe this is ready to merge. VVL and CTS have verified no impact, Vulkan-Hpp is causing no issues, and Ash has a fix queued up.

oddhack:
@MarijnS95 thanks, re-running CI to see if that will fix the problem.

MarijnS95:
@oddhack looks like it did :tada: 

In the future there's a "re-run CI" button, with some nice "Run number #x" UX.

oddhack:
> @oddhack looks like it did 🎉
> 
> In the future there's a "re-run CI" button, with some nice "Run number #x" UX.

I did initially re-run the failing job, which continued to fail, making me think the dependency had succeeded but generated incorrect code for the build-test job. Could probably have re-ran all jobs with equal effect to pushing the empty commit, true.

MarijnS95:
> making me think the dependency had succeeded but generated incorrect code for the build-test job.

Correct, the generator succeeded but didn't find any constants by looking for `<enums>` without `type` attribute (since it's `type="constants"`) now, so omitted generating them, resulting in compilation failures in the second step.

PR #2380, opened 2024-06-05
Add driver ID for Honeykrisp

alyssarosenzweig:
The Vulkan driver for Apple Silicon. 1.3 conformant [1] and should ship later this year

AGXV never shipped or passed CTS, so dropping its ID should be relatively safe from a backwards compatibility perspective.

[1] https://rosenzweig.io/blog/vk13-on-the-m1-in-1-month.html

--

cc @1ace @ella-0


alyssarosenzweig:
Resubmitting internally in khronos

PR #2396, opened 2024-07-22
Remove dEQP-ism from generated CTS helpers

ShabbyX:
CTS no longer uses the de* types.

### DIFF
--- a/scripts/json_generator.py
+++ b/scripts/json_generator.py
@@ -118,10 +118,10 @@ encodeBase64CodeCTS = """
 class Base64Formatter
 {
 public:
-	const deUint8*	data;
+	const uint8_t*	data;
 	int				numBytes;
 
-	Base64Formatter(const deUint8* data_, int numBytes_) : data(data_), numBytes(numBytes_) {}
+	Base64Formatter(const uint8_t* data_, int numBytes_) : data(data_), numBytes(numBytes_) {}
 };
 
 std::ostream& operator<< (std::ostream& str, const Base64Formatter& fmt)
@@ -135,7 +135,7 @@ std::ostream& operator<< (std::ostream& str, const Base64Formatter& fmt)
 		'0','1','2','3','4','5','6','7','8','9','+','/'
 	};
 
-	const deUint8*	data = fmt.data;
+	const uint8_t*	data = fmt.data;
 	int				numBytes = fmt.numBytes;
 	int				srcNdx = 0;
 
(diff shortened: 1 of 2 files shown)


oddhack:
> @oddhack not sure where vulkan_json_parser.hpp and vulkan_json_data.hpp are used outside CTS. If you know of any users, I'd appreciate it if you could verify that this change is a no-op for them.

I believe these are Vulkan SC-specific. @dgkoch can you confirm and advise how to proceed? Putting it in the upstream like this seems OK if it's not an urgent fix.

dgkoch:
Mobica is actually about to update these in VK-GL-CTS, so they'll probably be interested in this fix. 
I wonder if there is more of the "isCTS" logic we can get rid of now that cts is no longer using the "de" types. 

I might cherry-pick this straight into the vulkansc repo, where we added another change recently.

scripts/json_generator.py
```diff
@@ -515,23 +515,10 @@ def beginFile(self, genOpts):
...
-                                  "char"      : "char"
-                                }
-
         write(headerGuardTop, file=self.outFile, end='')
         write(copyright, file=self.outFile)
         if self.isCTS:
-            write(predefinedCode % ("deUint32", "deUint32"), file=self.outFile)
+            write(predefinedCode % ("uint32_t", "uint32_t"), file=self.outFile)
         else:
             write(predefinedCode % ("uint32_t", "uint32_t"), file=self.outFile)
```
dgkoch, line 523: These two branches can just be collapsed now.

scripts/json_generator.py
```diff
@@ -840,7 +827,7 @@ def genArrayCode(self, structName, name, typeName, str2, arraySize, needStrPrint
                     code += "           print_%s(%s%s[i], tmp.str(), isCommaNeeded);\n" %(typeName, str2, name)
                 else:
                     if self.isCTS and name == "pipelineIdentifier":
-                        code += "           print_uint32_t((%s)%s%s[i], %s, isCommaNeeded);\n" %(self.baseTypeListMap["uint32_t"], str2, name, printStr)
+                        code += "           print_uint32_t((uint32_t)%s%s[i], %s, isCommaNeeded);\n" %(str2, name, printStr)
                     else:
                         code += "           print_%s(%s%s[i], %s, isCommaNeeded);\n" %(typeName, str2, name, printStr)
```
dgkoch, line 832: Suspect these cases might be the same now as well.

dgkoch: apparently this one still differs (uint8_t vs uint32_t)

dgkoch:
I've taken this to internal merge request: https://gitlab.khronos.org/vulkansc/vulkansc/-/merge_requests/382
Since we have some other recent changes there that haven't been merged out yet.

scripts/json_generator.py
ShabbyX: Thanks for fixing it, I added a small update to bake these in `predefinedCode`

oddhack:
> I've taken this to internal merge request: https://gitlab.khronos.org/vulkansc/vulkansc/-/merge_requests/382 Since we have some other recent changes there that haven't been merged out yet.

Taking this off the Vulkan maintenance agenda - please close when appropriate.

dgkoch:
Something based on this was merged into VKSC branch and released in the VKSC 1.0.16 release, and was merged back up to Vulkan and published in the 1.3.299 release.

PR #2402, opened 2024-08-01
Add location order to the definition of from-reads

hernanponcedeleon:
We are currently [discussing a problem](https://github.com/KhronosGroup/Vulkan-MemoryModel/issues/36) and a [potential fix](https://github.com/KhronosGroup/Vulkan-MemoryModel/pull/37) with the members of the Memory Model TSG. If the fix gets accepted, the specification also needs to be updated to reflect this change.

### DIFF
--- a/appendices/memorymodel.adoc
+++ b/appendices/memorymodel.adoc
@@ -1016,8 +1016,8 @@ value written by the first operation.
 _From-reads_ is a relation between operations, where the first operation is
 a read, the second operation is a write, and the first operation reads a
 value written earlier than the second operation in the second operation's
-scoped modification order (or the first operation reads from the initial
-value, and the second operation is any write to the same locations).
+scoped modification order or location order (or the first operation reads from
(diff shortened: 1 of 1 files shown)


Robert-Qualcomm:
Added to agenda for the next MM TSG meeting on 2024-08-29

jeffbolznv (approved):

jeffbolznv:
The memory model TSG signed off on this today. The Vulkan WG should sign off on it next week.

oddhack:
> The memory model TSG signed off on this today. The Vulkan WG should sign off on it next week.

I happened to notice this in passing, but for future reference, I sweep new issues and PRs/MRs each week and classify things for the maintenance call or for one of the TSGs. I don't revisit old issues that have previously been assigned to a TSG just to see if something changed, though. In that case either the TSG can just sign off if it's within their remit, or ping me (or bring it up on the teleconference) if the entire WG should also sign off. I will put this on the next maintenance agenda.

Issue #2404, opened 2024-08-13
Multiple equivalent require blocks could be merged


Waffl3x:
I am fairly certain the former can be replaced with the latter as the require blocks are equivalent, only differing in their depends attribute. I can open a pull request for this later if this change is desirable.
```
        <extension name="VK_KHR_push_descriptor" number="81" type="device" author="KHR" depends="VK_KHR_get_physical_device_properties2,VK_VERSION_1_1" contact="Jeff Bolz @jeffbolznv" supported="vulkan" ratified="vulkan">
            <require>
                <enum value="2"                                             name="VK_KHR_PUSH_DESCRIPTOR_SPEC_VERSION"/>
                <enum value="&quot;VK_KHR_push_descriptor&quot;"            name="VK_KHR_PUSH_DESCRIPTOR_EXTENSION_NAME"/>
                <enum offset="0" extends="VkStructureType"                  name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PUSH_DESCRIPTOR_PROPERTIES_KHR"/>
                <enum bitpos="0" extends="VkDescriptorSetLayoutCreateFlagBits"   name="VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR"  comment="Descriptors are pushed via flink:vkCmdPushDescriptorSetKHR"/>
                <command name="vkCmdPushDescriptorSetKHR"/>
                <type name="VkPhysicalDevicePushDescriptorPropertiesKHR"/>
            </require>
            <require depends="VK_VERSION_1_1">
                <command name="vkCmdPushDescriptorSetWithTemplateKHR"/>
                <enum value="1" extends="VkDescriptorUpdateTemplateType"    name="VK_DESCRIPTOR_UPDATE_TEMPLATE_TYPE_PUSH_DESCRIPTORS_KHR" comment="Create descriptor update template for pushed descriptor updates"/>
            </require>
            <require depends="VK_KHR_descriptor_update_template">
                <command name="vkCmdPushDescriptorSetWithTemplateKHR"/>
                <enum value="1" extends="VkDescriptorUpdateTemplateType"    name="VK_DESCRIPTOR_UPDATE_TEMPLATE_TYPE_PUSH_DESCRIPTORS_KHR" comment="Create descriptor update template for pushed descriptor updates"/>
            </require>
        </extension>
```

```
[...]

PR #2406, opened 2024-08-14
Merge equivalent require blocks

Waffl3x:
Addresses #2404
Merges extension require blocks that have identical children, extending the depends attribute with a logical or.

### DIFF
--- a/scripts/find_equivalent_ext_require_blocks.py
+++ b/scripts/find_equivalent_ext_require_blocks.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+#
+# Copyright 2024 Waffl3x
+# SPDX-License-Identifier: Apache-2.0
+
+import xml.etree.ElementTree as etree
+import itertools
+
+tree = etree.parse('vk.xml')
+doc = tree.getroot()
+
+extensions = doc.find('extensions')
+
+def check_if_children_equal(req1, req2):
+    if len(req1) != len(req2):
+        return False
+    for child1, child2 in zip(req1, req2):
+        if child1.tag != child2.tag:
+            return False
+        if child1.attrib != child2.attrib:
+            return False
+    return True
+# def check_if_children_equal
+
+for ext in extensions:
+    req_with_depends = []
+    # put all candidates into a list
+    for req, count in zip(ext, itertools.count()):
+        if 'depends' in req.attrib:
+            req_with_depends.append((req, count))
+    
+    # compare candidates in a pairwise manner
+    for tup1, it2_start in zip(req_with_depends[:-1], range(1, len(req_with_depends))):
+        for tup2 in req_with_depends[it2_start:]:
+            req1, req_count1 = tup1
+            req2, req_count2 = tup2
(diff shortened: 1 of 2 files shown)


Waffl3x:
Here is a script I made in python that finds these cases, I'm not really sure how to handle license stuff, and I'm not sure where it should go so I didn't add it with this pull request. I'm also not very familiar with python so it has some flaws, namely that it just looks for a vk.xml in the same directory as the script, so I don't feel like it's good enough to be included as-is.
```python
import xml.etree.ElementTree as etree
import itertools

tree = etree.parse('vk.xml')
doc = tree.getroot()

extensions = doc.find('extensions')

def check_if_children_equal(req1, req2):
    if len(req1) != len(req2):
        return False
    for child1, child2 in zip(req1, req2):
        if child1.tag != child2.tag:
            return False
        if child1.attrib != child2.attrib:
            return False
    return True
# def check_if_children_equal

for ext in extensions:
    req_with_depends = []
    # put all candidates into a list
    for req, count in zip(ext, itertools.count()):
        if 'depends' in req.attrib:
            req_with_depends.append((req, count))
    
    # compare candidates in a pairwise manner
    for tup1, it2_start in zip(req_with_depends[:-1], range(1, len(req_with_depends))):
        for tup2 in req_with_depends[it2_start:]:
            req1, req_count1 = tup1
            req2, req_count2 = tup2
            all_match = check_if_children_equal(req1, req2)
            if all_match:
                print('Found matching require block in extension {}:'.format(ext.attrib['name']))
                print('blocks {} and {} are equal'.format(req_count1, req_count2))
                print('require block {} attributes: {}'.format(req_count1, req1.attrib))
                print('require block {} attributes: {}'.format(req_count2, req2.attrib))
                print('  block children:')
                for child1, child2 in zip(req1, req2):
                    print('    ', req_count1, ': ', child1.attrib)
                    print('    ', req_count2, ': ', child2.attrib)
```

oddhack:
Thanks! If you are comfortable putting this at the start:

```
#!/usr/bin/env python3
#
# Copyright 2024 <your actual name, or other organization holding copyright>
# SPDX-License-Identifier: Apache-2.0
```

in the header, that will place it under Apache 2.0 on the same terms as other scripts in the repo and allow us to redistribute it while still giving you credit.

Waffl3x:
Should be good now I think

scripts/find_equivalent_ext_require_blocks.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
+#
+# Copyright 2024 <your actual name, or other organization holding copyright>
```
oddhack, line 3: Please replace "your actual name, or other organization holding copyright" with, well, what it says :-)

Waffl3x: I believe my pseudonym should be fine, I contributed to GCC under this name, the Free Software Foundation has the paperwork for that. I don't know if that's compatible or whatnot, but hopefully it's fine. If not I guess I can use my legal name but I would like to avoid that if possible.

Issue #2307, opened 2024-02-13
Color management on the Vulkan Wayland platform


ppaalanen:
We are developing [a color-management Wayland protocol extension](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/14), and expect Vulkan WSI implementations to use it on behalf of the application, because Vulkan already defines the extensions to do so. We need to [define the rules](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/14#note_2276149) how that works in `VK_KHR_wayland_surface`.

The main driver of color-management information is `VkColorSpaceKHR` enumeration. Other Vulkan extension may add more, e.g. `VK_EXT_hdr_metadata`. These values shall be converted into Wayland color-management extension protocol by a WSI.

We need to define how exactly the protocol values are derived from Vulkan values. We also need to define when Vulkan WSI shall not instantiate a `wp_color_management_surface_v1` for a `wl_surface`. The latter is needed to allow applications drive color-management protocol themselves, because the protocol allows things not exposed in Vulkan API. The use of ICC profile files is one such feature.

Presumably `VK_COLOR_SPACE_PASS_THROUGH_EXT` would be the value to guarantee that no `wp_color_management_surface_v1` is created.

How to handle `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` is an open question. There are two possibilities:
- Define that color-management communicates explicit sRGB primaries, white point, SDR, and transfer function.
- Define that no `wp_color_management_surface_v1` is created, leaving the Wayland image description unspecified.

By Vulkan definition, `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` is sRGB. However, window systems have probably relayed such pixel data as-is to all (SDR) displays, meaning that its presentation may not have been sRGB. If its presentation is now changed to sRGB, it might appear visually different. On the other hand, if no image description is set for `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR`, Wayland compositors may continue presenting the content like they always have, or the presentation may differ from previous, and the presentation may differ from content that is explicitly sRGB. All this is up to the policy of each Wayland compositor implementation, what their users want.

We need to document how Vulkan WSI must use Wayland color-management extension.

versalinyaa:
[...]

PR #2410, opened 2024-08-16
Add VK_EXT_disable_wayland_color_management

colinmarc:
### VK_EXT_disable_wayland_color_management

This adds a small extension, allowing applications to request that WSI implementations avoid using wayland color management, so that they can do it themselves. This addresses #2307.

## Background

Wayland is growing color management and HDR support, in the soon-to-be-finalized `wp_color_management` protocol[^1]. The protocol has support for tagging surfaces with colorspace information and HDR metadata, generally in ways matching up to Vulkan's [`VK_KHR_swapchain_colorspace`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_swapchain_colorspace.html) and [`VK_EXT_hdr_metadata`](https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/VK_EXT_hdr_metadata.html), but also with facilities that are not currently expressible via Vulkan or other graphics APIs, such as using an ICC profile to describe a surface's content.

As such, applications wanting to do color management or output HDR content on Wayland will roughly fall into three buckets:

 1. Image viewers and other utility apps that use Wayland directly, but not Vulkan or other graphics APIs
 2. Games, video players, and the like, which will use Vulkan or other graphics APIs, but not the Wayland protocol directly
 3. Content creation apps, which are picky about color management and wish to use the Wayland protocol directly, but also want to use Vulkan/etc to submit GPU buffers

The protocol's design is such that only one Wayland client may do color management for a particular surface at once. Specifically, only one `color_management_surface` may exist for a given Wayland surface at a time. Applications in the first group will create this object themselves, while applications in the second group will use `VK_KHR_swapchain_colorspace` and other related extensions, and rely on WSI implementations to do the tagging for them. The third group wants to create the object themselves, but needs to prevent the WSI implementation from also creating the object, since that would be a protocol violation. Hence this extension.

### 1. Check whether `VK_KHR_swapchain_colorspace` is enabled

WSI implementations could check whether the application has enabled the extension, and only create `color_management_surface` objects if it is. This has a few problems:

 - It uses an instance extension to signal an option that really should be per-surface, changing the behavior of all surfaces created from the same instance.
 - Because the extension is disabled, the application has no access to `VK_COLOR_SPACE_PASS_THROUGH_EXT`, which matches the behavior they want, and must use `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR`, which may be completely unrelated to the content of their surface, to mean that the surface should remain untagged.

### 2. Switch on `VK_COLOR_SPACE_PASS_THROUGH_EXT`

This is more intuitively correct. Applications that set `PASS_THROUGH` would seem to be indicating that they want the surface to be untagged. However, this is difficult to implement. By the time the swapchain is create, the wayland surface has already been created, which means the driver must delay deciding whether to create the `color_management_surface` object. Furthermore, if another swapchain is created with a different colorspace, the `color_management_surface` must be created at that point, and then destroyed if the colorspace is again set to `PASS_THROUGH`. All of those operations could potentially race with the application's creation or destruction of the `color_management_surface` object, and such shenanigans would be pretty error-prone in general.

## Note

I would like to create a validation check to ensure that surfaces created with `VK_WAYLAND_SURFACE_CREATE_DISABLE_COLOR_MANAGEMENT` must only be used to create swapchains with `VK_COLOR_SPACE_PASS_THROUGH_EXT`. Should that go in this PR, or somewhere else? Should the validation be on `vkCreateSwapchainKHR` or on `VkWaylandCreateSurafceKHR`?

[^1]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/14

### DIFF
--- a/appendices/VK_EXT_disable_wayland_color_management.adoc
+++ b/appendices/VK_EXT_disable_wayland_color_management.adoc
@@ -1,0 +1,31 @@
+// Copyright 2016-2024 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_EXT_disable_wayland_color_management.adoc[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2024-08-16
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+    Colin Marc
+
+=== Description
+
+This extension expands tlink:VkWaylandSurfaceCreateFlagsKHR to allow
+applications to request that implementations avoid using the
+`wp_color_management` protocol when creating or managing the surface.
+
+include::{generated}/interfaces/VK_EXT_disable_wayland_color_management.adoc[]
+
+=== Issues
+
+None.
+
+=== Version History
+
+  * Revision 1, 2024-08-16 (Colin Marc)
+  ** Initial draft
--- a/appendices/VK_KHR_wayland_surface.adoc
+++ b/appendices/VK_KHR_wayland_surface.adoc
@@ -90,3 +90,6 @@ an onerous restriction on application developers.
      ename:VK_PRESENT_MODE_MAILBOX_KHR.
   ** Added wording about interactions between flink:vkQueuePresentKHR and
      the Wayland requests sent to the compositor.
+
+  * Revision 7, 2024-08-16 (Colin Marc)
+  ** Added VkWaylandSurfaceCreateFlagBits.
--- a/chapters/VK_KHR_swapchain/wsi.adoc
+++ b/chapters/VK_KHR_swapchain/wsi.adoc
@@ -418,6 +418,14 @@ endif::VK_KHR_shared_presentable_image[]
     and pname:colorSpace members, respectively, of one of the
     sname:VkSurfaceFormatKHR structures returned by
     fname:vkGetPhysicalDeviceSurfaceFormatsKHR for the surface
+ifdef::VK_EXT_disable_wayland_color_management[]
+  * [[VUID-VkSwapchainCreateInfoKHR-imageColorSpace-00000]]
+    If pname:surface was created using flink:vkCreateWaylandSurfaceKHR and
+    sname:VkWaylandSurfaceCreateInfoKHR::pname:flags containing
+    ename:VK_WAYLAND_SURFACE_CREATE_DISABLE_COLOR_MANAGEMENT_BIT_EXT, then
+    sname:VkSwapchainCreateInfoKHR:pname:imageColorSpace must: be
+    ename:VK_COLOR_SPACE_PASS_THROUGH_EXT
+endif::VK_EXT_disable_wayland_color_management[]
   * [[VUID-VkSwapchainCreateInfoKHR-pNext-07781]]
 ifdef::VK_EXT_swapchain_maintenance1[]
     If a slink:VkSwapchainPresentScalingCreateInfoEXT structure was not
--- a/chapters/VK_KHR_wayland_surface/platformCreateSurface_wayland.adoc
+++ b/chapters/VK_KHR_wayland_surface/platformCreateSurface_wayland.adoc
@@ -35,7 +35,8 @@ include::{generated}/api/structs/VkWaylandSurfaceCreateInfoKHR.adoc[]
   * pname:sType is a elink:VkStructureType value identifying this structure.
   * pname:pNext is `NULL` or a pointer to a structure extending this
     structure.
-  * pname:flags is reserved for future use.
+  * pname:flags is a bitmask of elink:VkWaylandSurfaceCreateFlagBitsKHR
+    specifying additional surface creation parameters.
   * pname:display and pname:surface are pointers to the Wayland
     code:wl_display and code:wl_surface to associate the surface with.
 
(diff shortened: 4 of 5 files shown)


oddhack:
@colinmarc process-wise, we require that extension number and flag bits be reserved by a commit to main prior to the extension development branch actually being able to use them. I don't know where this extension stands deployment / release-wise, but we need to separate out the reservations and accept those first, to avoid a race condition (see extension 597 for an guide to what that PR would look like). In face because #2414 also just came up, we will have to tweak one or the other of those reservation PRs since you can't both have #603.

oddhack:
If you will go ahead and create the reservation PR using extension number 604, since #2416 just reserved 603, we should be good now (at least if nobody does an internal extension reservation MR first :-)).

r-potter:
> I would like to create a validation check to ensure that surfaces created with VK_WAYLAND_SURFACE_CREATE_DISABLE_COLOR_MANAGEMENT must only be used to create swapchains with VK_COLOR_SPACE_PASS_THROUGH_EXT. Should that go in this PR, or somewhere else? Should the validation be on vkCreateSwapchainKHR or on VkWaylandCreateSurafceKHR?

The Valid Usage statement should go in this PR and be on the VkWaylandCreateSurfaceKHR structure, next to the other two VUIDs. Just write it as an undecorated bullet point, and the actual VUID tag will get assigned as part of the spec merge process. 

Implementation of the check itself would then follow as a PR against https://github.com/KhronosGroup/Vulkan-validationlayers

colinmarc:
Thanks all for the help! I reserved the extension in https://github.com/KhronosGroup/Vulkan-Docs/pull/2419. I also rebased and added the validation check.

linyaa-kiwi:
cc @swick

swick:
The wayland protocol now supports has two surface extensions. `wp_color_management_feedback_surface_v1` can be instantiated multiple times and provides everything that's required until a swapchain is presenting with a VkColorSpaceKHR that's not `VK_COLOR_SPACE_PASS_THROUGH_EXT` or `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR`.

I'm not convinced this extension is necessary.

colinmarc:
@swick what mechanism should a Vulkan implementation use to determine whether or not to create a `wp_color_management_surface_v1`? I don't see how the feedback object helps with that. Or do you mean that the Vulkan implementation should always create one, and apps (wanting to use Vulkan) should never do so?

swick:
I'm suggesting that you only need to create a `wp_color_management_surface_v1` when a swapchain with `VkColorSpace != VK_COLOR_SPACE_PASS_THROUGH_EXT` is actually presenting. Just the existence of such a swapchain is irrelevant. Equally, if a swaptchain with `VkColorSpace == VK_COLOR_SPACE_PASS_THROUGH_EXT` is presenting, then there should be no `wp_color_management_surface_v1`.

colinmarc:
That sounds a lot like alternative 2., which I outlined above, except you're suggesting delaying the creation of the surface until presentation, rather than delaying it until swapchain presentation time. I don't think that offers any particular advantages over lazily creating it along with the swapchain, since the colorspace is known at swapchain creation time, and it's awkward for the implementation to wait until `VkQueuePresentKHR`. 

That solution suffers from the race conditions I mentioned - for example, what happens if you create a swapchain with `HDR10` or something and then destroy it and create another with `PASSTHROUGH? Should the `wp_color_management_surface_v1` be destroyed at that point? Note that swapchain destruction is usually asynchronous and follows creation and presentation on the new swapchain in most apps.

swick:
> I don't think that offers any particular advantages over lazily creating it along with the swapchain, since the colorspace is known at swapchain creation time, and it's awkward for the implementation to wait until VkQueuePresentKHR.

You could ref count the `wp_color_management_surface_v1` and when no swapchain with `VkColorSpace != VK_COLOR_SPACE_PASS_THROUGH_EXT` exists you destroy it. Then it's up to the app to make sure that all relevant swapchains have been destroyed before creating a `wp_color_management_surface_v1` itself. In practice, apps would probably never create a swapchain with `VkColorSpace != VK_COLOR_SPACE_PASS_THROUGH_EXT` on a surface they want to use `wp_color_management_surface_v1` on.

colinmarc:
> You could ref count the `wp_color_management_surface_v1` and when no swapchain with `VkColorSpace != VK_COLOR_SPACE_PASS_THROUGH_EXT` exists you destroy it. Then it's up to the app to make sure that all relevant swapchains have been destroyed before creating a `wp_color_management_surface_v1` itself.

Perhaps, but then apps wouldn't be able to tag their surface until after the first few frames are presented on the new swapchain, which would cause flickering. And swapchain destruction happens in a separate thread in a lot of apps, which could race.

> In practice, apps would probably never create a swapchain with VkColorSpace != VK_COLOR_SPACE_PASS_THROUGH_EXT on a surface they want to use wp_color_management_surface_v1 on.

I agree, so why not expressly and explicitly forbid it? That's what this extension does.

You're proposing a potential mechanism, but what Vulkan requires is policy. Formulating your suggested mechanism as policy would be difficult to understand or enforce via validation. It would be something like 

> implementations must only create a `wp_color_management_surface_v1` if there exists an extant swapchain with a colorspace value other than `VK_COLOR_SPACE_PASS_THROUGH_EXT`, and destroy it as soon as, for a given surface, there is no extant swapchain with a colorspace value other than  `VK_COLOR_SPACE_PASS_THROUGH_EXT`.

Compare that to these other *policy* alternatives:

 > "Applications must not create `wp_color_management_surface_v1` objects if they are using Vulkan"

(for the record, I prefer this option). or: 

 > "Implementations must not create `wp_color_management_surface_v1` objects if `VK_WAYLAND_SURFACE_CREATE_DISABLE_COLOR_MANAGEMENT` is set on the surface"

I find both of those much more understandable and enforceable.

Zamundaaa:
Ref counting non-`VK_COLOR_SPACE_PASS_THROUGH_EXT` swapchains sounds acceptable to me as well, though I like that this extension is more explicit about its purpose. In practice the difference probably doesn't matter a lot, apps that do color management themselves likely only have one swapchain per surface anyways.

ppaalanen:
I have a small complaint about the nomenclature here. I find "disable color management" misleading. In this proposal it means "do not use color management protocol", but I believe people will intuitively understand it as "push pixel values unmodified to display". The documentation is clear, though.

How about something like VK_EXT_wayland_no_color_description? no_image_description would be straight from the Wayland extension, but I suspect "image description" could be easily confused with unrelated Vulkan terminology.

As for whether this extension as a whole is a good idea, I cannot say. I'm not familiar with Vulkan.

swick:
Apparently `VK_COLOR_SPACE_PASS_THROUGH_EXT` is not very well defined. Multiple people had different interpretations of what it means.

We would have to work on defining it better and probably also explicitly define the interaction with the wayland color management protocol, or come up with a new VkColorSpace for this purpose.

jreck:
> Content creation apps, which are picky about color management and wish to use the Wayland protocol directly, but also want to use Vulkan/etc to submit GPU buffers

I don't see it listed as an alternative so apologies if this is already considered & rejected, but Android's approach to this is to just say that those apps are beyond the scope of VK swapchain entirely. We strive to ensure that what the platform-provided Vk swapchain implementation can do is also doable directly with extensions such as VK_ANDROID_external_memory_android_hardware_buffer.

So applications that wish to use our presentation engine's protocol directly can do so, including buffer allocation, present handling, etc... This is how applications such as Chrome can also have cross-surface atomic presents, which is also something that VK swapchain doesn't support.

swick:
> Android's approach to this is to just say that those apps are beyond the scope of VK swapchain entirely

That's a fair stance to have, but I do believe that the VK swapchain contains a fair deal of non-trivial code for wayland. It would be unfortunate if one had to re-implement all of that just to use the more advanced color management features.

cubanismo:
The latest Vulkan spec added language which clarified that Vulkan implementations are not allowed to use the wayland color management mechanisms when there are no swapchains referencing a given surface, or when all swapchains referencing it use the colorspace `VK_COLOR_SPACE_PASS_THROUGH_EXT`.  Hard to link to exactly, but you can scroll down a bit from here to find the normative text: https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VkWaylandSurfaceCreateInfoKHR

@colinmarc, is that sufficient to satisfy your use case? If so, can we close this MR?

colinmarc:
Yes - just to confirm I understand, this means that in practice implementations create the color surface lazily, and don't create it lazily if the swapchain has PASS_THROUGH? 

I believe that works around the issue as I framed it.

swick:
> this means that in practice implementations create the color surface lazily, and don't create it lazily if the swapchain has PASS_THROUGH?

Yes. That has always been the case for mesa and is now required by the spec.

PR #2414, opened 2024-08-20
Add VK_EXT_external_memory_metal

aitor-lunarg:


oddhack:
@aitor-lunarg process-wise, we require that extension number and flag bits be reserved by a commit to main prior to the extension development branch actually being able to use them. I don't know where this extension stands deployment / release-wise, but we need to separate out the reservations and accept those first, to avoid a race condition (see extension 597 for an guide to what that PR would look like).

aitor-lunarg:
Before moving forward with this extension, there's one concern that we need to address. Metal does not provide all Vulkan formats, so there are times where an implementation will emulate some of them. Why is this an issue:

Implementations may decide to emulate those formats with an internal representation such that those formats are backed by multiple textures. Taking MoltenVK as an example implementation with `VK_FORMAT_G8_B8_R8_3PLANE_420_UNORM`. This format is backed by 3 textures (one for each plane), so it is not possible to export all of the backing textures through a single handle unfortunately as of now.

There are 3 potential solutions that come to mind to fix this issue as we stand:
1. Force implementation to have those multi-planar emulated implementations in a single texture if they want to export it. Unsure if there could be issues with precision when sampling that would prevent this approach.
2. Modify the API for this extension to allow exporting more than one single handle for these cases. This would lead to a query call first to understand how many handles the user will get, and then get the handles. However, this is quite counter intuitive since we imply that a `VkDeviceMemory` object may not be a single handle. When importing we would also need to provide all handles as the implementation expects them. I don't really like this approach.
3. Modify the handle to `MTLHeap` and force implementations to have all resources stored in those objects if they want to export them. Again, this also has issues since not all Apple hardware may support `MTLHeap` and there are a few known restrictions that make it ugly for non-Apple silicon GPUs.

@linyaa-kiwi would it be possible to bring this up for discussion this coming SI meeting? I would like to understand if other similar extensions have run into similar issues and how they've tackled them if so. Or just get general feedback/ideas about this issue.

cc/ @kocdemir

linyaa-kiwi:
> @linyaa-kiwi would it be possible to bring this up for discussion this coming SI meeting? I would like to understand if other similar extensions have run into similar issues and how they've tackled them if so. Or just get general feedback/ideas about this issue.

@aitor-lunarg Would you like to discuss in today's SI meeting? If not, what week works best for you?

proposals/VK_EXT_external_memory_metal.adoc
```diff
@@ -0,0 +1,110 @@
...
+extensions like link:{refpage}VK_KHR_external_memory_win32.html[VK_KHR_external_memory_win32] and
+link:{refpage}VK_ANDROID_external_memory_android_hardware_buffer.html[VK_ANDROID_external_memory_android_hardware_buffer] to name a few.
+
+This means that external Metal resources will be imported through link:{refpage}VkDeviceMemory.html[VkDeviceMemory] to later be bound.
+
+3 new bit values will be added to link:{refpage}VkExternalMemoryHandleTypeFlagBits.html[VkExternalMemoryHandleTypeFlagBits]:
+
+ - `VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLBUFFER_BIT_EXT` to accommodate for Metal buffers (MTLBuffer)
+ - `VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT` to accommodate for Metal textures (MTLTexture)
+ - `VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLHEAP_BIT_EXT` to accommodate for Metal textures (MTLHeap)
```
kocdemir, line 41: typo: 'Metal textures' instead of heaps.

chapters/VK_EXT_external_memory_metal/device_memory.adoc
```diff
@@ -0,0 +1,153 @@
...
+  * pname:device is the logical device that created the device memory being
+    exported.
+  * pname:pGetMetalHandleInfo is a pointer to a
+    slink:VkMemoryGetMetalHandleInfoEXT structure containing parameters of
+    the export operation.
+  * pname:pHandle will return the Metal handle representing the payload of
+    the device memory object.
+
+The lifespan of the returned handle object is the same as the associated `VkDeviceMemory` object.
+The app must: retain the handle object if it intends to use it beyond that lifespan.
```
linyaa-kiwi, line 70: I feel that the lifetime language is incorrect, unless I misunderstand the meaning of "retain". The text says that the lifespan of the Metal handle is the same as the VkDeviceMemory; therefore, destruction of the VkDeviceMemory also invalidates the Metal handle. However, the text's sentence suggests that the lifetime can be extended by "retaining" the handle; this sentence contradicts the first sentence.

chapters/capabilities.adoc
```diff
@@ -568,6 +576,11 @@ endif::VK_NV_external_memory_sci_buf[]
 ifdef::VK_QNX_external_memory_screen_buffer[]
 | ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_SCREEN_BUFFER_BIT_QNX | No restriction | No restriction
 endif::VK_QNX_external_memory_screen_buffer[]
+ifdef::VK_EXT_external_memory_metal[]
+| ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLBUFFER_BIT_EXT | Must match | Must match
+| ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT | Must match | Must match
+| ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLHEAP_BIT_EXT | Must match | Must match
+endif::VK_EXT_external_memory_metal[]
```
linyaa-kiwi, line 583: I'm concerned about the choice of "Must match" in this table. I suspect that the constraint is impossible to satisfy in the extension's intended use cases.

When sharing a Metal handle with a different VkDevice in a different VkInstance, the "must match" constraint is easy to satisfy. That's not my concern.

I assume that an intended use case for the extension is to share Metal handles *outside* of Vulkan. (Please correct me if I'm wrong). In that case, do the relevant sharing apis expose the equivalent of `driverUUID`, and does that UUID match the Vulkan driver's `driverUUID`?

kocdemir: > I assume that an intended use case for the extension is to share Metal handles outside of Vulkan.

For the Android Emulator, sharing outside of the Vulkan is not a priority, as we won't be sharing the handle with OpenGL on Mac, and we share the buffers and images only between different Vulkan instances and devices.

chapters/VK_EXT_external_memory_metal/device_memory.adoc
aitor-lunarg: Metal objects are reference counted and only destroyed when the count reaches 0. This part of the spec is trying to state that unless the user calls `retain` (which increases the count) on the returned Metal object, the lifespan will be that of the `VkDeviceMemory`. That is why `beyond that lifespan` is used.

I will try to think of a better way to word this part since I see it can lead to confusion. If you have any suggestion on how to word it, do let me know!

chapters/capabilities.adoc
aitor-lunarg: Maybe changing `driverUUID` here to `No restriction` should be the way forward since Metal does not have an equivalent. The only requirement should be that the `deviceUUID` matches, and Metal does provide an equivalent.

chapters/memory.adoc
```diff
@@ -1709,6 +1720,15 @@ ifdef::VK_EXT_metal_objects[]
     pname:exportObjectType member must: be
     ename:VK_EXPORT_METAL_OBJECT_TYPE_METAL_BUFFER_BIT_EXT
 endif::VK_EXT_metal_objects[]
+ifdef::VK_EXT_external_memory_metal[]
+  * If the parameters define an import operation and the external handle is
+    a ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT, then
+    pname:pNext must: include a slink:VkMemoryDedicatedAllocateInfo.
+  * If the parameters define an import operation and the external handle is
+    a ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT, then
```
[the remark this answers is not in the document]
aitor-lunarg: This should be the VU for `VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLHEAP_BIT_EXT`, but now that I take a second look at it, I don't see why it can't be merged to the one above

chapters/VK_EXT_external_memory_metal/device_memory.adoc
aitor-lunarg: Let me know if it's clearer now!

proposals/VK_EXT_external_memory_metal.adoc
aitor-lunarg: Done

chapters/memory.adoc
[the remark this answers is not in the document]
aitor-lunarg: Merged into above VU

chapters/capabilities.adoc
linyaa-kiwi: Based on today's discussion in the SI meeting, this proposal sgtm. No requirement on driverUUID, but require matching deviceUUID.

appendices/VK_EXT_external_memory_metal.adoc
```diff
@@ -0,0 +1,27 @@
...
+    2024-07-18
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Aitor Camacho Larrondo, LunarG Inc.
+
+=== Description
+
+An application may wish to reference device memory in multiple Vulkan device instances, in multiple processes, and/or in Metal API.
+This extension enables an application to export non-Vulkan handles from Vulkan memory objects such that the underlying resources can
```
linyaa-kiwi, line 19: s/non-Vulkan handles/Metal handles/

linyaa-kiwi: The text says "export", but the extension supports import too.

chapters/capabilities.adoc
```diff
@@ -707,6 +720,9 @@ ifdef::VK_QNX_external_memory_screen_buffer[]
   * ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_SCREEN_BUFFER_BIT_QNX for images
     only
 endif::VK_QNX_external_memory_screen_buffer[]
+ifdef::VK_EXT_external_memory_metal[]
+  * ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT
```
linyaa-kiwi, line 724: Elsewhere in the MR, a VU for VkMemoryAllocateInfo says that VkDedicatedAllocationInfo must be used if the handle is MTLTEXTURE or MTLHEAP. Should MTLHEAP also be listed here? Should MTLHEAP be dropped from the VU?

chapters/memory.adoc
```diff
@@ -1712,6 +1723,12 @@ ifdef::VK_EXT_metal_objects[]
     pname:exportObjectType member must: be
     ename:VK_EXPORT_METAL_OBJECT_TYPE_METAL_BUFFER_BIT_EXT
 endif::VK_EXT_metal_objects[]
+ifdef::VK_EXT_external_memory_metal[]
+  * If the parameters define an import operation and the external handle is
+    a ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT or
+    ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLHEAP_BIT_EXT, then
```
linyaa-kiwi, line 1729: For MTLHEAP, which member of VkMemoryDedicatedAllocateInfo is used? The image or the buffer? This VU should explicitly state which field of VkMemoryDedicatedAllocateInfo must be non-null.

proposals/VK_EXT_external_memory_metal.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+ - `VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLBUFFER_BIT_EXT` to accommodate for Metal buffers (MTLBuffer)
+ - `VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT` to accommodate for Metal textures (MTLTexture)
+ - `VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLHEAP_BIT_EXT` to accommodate for Metal heaps (MTLHeap)
+
+Due to MTLTexture's semantics in Metal API roughly align with that of an image with a dedicated allocation in Vulkan,
+implementations will be required to report `VK_EXTERNAL_MEMORY_FEATURE_DEDICATED_ONLY_BIT` for
+`VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT`. When importing a MTLTexture through the creation of a
+link:{refpage}VkDeviceMemory.html[VkDeviceMemory], the link:{refpage}VkImage.html[VkImage] it will be bound to
+must have the same creation parameters as the MTLTexture used at import.
```
linyaa-kiwi, line 47: "must have same creation parameters as the MTLTexture used at import"

This needs a VU. A good example is VUID-VkMemoryAllocateInfo-pNext-02388, which is about AHardwareBuffer. There are more VUs in that neighborhood that are worth considering too. The new VU (new VUs) should enumerate which fields of VkImageCreateInfo must match the MTLTexture's creation params.

linyaa-kiwi (commented):
I found some more points for improvement today.

appendices/VK_EXT_external_memory_metal.adoc
aitor-lunarg: Let me know if the new wording works!

chapters/capabilities.adoc
aitor-lunarg: I just removed MTLHEAP from the VU since that was an oversight on my end. MTLHeaps can create resources (buffers and textures) so there's no need to make them dedicated. MTLTextures, on the other hand, do not allow to create buffers that point to the same address as long as they were created as an stand alone MTLTexture with no MTLHeap (this is why the MTLTEXTURE bit exists).

Thank you for catching this!

chapters/memory.adoc
aitor-lunarg: Removed MTLHEAP, check comment above

proposals/VK_EXT_external_memory_metal.adoc
aitor-lunarg: Added a new VU to handle this case. I am thinking that maybe we could add a few more to force the users to provide the right size at allocation when importing buffers or heaps, but it kind of seems slightly pointless. Right now, we just say we don't really care about the size for buffers and heaps since we will use Metal to know the size of said resources. Any way, let me know your thoughts!

appendices/VK_EXT_external_memory_metal.adoc
linyaa-kiwi: So close! "Import" is misspelled as "impor". Other than the typo, the text looks good.

chapters/VK_EXT_external_memory_metal/device_memory.adoc
```diff
@@ -0,0 +1,153 @@
...
+properties of the defined external memory handle types.
+
+.Valid Usage
+****
+  * pname:memory must: have been created with a valid
+    slink:VkExportMemoryAllocateInfo
+  * pname:handleType must: have been included in
+    slink:VkExportMemoryAllocateInfo::pname:handleTypes when pname:memory
+    was created
+  * pname:handleType it must: be
```
linyaa-kiwi, line 101: `s/pname:handleType it must: be/pname:handleType must: be/`

chapters/VK_EXT_external_memory_metal/device_memory.adoc
```diff
@@ -0,0 +1,153 @@
...
+  * pname:handle is the handle which will be imported.
+  * pname:pMemoryMetalHandleProperties is a pointer to a
+    slink:VkMemoryMetalHandlePropertiesEXT structure in which properties of
+    pname:handle are returned.
+
+.Valid Usage
+****
+  * pname:handle must: point to a valid id<MTLBuffer>, id<MTLTexture> or
+    id<MTLDevice>
+  * pname:handleType it must: be
```
linyaa-kiwi, line 131: `s/pname:handleType it must: be/pname:handleType must: be/`

chapters/memory.adoc
```diff
@@ -1712,6 +1731,11 @@ ifdef::VK_EXT_metal_objects[]
     pname:exportObjectType member must: be
     ename:VK_EXPORT_METAL_OBJECT_TYPE_METAL_BUFFER_BIT_EXT
 endif::VK_EXT_metal_objects[]
+ifdef::VK_EXT_external_memory_metal[]
+  * If the parameters define an import operation and the external handle is
+    a ename:VK_EXTERNAL_MEMORY_HANDLE_TYPE_MTLTEXTURE_BIT_EXT, then
+    pname:pNext must: include a slink:VkMemoryDedicatedAllocateInfo.
```
linyaa-kiwi, line 1737: This VU needs a little more text.

> [...] must: include a slink:VkMemoryDedicatedAllocateInfo structure with pname:image not equal to dlink:VK_NULL_HANDLE.

proposals/VK_EXT_external_memory_metal.adoc
linyaa-kiwi: I don't have a strong opinion on whether the spec should (a) require `size` to be correct or (b) ignore `size`. That's a decision is yours. In my opinion, the only practical difference is that choice (b) inhibits some VVL checks.

Regardless of the choice of (a) or (b) though for MTLBuffer and MTLHeap, more spec text is needed. If (a), then a VU is needed (similar to VUID-VkMemoryAllocateInfo-allocationSize-00647 or VUID-VkMemoryAllocateInfo-allocationSize-02383). If (b), then text similar to the MTLTexture "ignore" text is needed.

However, if you do choose (b), I prefer that the value not be actually ignored. In the past two years, several members of the Vulkan WG have become frustrated with the unintended consequences caused by ignored values in the spec. Superficially, it seems that an ignored `allocationSize` would cause no problems. But sometimes it can cause surprising difficulties for tools and layers. Instead of saying that `allocationSize` is ignored, I think it's better to add a VU that requires it to be 0.

aitor-lunarg: I went (b), added a VUID to enforce `allocationSize` to be `0` when importing MTLTextures so there's no issue in the future. Let me know what you think!

chapters/VK_EXT_external_memory_metal/device_memory.adoc
aitor-lunarg: Done

chapters/VK_EXT_external_memory_metal/device_memory.adoc
aitor-lunarg: Done

chapters/memory.adoc
aitor-lunarg: Added some more context. Let me know if that is clear enough or if a different wording is needed!

linyaa-kiwi:
Thanks. I'll give it another review today.

appendices/VK_EXT_external_memory_metal.adoc
aitor-lunarg: Forgot about this, but just pushed the fix!

linyaa-kiwi:
Thanks for the fixes! LGTM and I approve for merge.

aitor-lunarg:
Rebased to get Vulkan 1.4 so things don't explode when I propagate to VVL and CTS for release

oddhack:
@linyaa-kiwi @aitor-lunarg on the maintenance call, we were unsure about where this sits for release - do we want to ratify it first? Please loop in @r-potter but I'm expecting this will not go out this week, at least.

PR #2416, opened 2024-08-21
Reserve extension 603

aitor-lunarg:
@oddhack  As per https://github.com/KhronosGroup/Vulkan-Docs/pull/2414#issuecomment-2301342743, here's the required extension reserve PR. I have no preference on the number, I can wait until https://github.com/KhronosGroup/Vulkan-Docs/pull/2410 reserves 603 and I can reserve 604. If having this PR in the internal end makes things easier for you, let me know and I will do so!


oddhack:
Thanks. I'll propagate it internally momentarily. It is a little more convenient when we start in gitlab but it's not a big deal either way, and this is fine.

PR #2427, opened 2024-08-31
Added proposal: VK_KHR_indexed_vkformat

J-Vernay:
Hello,

This proposal adds new enumeration values to VkFormat, representing indexed/paletted bitmaps.

Have a nice day,
Julien Vernay.

### DIFF
--- a/proposals/VK_KHR_indexed_vkformat.adoc
+++ b/proposals/VK_KHR_indexed_vkformat.adoc
@@ -1,0 +1,88 @@
+// Copyright 2021-2024 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+= Proposal Template
+:toc: left
+:refpage: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/
+:sectnums:
+
+
+The link:{refpage}VkFormat.html[VkFormat] enumeration is extended to support indexed bitmap formats. Implementers are not required to support them.
+
+== Problem Statement
+
+Images are processed in a variety of formats, optimized for different needs: manipulation, storage, random-access… Most image processing software handle multiple formats, and thus need a way to disambiguate a given array of bytes. Vulkan uses the VkFormat enumeration for this purpose. Other libraries, platforms and middleware also expose their own different enumerations.
+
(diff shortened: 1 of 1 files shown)


oddhack:
We are definitely not going to implement this as a Vulkan extension, so the proposal document is not needed. However, treating it as an issue of reserving format enums for use only with KTX, which is how we think this is intended, they can be added to the existing (disabled) extension 289 in vk.xml which is already used for this purpose. If @MarkCallow as the KTX maintainer agrees this needs to be done, you or Mark can propose a PR against vk.xml to that effect.

MarkCallow:
> We are definitely not going to implement this as a Vulkan extension, so the proposal document is not needed. However, treating it as an issue of reserving format enums for use only with KTX, which is how we think this is intended, they can be added to the existing (disabled) extension 289 in vk.xml which is already used for this purpose. If @MarkCallow as the KTX maintainer agrees this needs to be done, you or Mark can propose a PR against vk.xml to that effect. I suggest closing this PR and opening an issue which I will assign to Mark to deal with.

Do not add anything to extension 289. That the ASTC 3D extension. Presumably because no Vulkan WG member yet has an implementation, the extension has not been released. Because there are implementations of the equivalent OpenGL extension, KTX-Software needs the enum values hence we include 289 when generating our various include files.

The proposal here is not specific to KTX and may not even be related. The way to carry such formats in a KTX v2 container is to set the vkFormat header field to `VK_FORMAT_UNDEFINED`, describe the format in the Data Format Descriptor and, if applicable, include format-mapping metadata giving the OpenGL, DXGI or Metal API formats. If you are dealing with another API you can use a vendor-specific metadata key to give the format mapping.

For the `*INDEX{8,16,32}` formats you can use `VK_FORMAT_R{8,16,32}_UINT` together with vendor-specific metadata that says to use nearest filtering.

You can also use a vendor-specific metadata key with any of them to carry the palette.

Of course vendor-specific metadata is not very portable. If you feel there is wide demand you can propose adding these items as official KTX metadata.

oddhack:
@J-Vernay given Mark's response, I don't see any way to progress this in a way either the Vulkan WG or the KTX maintainers are going to be in favor of. I suggest you close it and pursue the directions Mark and @fluppeteer are suggesting.

oddhack:
@J-Vernay we're going to go ahead and close this as not actionable - please pursue the alternate directions discussed above.

PR #2432, opened 2024-09-11
Rename Lina's GitHub username to @linyaa-kiwi

linyaa-kiwi:
This is the final GitHub rename, I promise ;-)

I'm standardizing my usernames across the web to be @linyaa when available, and otherwise match regex /linyaa[._-]?kiwi/, depending on which chars the site allows.


oddhack:
NP. Were there others I missed?

N.b. I actually went by the farm which claims to be the origin of the kiwi fruit (really just the origin of the name as "Chinese Gooseberry" lacks a certain flair) when visiting NZ.

PR #2436, opened 2024-09-14
vkGetInstanceProcAddr get device extension dispathcable command that extension must been enabled

FuXiii:
In my system case, if I didn't enabled device extension, but fetch function pointer directly, it will return `NULL` .

I think we had better add the limilt about `enabled device extension` .

✧(≖ ◡ ≖)

### DIFF
--- a/chapters/initialization.adoc
+++ b/chapters/initialization.adoc
@@ -64,7 +64,7 @@ endif::VK_VERSION_1_2[]
 | instance         | flink:vkGetInstanceProcAddr                  | fp
 | instance         | core _dispatchable command_                  | fp^3^
 | instance         | enabled instance extension dispatchable command for pname:instance    | fp^3^
-| instance         | available device extension^4^ dispatchable command for pname:instance | fp^3^
+| instance         | enabled and available device extension^4^ dispatchable command for pname:instance | fp^3^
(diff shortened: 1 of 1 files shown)


TomOlson:
@FuXiii we don't understand what you are trying to fix.  Could you post exactly what entry point you called with what parameters, and tell us what result you got and why it is not what you expected?

FuXiii:
I means, before we get device extension dispatchable command for instance, should we must enabled it?

r-potter:
As long as at least one physical device supports the required extension, you can call `vkGetInstanceProcAddr` to retrieve a device extension dispatchable command, and you should get a non-null result back. 

You can do this before you have even created a `VkDevice`, which implies that the device extension does not need to have been enabled. However, for non-physical device level functions (i.e. anything that takes `VkDevice` or any of it's children) you will need to enable the device extension before actually **calling** the returned function.

FuXiii:
Is that means: I can use `vkGetInstanceProcAddr` to load any device extension dispatchable command that physical device support support, in this case I just can load them but can not calling them. If I want calling them I must need enabled relevant device extension.

FuXiii:
Ok, thank you very much.

PR #2442, opened 2024-10-04
Update VK_AMDX_shader_enqueue to v2

Tobski:
- Adds mesh shader node support
 - Adds two new limits to workgroups
 - Variable scratch size support

### DIFF
--- a/appendices/VK_AMDX_shader_enqueue.adoc
+++ b/appendices/VK_AMDX_shader_enqueue.adoc
@@ -7,7 +7,7 @@ include::{generated}/meta/{refprefix}VK_AMDX_shader_enqueue.adoc[]
 === Other Extension Metadata
 
 *Last Modified Date*::
-    2021-07-22
+    2024-07-17
 
 *Provisional*::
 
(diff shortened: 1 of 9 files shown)


Tobski:
From a quick look, it looks like the ash build is failing because it has old definitions for commands that were updated in this extension already in the crate, but would be good if we can find someone to check that...

MarijnS95:
It looks like our hand-written "higher level" command wrappers for this extension are incompatible with the new signature, and would have to be updated via releasing the changes d regenerating + updating the `ash` crate.

However, I couldn't quickly find the right documentation on this but at least https://docs.vulkan.org/spec/latest/chapters/extensions.html#VkExtensionProperties says for the `specVersion` field (that's bumped from `1` to `2` here):

> `specVersion` is the version of this extension. It is an integer, incremented with backward compatible changes.

Is changing the function signature considered "backwards compatible"?

oddhack:
> Is changing the function signature considered "backwards compatible"?

No. But this is an old vendor "experimental" extension (AMDX vendor ID), which is something like a "beta" extension is in current practice. That means that the API is not necessarily going to be stable.

Tobski:
> > Is changing the function signature considered "backwards compatible"?
> 
> No. But this is an old vendor "experimental" extension (AMDX vendor ID), which is something like a "beta" extension is in current practice. That means that the API is not necessarily going to be stable.

Yea it's not meant to be stable, that's why it has the 'provisional="true"' attribute in the xml, as well as the AMDX tag. New revisions can just as well be entirely new extensions for provisional extensions. They're not generally available in consumer drivers, and are provided for developer feedback rather than productization.

oddhack (approved):
LGTM. Please tag Signed Off to Merge when you're ready (ideally today).

Tobski:
> LGTM. Please tag Signed Off to Merge when you're ready (ideally today).
> 
> (Edit: and resolve conflicts, of course).

Done. Still failing Ash crate CI but @MarijnS95's PR should fix that once accepted.

MarijnS95:
Thanks! We're doing the update in https://github.com/ash-rs/ash/pull/951, and attempting to prevent issues stemming from _allowed_ breaking changes to `provisional` extensions in https://github.com/ash-rs/ash/pull/952.

PR #2447, opened 2024-10-21
Primary node clarification for VK_EXT_physical_device_drm

enunes:
The extension already hinted that the primary node represents display capabilities as its intended purpose, but the current text let it ambiguous enough that there has been confusion and divergence in proposed implementations.

Add clarification notes to further define the scope of primary node in the context of this extension.

There has been a long discussion about this issue mainly at https://gitlab.freedesktop.org/mesa/mesa/-/issues/9920 .
That discussion has evolved into solving many problems of split display/render systems, but here I would like to avoid doing that again and focusing only on defining the scope of this extension.

I also discussed with the involved people again recently about proposing this note to the extension.

cc @emersion @cubanismo @fooishbar

Other references:
https://people.freedesktop.org/~cbrill/dri-log/?channel=dri-devel&highlight_names=&date=2023-07-25&show_html=true
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25431
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/31635

### DIFF
--- a/appendices/VK_EXT_physical_device_drm.adoc
+++ b/appendices/VK_EXT_physical_device_drm.adoc
@@ -7,7 +7,7 @@ include::{generated}/meta/{refprefix}VK_EXT_physical_device_drm.adoc[]
 === Other Extension Metadata
 
 *Last Modified Date*::
-    2021-06-09
+    2024-11-07
 *IP Status*::
     No known IP claims.
 *Contributors*::
@@ -25,15 +25,21 @@ Unlike the EGL extension, this extension does not expose a string containing
 the name of the device file and instead exposes device minor numbers.
 
 DRM defines multiple device node types.
-Each physical device may have one primary node and one render node
-associated.
-Physical devices may have no primary node (e.g. if the device does not have
-a display subsystem), may have no render node (e.g. if it is a software
-rendering engine), or may have neither (e.g. if it is a software rendering
-engine without a display subsystem).
+Each physical device can only be associated with a single underlying DRM
+device, which has one primary node and may have one render node.
+
+A physical device always has the primary node if it is associated with a DRM
+device. It may have no render node if e.g. it uses a software rendering engine.
 
 To query DRM properties for a physical device, chain
 slink:VkPhysicalDeviceDrmPropertiesEXT to slink:VkPhysicalDeviceProperties2.
+
+[NOTE]
+====
+This extension only returns the primary node and render node from the same
+DRM device for a Vulkan physical device. Implementations should not return the
+primary node and render node from different DRM devices.
+====
 
 include::{generated}/interfaces/VK_EXT_physical_device_drm.adoc[]
 
@@ -43,6 +49,10 @@ include::{generated}/interfaces/VK_EXT_physical_device_drm.adoc[]
     https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_device_drm.txt[`EGL_EXT_device_drm`]
 
 === Version History
+
+  * Revision 2, 2024-11-07
+  ** Clarification about the meaning of the primary node
+  ** Clarification to not advertise more than one DRM device
 
   * Revision 1, 2021-06-09
   ** First stable revision
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -22497,7 +22497,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
         </extension>
         <extension name="VK_EXT_physical_device_drm" number="354" author="EXT" type="device" contact="Simon Ser @emersion" supported="vulkan" depends="VK_KHR_get_physical_device_properties2,VK_VERSION_1_1">
             <require>
-                <enum value="1"                                             name="VK_EXT_PHYSICAL_DEVICE_DRM_SPEC_VERSION"/>
+                <enum value="2"                                             name="VK_EXT_PHYSICAL_DEVICE_DRM_SPEC_VERSION"/>
                 <enum value="&quot;VK_EXT_physical_device_drm&quot;"        name="VK_EXT_PHYSICAL_DEVICE_DRM_EXTENSION_NAME"/>
 
                 <enum offset="0" extends="VkStructureType" name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRM_PROPERTIES_EXT"/>


appendices/VK_EXT_physical_device_drm.adoc
```diff
@@ -25,12 +25,19 @@ Unlike the EGL extension, this extension does not expose a string containing
...
+that device node to be used as a display subsystem.
+Note that this is not the same definition as a "DRM primary device node" on
+Linux, which may exist even for devices without a display subsystem.
+
 Physical devices may have no primary node (e.g. if the device does not have
 a display subsystem), may have no render node (e.g. if it is a software
 rendering engine), or may have neither (e.g. if it is a software rendering
 engine without a display subsystem).
+Physical devices which don't have a display subsystem implemented as part of
+the same underlying DRM device report that as hasPrimary=false.
```
cubanismo, line 40: This is not what our implementation is doing right now. Is this what other drivers are doing currently?

appendices/VK_EXT_physical_device_drm.adoc
```diff
@@ -25,12 +25,19 @@ Unlike the EGL extension, this extension does not expose a string containing
 the name of the device file and instead exposes device minor numbers.
 
 DRM defines multiple device node types.
-Each physical device may have one primary node and one render node
-associated.
+Each physical device is associated with a single underlying DRM device, which
+may have one primary node and one render node.
+The primary node, in the context of this extension, refers to the capability of
+that device node to be used as a display subsystem.
```
cubanismo, line 31: Is this saying you can return primary node and render node from separate DRM devices on the same Vulkan physical device or not? It's not clear to me.

appendices/VK_EXT_physical_device_drm.adoc
enunes: This is not what drivers are currently doing but what I meant to put explicitly from what was hinted in the next paragraph (`Physical devices may have no primary node (e.g. if the device does not have a display subsystem)`).

To my knowledge what drivers in Mesa are doing is:
- v3dv: uses the other display DRM device as the primary. This is what we would like to fix here and prevent other drivers from following and applications to rely on.
- panvk: uses the DRM primary (cardN) node of the same render device as the primary, because it technically exists and from the DRM terminology it is in fact the "DRM primary node" (but is not a "display subsystem").
- pvr: same as panvk (but still pending to upstream Mesa).

I guess what panvk and pvr do is also acceptable to standardize on while similarly achieving the same goal, if we want to settle on something based on existing implementations. If so we also need to change the text to explicitly say that and probably remove parts of the text which associate primary devices as being display subsystems.

appendices/VK_EXT_physical_device_drm.adoc
enunes: In this snippet I just meant to clarify the terminology for the context of the extension. As in "VK_EXT_physical_device_drm primary node" != "DRM primary node". This is also due to the existing sentence which says `Physical devices may have no primary node (e.g. if the device does not have a display subsystem)`.

For the overall extension I want to say that it should not mix and return primary node and render node from separate DRM devices on the same Vulkan physical device.

I'm open to suggestions for phrasing this the best way possible.

cubanismo: The version is tracked in the XML file, which generates content in vulkan_core.h.

appendices/VK_EXT_physical_device_drm.adoc
fooishbar: Neither panvk nor pvr are (yet) conformant, and require scary environment variables to unmask them, so I'm comfortable with changing their behaviour to match new wording.

appendices/VK_EXT_physical_device_drm.adoc
fooishbar: Yeah, I'm pretty comfortable with the suggested language; users can use libdrm to map between render nodes and primary nodes of the same device.

emersion: To me, the important bit is to forbid advertising two separate DRM devices with this extension. I don't have a strong preference regarding the primary device being exposed only if the device is KMS-capable. Some notes:

- If the device is not KMS-capable, the primary node is fairly useless.
- In general, applications can't expect to have the permission to open the primary node.

> The primary node, in the context of this extension, refers to the capability of that device node to be used as a display subsystem.

That's a bit confusing IMHO. Maybe instead of trying to override what having a primary node means, we could add "If the DRM device is KMS-capable" elsewhere in the extension?

enunes:
I pushed a new version considering the points I mentioned in comment https://github.com/KhronosGroup/Vulkan-Docs/pull/2447#discussion_r1831280891

Since there is no strong need to hide the primary node (assuming that it belongs to the same DRM device) I think now that it is more intuitive to just include it there, since the primary node does exist and likely is not going away soon. So we don't introduce an entirely new rule here and have to go change every other implementation.

This new revision is less intrusive and just removes the term "display subsystem" and its association with "primary node". It also now explicitly forbids other DRM devices being used, so the only possible option is to expose the primary node of the same device. If it is not useful applications can ignore it, but I guess it is less harm it being there than having to introduce more potentially breaking changes for this extension.

appendices/VK_EXT_physical_device_drm.adoc
```diff
@@ -25,12 +25,17 @@ Unlike the EGL extension, this extension does not expose a string containing
...
-Each physical device may have one primary node and one render node
-associated.
-Physical devices may have no primary node (e.g. if the device does not have
-a display subsystem), may have no render node (e.g. if it is a software
-rendering engine), or may have neither (e.g. if it is a software rendering
-engine without a display subsystem).
+Each physical device can only be associated with a single underlying DRM
+device, which may have one primary node and one render node.
+
+Physical devices may have no primary node (e.g. it is not associated with a
```
emersion, line 31: I don't think that case is possible, actually. A render node always has a primary node.

enunes: I agree, I think the only case is the same as a software implementation which does not use DRM. Is it better to switch this text to something like "it always has the primary node if it is associated with a DRM device"?

appendices/VK_EXT_physical_device_drm.adoc
cubanismo: I guess I have a slightly stronger opinion than I thought here. Right now our logic to deduce the primary node and advertise it here doesn't have to know whether there's a display subsystem on the GPU in question. If we have to go and toggle that logic based on whether there's a display subsystem associated with the GPU, which is slightly more than a trivial amount of work in the driver.

So in other words, I'm fine with not advertising two separate DRM devices in primary and render, but I'd have a preference for not requiring drivers to hide primary nodes for displayless GPUs, because it's not what our driver does now and because it's non-trivial to switch to that behavior. Looking at it separately from the principle of least surprise POV, I agree the primary node isn't necessarily that useful on displayless GPUs, and apps can't always access it, but it's there. It seems a little weird to hide it in Vulkan when it isn't hidden on the filesystem.

That said, if everyone is set on the idea of hiding it, I could be persuaded. I'd probably want some time to go make sure there aren't any larger hurdles than anticipated finding the "has display" boolean early enough to set this field up properly though before giving my final thumbs up.

enunes: In the last (current as of now) update to this PR, I dropped the part about hiding the primary device. After the first round of comments I also realized that explicitly hiding it only makes things more complicated for little benefit.

The current proposal is to just expose the primary node of the render device anyway since it is there. So the focus of the change now is to just expliticly write to not advertise nodes from two different DRM devices.

appendices/VK_EXT_physical_device_drm.adoc
fooishbar: Yes please.

fooishbar:
@enunes this more minimal version lgtm and certainly solves the problem

appendices/VK_EXT_physical_device_drm.adoc
enunes: Updated that part to "A physical devices always has the primary node if it is associated with a DRM device.". Thanks!

enunes:
Ping, this stalled again. Any more feedback on this last version?

emersion (approved):
LGTM

fooishbar:
also lgtm - @cubanismo can we please land?

fooishbar:
> would others here be opposed to a VK_EXT_phsyical_device_drm_display or something that enables correlation of the VK_KHR_display-portions of the Vulkan spec with a DRM device

On the contrary, I think it makes loads of sense!

emersion:
Makes sense to me too, if it's specifically tied to `VK_KHR_display`.

enunes:
A separate extension in the scope of `VK_KHR_display` also makes sense to me to cover that use case.

Any update on the pending discussion so that we could proceed with this?

enunes:
Ping, this stalled again. Any updates on this?

cubanismo:
Sorry, I'm getting the pings, I just haven't been able to convene the necessary meeting of minds within NV yet. It's high on my TODO list.

russellcnv (changes requested):
Minor grammar nits.

cubanismo:
We've had more discussion, but are still stuck internally on roughly the same spot: Is it problematic if this causes our EGL implementation of [EGL_EXT_device_drm](https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_device_drm.txt) and [EGL_EXT_device_drm_render_node](https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_device_drm_render_node.txt) to diverge from our Vulkan behavior. Basically, we must advertise the display-capable DRM device from EGL_EXT_device_drm, because that's how EGLOutput+EGLStream-based apps establish their DRM "master" status, enumerate DRM-based EGLOutputs, etc. However, that would cause our EGLDevice implementation behavior to diverge from our Vulkan implementation behavior if we conform to what this clarification suggests. However, it's unclear if that's going to be problematic in practice.

Is the behavior of EGL_EXT_device_drm/EGL_EXT_device_drm_render_node in Mesa consistent with the clarifications suggested here?

enunes:
> Is the behavior of EGL_EXT_device_drm/EGL_EXT_device_drm_render_node in Mesa consistent with the clarifications suggested here?

I had not considered to make this fully consistent with the EGL extensions before, so I had not checked EGL_EXT_device_drm/EGL_EXT_device_drm_render_node for Mesa.

I checked the implementation now and tested on v3d (Raspberry Pi 4) and panfrost (Mali-G31), those are two split display/render platforms which use slightly different implementations of Mesa "renderonly". Both platforms return the primary node of the render device. The implementation for the EGL extensions actually is driver independent in Mesa. So the behavior of EGL_EXT_device_drm/EGL_EXT_device_drm_render_node in Mesa is consistent with the clarification proposed here.

panfrost
```
EGL_EXT_device_drm device: /dev/dri/card0
EGL_EXT_device_drm_render_node device: /dev/dri/renderD128

/dev/dri/by-path/
platform-ff900000.vpu-card -> ../card1
platform-ffe40000.gpu-card -> ../card0
platform-ffe40000.gpu-render -> ../renderD128
```

v3d
```
EGL_EXT_device_drm device: /dev/dri/card1
EGL_EXT_device_drm_render_node device: /dev/dri/renderD128

/dev/dri/by-path/
platform-fec00000.gpu-card -> ../card1
platform-fec00000.gpu-render -> ../renderD128
platform-gpu-card -> ../card2
```

cubanismo:
Thanks for checking Erico. I assume neither of those devices implement EGL_EXT_output_drm though, in which case they wouldn't run in to the fundamental incompatibility between that implementation choice and the use of DRM mode setting through EGLOutput+EGLStream usage that we're wrestling with.

enunes:
Ping, any chance we can get back and proceed with this?

PR #2453, opened 2024-10-28
Add docs.vulkan.org page information to validusage.json

oddhack:
The VUID information in validusage.json is now augmented with a 'page' key which contains the relative path under 'https://docs.vulkan.org/spec/latest/' to the HTML page containing that VUID. This is intended to allow the validation layers to link into the (much faster loading) docs site.

Note that you must prefix the 'page' key with the docs site URL given above to produce a complete URL.

This should close internal issue 3617.


spencer-lunarg (approved):
@oddhack to my knowledge this change is only chaging the `page` field

We still have yet to hook that part of the `validusage.json` so this change will not effect VVL. Once this is merged, I can then wait until the next headers with it in merged into VVL and from there, easily swap the link from pointing to the current URL to the Antora version

PR #2476, opened 2024-12-25
Improve HTML rendering performance

kanashimia:
Couple lines of CSS cut page loading and resize times by 100 times (on my machine).

Description of what content-visibility is: https://web.dev/articles/content-visibility Basically browser now doesn't render/layout invisible sections. As a downside it makes scrollbar jump around slightly  when you scroll the page, as browser recalculates the layout.

Section size was calculated using the following JS code:

const getAverage = (arr) => arr.reduce((p, c) => p + c, 0) / arr.length getAverage(Array.from(document.querySelectorAll(".sect1").entries().map(([idx, el]) => el.getBoundingClientRect().height ))) getAverage(Array.from(document.querySelectorAll(".sect2").entries().map(([idx, el]) => el.getBoundingClientRect().height )))

### DIFF
--- a/config/khronos.css
+++ b/config/khronos.css
@@ -730,3 +730,9 @@ li > p > a[id^="VUID-"].link { color: black; text-decoration: none; }
 li > p > a[id^="VUID-"].link:hover { color: black; }
 
 .vuid { color: #4d4d4d; font-family: monospace; }
+
+/* Improve page rendering performance */
+/* contain-intrinsic-size is the approximate average height of all sections */
+.sect1 { content-visibility: auto; contain-intrinsic-size: 98000px; }
+
+.sect2 { content-visibility: auto; contain-intrinsic-size: 13000px; }


oddhack:
By "invisible" do you mean, not currently appearing in the browser window?

kanashimia:
> By "invisible" do you mean, not currently appearing in the browser window?

Yes, you can read more about it on that link.
With this change browser now assumes that invisible sections have dimensions of `contain-intrinsic-size`.
So that is why scrollbar cursor jumps when browser actually calculates the size of the sections.

kanashimia:
Converted to a draft for now as there is a problem that if you open a page through URL with `#anchor` it doesn't jump to it, which is very important.
It works if you click on the link in the nav strip or press enter on an already opened URL in the search bar, so it is a problem only on the initial load or after reload. Not sure why is that but probably can hack around it with JS.
I will investigate this later, after the holidays. Merry Christmas!

oddhack:
> Converted to a draft for now as there is a problem that if you open a page through URL with `#anchor` it doesn't jump to it, which is very important. It works if you click on the link in the nav strip or press enter on an already opened URL in the search bar, so it is a problem only on the initial load or after reload. Not sure why is that but probably can hack around it with JS. I will investigate this later, after the holidays. Merry Christmas!

I have some WIP to turn the VU sections into collapsible blocks and auto-expand them when landing on them via a #anchor, via a JS fragment run on pageshowHashEvent, so there is very possibly another interaction here as well. Not ready for turning into an MR yet but wanted to give you a headsup.

oddhack:
Also to note, getting this stuff working cross-browser and cross-platform is a right pain. Safari in particular seems to handle some of this stuff less robustly / inconsistently with desktop browsers. What you're doing here appears pretty straightforward but how it will interact with the layout algorithms and assumptions in different browsers needs to be established.

oddhack:
I'm not sure how you mean "single page documents"? It's used in the HTML artifacts created out of the repository, but not in the Antora site at docs.vulkan.org, which has a different styling system, if that answers the question.

fluppeteer:
Sorry for the delay. I'm still testing, but I think I've seen a problem: it clearly helps rendering speed on Chrome (Linux, macOS) but Safari (macOS) doesn't seem to find the subsection headers until they've been rendered - so the table of contents links only work for the top level links until the corresponding section has loaded.

I imagine that problem would go away if content-visibility were applied only at the subsection level? (I can test that.) Good improvement to have, at least for initial spec load, where it works, though.

CanIUse seemed to think content-visibility is pretty common now, IIRC (and if not it should be harmlessly ignored). I'll keep testing other browsers, anyway.

kanashimia:
@fluppeteer thank for your report! I've tested this on Epiphany browser, which is based on WebKitGTK, and could reproduce the same behaviour. It is indeed very erratic there. There are problems with having nested content-visibility sections there, but also that loading indicator script breaks navigation in a different way, which isn't solved in a same way as on chromium / firefox. Applying content-visibility to only one type of the sections really cuts down on the performance for chromium.
I'll see what can be done, at worst we could just gate this to firefox and chromium only.
I think maybe it is also worth to change loading indicator script to not hide body contents.

PR #2477, opened 2025-01-02
Specify that the stored size of headerVersion in VkPipelineCacheHeaderVersionOne is 4 bytes

elasota:
Pedantic fix.  C/C++ allow enums to be larger than 4 bytes.  This specifies that the stored VkPipelineCacheHeaderVersion in VkPipelineCacheHeaderVersionOne is exactly 4 bytes.

### DIFF
--- a/chapters/pipelines.adoc
+++ b/chapters/pipelines.adoc
@@ -7838,10 +7838,10 @@ with identical pipeline identifiers in the same pipeline cache.
 
 Unlike most structures declared by the Vulkan API, all fields of this
 structure are written with the least significant byte first, regardless of
-host byte-order.
+host byte-order, and the headerVersion field is written as exactly 4 bytes.
 
 The C language specification does not define the packing of structure
-members.
+members or the size of enum types.
 This layout assumes tight structure member packing, with members laid out in
 the order listed in the structure, and the intended size of the structure is
 56 bytes.


chapters/pipelines.adoc
```diff
@@ -7838,10 +7838,10 @@ with identical pipeline identifiers in the same pipeline cache.
 
 Unlike most structures declared by the Vulkan API, all fields of this
 structure are written with the least significant byte first, regardless of
-host byte-order.
+host byte-order, and the headerVersion field is written as exactly 4 bytes.
```
dgkoch, line 7841: This is the SC-specific version of the structure (which should be updated as well). The vulkan version is at line 7793. 

And we suspect there might be other structures which need similar language.

dgkoch: * VkDeviceFaultVendorBinaryHeaderVersionOneEXT the same language and should be updated as well.
* [VkValidationCacheHeaderVersionEXT](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VkValidationCacheHeaderVersionEXT) has a table which explicitly lists the offsets, so it might be ok. 
* VulkanSC also has VkPipelineCacheHeaderVersionSafetyCriticalOne structure and the validationVersion field needs the same handling. 

Feels like maybe we should make a common include for two paragraphs that are replicated numerous places. (Might need to parameterize it by the intended structure size).

oddhack: > Feels like maybe we should make a common include for two paragraphs that are replicated numerous places. (Might need to parameterize it by the intended structure size).

No need for a separate include - write it in one place then use asciidoc transclusion like we do for boilerplate in the limits and features structs.

r-potter:
Thanks for the contribution. We resolved in 1.4.307 via an MR inspired by this one. Closing this MR, but please feel free to reraise if you feel the change is incomplete

PR #2479, opened 2025-01-14
Remove Ash CI

oddhack:


MarijnS95:
Again, there are no issues _in_ the repository: using our existing version of the `generator` version with a variable version of `vk.xml` (the latest stored in this repository) is bound to compatibility issues.

Given that we'll keep running into this, please don't comment this out temporarily but delete it entirely.

MarijnS95:
On a separate note, should we delete merged branches (and enable the "automatically delete source branch on merge") here? Forks get quite clobbered when they pull in all this historic "noise": https://github.com/KhronosGroup/Vulkan-Docs/branches/all

PR #2492, opened 2025-02-04
Fix promoted extension dependencies in vk.xml

### LINKED ISSUE #2491: Inconsistent extension dependencies using promoted extensions
It seems that typically in the vk.xml a promoted extension will has a separate clause added to any dependency on it to account for this promotion status. These can get quiet complex. For example `VK_KHR_imageless_framebuffer` has the following: `(((VK_KHR_get_physical_device_properties2+VK_KHR_maintenance2),VK_VERSION_1_1)+VK_KHR_image_format_list),VK_VERSION_1_2`.

However i noticed that in some very few cases this is not done which has caused problems for my code generation scripts. So i wonder is this intentional or an oversight?

Using a script i found at least the following cases of this happening:

- `VK_KHR_ray_tracing_pipeline` has dependency `VK_KHR_spirv_1_4+VK_KHR_acceleration_structure` but `VK_KHR_spirv_1_4` is promoted to 1.2
- `VK_KHR_ray_query` has dependency `VK_KHR_spirv_1_4+VK_KHR_acceleration_structure` but `VK_KHR_spirv_1_4` is promoted to 1.2
- `VK_KHR_shader_quad_control` has dependency `VK_VERSION_1_1+VK_KHR_vulkan_memory_model+VK_KHR_shader_maximal_reconvergence` but `VK_KHR_vulkan_memory_model` is promoted to 1.2
- `VK_EXT_mesh_shader` has dependency `VK_KHR_spirv_1_4` which is 1.2
- `VK_KHR_pipeline_binary` has dependency `VK_KHR_maintenance5` which is 1.4
- `VK_EXT_mutable_descriptor_type` has dependency `VK_KHR_maintenance3` which is 1.1
- `VK_KHR_compute_shader_derivatives` has dependency `VK_KHR_get_physical_device_properties2` which is 1.1
- `VK_KHR_video_encode_quantization_map` has dependency `VK_KHR_video_encode_queue+VK_KHR_format_feature_flags2` but `VK_KHR_format_feature_flags2` is 1.3
- `VK_EXT_device_generated_commands` has dependency `VK_KHR_buffer_device_address+VK_KHR_maintenance5` which is 1.2 and 1.3

CodingRays:
Fixes #2491

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -20760,7 +20760,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
                 <feature name="descriptorIndexing" struct="VkPhysicalDeviceVulkan12Features"/>
             </require>
         </extension>
-        <extension name="VK_KHR_ray_tracing_pipeline" number="348" type="device" depends="VK_KHR_spirv_1_4+VK_KHR_acceleration_structure" author="KHR" contact="Daniel Koch @dgkoch" supported="vulkan" sortorder="1" ratified="vulkan">
+        <extension name="VK_KHR_ray_tracing_pipeline" number="348" type="device" depends="(VK_KHR_spirv_1_4,VK_VERSION_1_2)+VK_KHR_acceleration_structure" author="KHR" contact="Daniel Koch @dgkoch" supported="vulkan" sortorder="1" ratified="vulkan">
             <require>
                 <enum value="1"                                             name="VK_KHR_RAY_TRACING_PIPELINE_SPEC_VERSION"/>
                 <enum value="&quot;VK_KHR_ray_tracing_pipeline&quot;"       name="VK_KHR_RAY_TRACING_PIPELINE_EXTENSION_NAME"/>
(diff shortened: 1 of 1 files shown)


oddhack:
Two comments on superficial consistency review (mostly for @spencer-lunarg who signed up as the other reviewer):

- VK_KHR_shader_quad_control - VK_KHR_vulkan_memory_model was promoted to 1.2, but the vulkanMemoryModel capability was made optional
- VK_EXT_device_generated_commands - VK_KHR_buffer_device_address was promoted to 1.2, but the bufferDeviceAddress feature capability was made optional

Possibly this would affect use of these extensions when 1.2 is supported but the corresponding promoted extension was not.

oddhack:
@asuessenbach this is getting a Vulkan-Hpp CI failure, but it appears more likely to be a syntax error in the generated code than a problem with this PR - could you take a look?

oddhack:
Followup: we think that the two partial promotions above should not be reflected in the XML, because they basically make the entire functionality of the extension optional if only 1.2 is supported. @CodingRays could you make those reversions?

spencer-lunarg (approved):
ran through Validation code gen, didn't have any issues, so tooling PoV - LGTM

oddhack:
We discussed and you are correct about the other cases. We think the path forward will be to document the feature restrictions in VU statements since capturing it in the XML is not currently possible. So, signing this off, but we'll have to sort out the CI failure - I think Vulkan-Hpp may have a bug which may or may not have been fixed, so will try re-running it to see if that helps.

oddhack:
I'm not sure what's up with the hpp-compile CI stage but it's not specific to this PR, so accepting it. Guessing that the underlying image is running an out of date g++ or something like that, as it's affecting other branches as well.

PR #2494, opened 2025-02-07
Replace fetches with fetch

vocasle:
A tiny typo fix to Pipelines article.
When I was trying to find the string I also found that same typo was present in appendices/VK_HUAWEI_cluster_culling_shader.adoc.


oddhack:

PR #2519, opened 2025-04-01
Reserve Extension number 623 for Robert Mader

rmader:
For upcoming format extension, see
https://github.com/rmader/Vulkan-Docs/commits/ycbcr-16bit-lsb-formats/

---

Please excuse me if I missed some registration step.


oddhack:
It's best to do these internally (which you should be able to do if working for Collabora) since there is potential for conflicts - several other extension numbers had already run over 623 so I've reserved 626 for you in the member gitlab and that will show up on github in the next spec update on Friday.

PR #2521, opened 2025-04-07
Fix "If If" in Texel Replacement section

kainino0x:
https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#textures-texel-replacement

> <del>If </del>If the robustImageAccess2 feature is enabled, zero values must be returned. If only the sample index was invalid, the values returned are undefined.

I think this is the correct fix.


oddhack:
Signed off as editorial fix.

PR #2523, opened 2025-04-10
Fix missing `more` in `zero or more`

LiterallyVoid:


spencer-lunarg:
This section explaining `nextStage` has another update (https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7264) that rewrote it completely to make it more clear... it will be the next 1.4.313 spec

LiterallyVoid:
Thanks!

PR #2527, opened 2025-04-25
Match pname with vk.xml.

pawchen:
Same type of changes for other files I also want to include in this PR, if this one is OK.


oddhack:
Signed off. If there are additional similar change, please do a new PR.

PR #2552, opened 2025-07-01
Update language feature check in vk.xml to include nullptr C23 support

mematthias:
This pull request updates a single preprocessor condition in vk.xml to improve `nullptr` with modern C23 standard


oddhack:
Thanks for the PR. However, Vulkan is explicitly defined as a C99 API and unlike compiling with C++, we see little incremental value in adding this, so will not be taking it.

PR #2558, opened 2025-07-20
Fixed missing type reference `VkPipelineCacheCreateFlagBits`

mematthias:
- Added missing type reference `VkPipelineCacheCreateFlagBits`


oddhack:
Seems OK to me although I think the "temporary workaround" comment should be taken out (and of the other place it currently appears in the XML) as we are being more explicit about APIs introduced by extensions even when that information is derivable from other XML dependencies.

@dgkoch similar change should likely also be done in VKSC_VERSION_1_0.

oddhack:
@mematthias signed off by the WG once the CLA is executed. There are a couple of followon actions for @oddhack once this is merged, elsewhere in the XML.

oddhack:
Thanks!

PR #2569, opened 2025-08-22
Fix asciidoc conditional

garrettleach:
Added missing brackets on conditional. This is meant to ensure only the present_wait or present_wait2 version of the Present Wait section is output.


oddhack:

PR #2572, opened 2025-08-24
Fix two occurrences of missing "be" in "must be less than" statements

gabecampb:


oddhack:

PR #2582, opened 2025-09-05
Fix mention of "C++" in "API Registry"

Rageking8:
It renders as such at [`https://registry.khronos.org/vulkan/specs/latest/registry.html`](https://registry.khronos.org/vulkan/specs/latest/registry.html):

image

---

I have traced the change to commit https://github.com/KhronosGroup/Vulkan-Docs/commit/fd0e4c3b67dd07877eae97965e863f2945104d9e. Looking at the [diff](https://github.com/KhronosGroup/Vulkan-Docs/commit/fd0e4c3b67dd07877eae97965e863f2945104d9e.diff), it seems that some "+" characters were mistakenly changed to "`":

image

### DIFF
--- a/registry.adoc
+++ b/registry.adoc
@@ -2786,7 +2786,7 @@ tag:type is:
 --------------------------------------
 
 In this case, the type `VkInstance` is defined by a special compile time
-macro which defines it as a derived class of `VkObject` (for `C```) or a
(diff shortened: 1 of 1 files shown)


oddhack:
Thanks!

PR #2589, opened 2025-09-26
Create Antora spec nav.adoc from vkspec.adoc

### LINKED ISSUE #2585: Fix spec module Antora navigation to always match single-file HTML spec
Currently the nav.adoc used for the Antora 'spec' module pages is hand-written. When chapters are added to or rearranged within the top-level vkspec.adoc, this is not reflected in nav.adoc. This can be fixed with some small markup tweaks to vkspec.adoc and a sed script to extract and rewrite the 'include::' chapter directives therein.

Note that this relies on our convention that every spec chapter is a distinct file included from vkspec.adoc, though the chapters can include many other files at deeper structural levels.

oddhack:
Prevent inconsistencies when a new chapter is added to vkspec.adoc and the chapter author does not also add it to nav.adoc. I have verified that the site built from this branch is identical to the site built from main branch, up to the addition of the spec preamble.adoc (copyright).

Closes #2585


oddhack:
Signed off as editorial.

PR #2591, opened 2025-09-30
Fix some spelling errors

Rageking8:


Rageking8:
> @Rageking8 did you use a tool to find these?

No. It's done iteratively, where I choose common terms and generate their typo permutations before checking for matches.

oddhack:
I am a little surprised codespell did not catch a few of those, but it would not have for most of them given the way in which it works.

PR #2599, opened 2025-10-09
Add step to Github CI to match gitlab CI

oddhack:


oddhack:
@r-potter per gitlab discussion. Just adds a CI job to github that was only getting done on gitlab. Not a very important one, just detecting an occasional minor nuisance.

PR #2601, opened 2025-10-23
Fix const pNext in returnedonly="true" structs

mematthias:
Some structs marked `returnedonly="true"` declare `pNext` as `const void*`.
This is inconsistent - returned-only structs are output structs, and their `pNext` must be writable by the implementation.

This change updates these structs to use `void* pNext` for consistency with `VkBaseOutStructure`.


oddhack:
Signed off once representatives for ARM, SI TSG, and Android approve. I have a script queued up to detect these in CI going forward which will also be added.

janharaldfredriksen-arm:
Approved for Arm.

oddhack:
We put this on the auto-approve queue for next week unless @cubanismo or @ianelliottus disapprove.

cubanismo (approved):
Changes to VkDisplayModeStereoPropertiesNV and VkPhysicalDevicePresentationPropertiesANDROID look fine to me.

PR #2602, opened 2025-10-23
vkGetPipelinePropertiesEXT - remove validstructs attribute, use concrete struct type

mematthias:
Use `VkPipelinePropertiesIdentifierEXT*` instead of `VkBaseOutStructure*` for `pPipelineProperties` and drop the obsolete validstructs attribute.

VkPipelinePropertiesIdentifierEXT is always the required type, so this change improves clarity and makes the C API type-safe.

The `validstructs` attribute is no longer used in the current Vulkan registry.

### DIFF
--- a/xml/vk.xml
+++ b/xml/vk.xml
@@ -17233,7 +17233,7 @@ typedef void* <name>MTLSharedEvent_id</name>;
             <proto><type>VkResult</type> <name>vkGetPipelinePropertiesEXT</name></proto>
             <param><type>VkDevice</type> <name>device</name></param>
             <param>const <type>VkPipelineInfoEXT</type>* <name>pPipelineInfo</name></param>
-            <param noautovalidity="true" validstructs="VkPipelinePropertiesIdentifierEXT"><type>VkBaseOutStructure</type>* <name>pPipelineProperties</name></param>
+            <param noautovalidity="true"><type>VkPipelinePropertiesIdentifierEXT</type>* <name>pPipelineProperties</name></param>
         </command>
         <command>
             <proto><type>void</type> <name>vkExportMetalObjectsEXT</name></proto>


oddhack:
@pdaniell-nv I think this was your issue (per comments on internal MR 4378), so you are probably the best person to reply.

pdaniell-nv:
> @pdaniell-nv I think this was your issue (per comments on internal MR 4378), so you are probably the best person to reply.

You may be thinking of @dgkoch.

In the extension page (https://docs.vulkan.org/refpages/latest/refpages/source/VK_EXT_pipeline_properties.html) there is an issue about this:
> (2) This is intended to be a general pipeline properties query, but is currently only retrieving the pipeline identifier. Should the pipeline identifier query be mandatory for this extension and for all queries using this command?
> 
> RESOLVED: Use VkBaseOutStructure for the return parameter. Currently this is required to actually be a VkPipelinePropertiesIdentifierEXT structure, but that could be relaxed in the future to allow other structure types or to allow other structures to be chained in along with this one.

So I think it's like this on purpose.

oddhack:
> > @pdaniell-nv I think this was your issue (per comments on internal MR 4378), so you are probably the best person to reply.
> 
> You may be thinking of @dgkoch.

I was looking at your long comment on the internal issue in support of it, but yes, it looks like it was Daniel's idea for a new query paradigm.

mematthias:
@oddhack @pdaniell-nv Instead of introducing a "new" query paradigm, we should stick with universally consistent `pNext` chain approach. I don’t object to the new paradigm in principle, but since `pNext` is used everywhere else and this command is the only exception, it seems unnecessary to deviate from the established model.

To align with existing `*Properties2*` patterns, this proposal:
1. introduces a simple base container `VkPipelinePropertiesEXT`,
2. keeps `VkPipelinePropertiesIdentifierEXT` as an extension struct via `structextends`, and
3. adds `vkGetPipelineProperties2EXT` as the `pNext`-compatible entry point, while deprecating the original command.

**Types:**
```xml
<type category="struct" name="VkPipelinePropertiesEXT" returnedonly="true">
    <member values="VK_STRUCTURE_TYPE_PIPELINE_PROPERTIES_EXT"><type>VkStructureType</type> <name>sType</name></member>
    <member optional="true"><type>void</type>* <name>pNext</name></member>
</type>
<type category="struct" name="VkPipelinePropertiesIdentifierEXT" returnedonly="true" structextends="VkPipelinePropertiesEXT">
    <member values="VK_STRUCTURE_TYPE_PIPELINE_PROPERTIES_IDENTIFIER_EXT"><type>VkStructureType</type> <name>sType</name></member>
    <member optional="true"><type>void</type>*              <name>pNext</name></member>
    <member><type>uint8_t</type>                            <name>pipelineIdentifier</name>[<enum>VK_UUID_SIZE</enum>]</member>
</type>
```

**Commands**
```xml
<command successcodes="VK_SUCCESS" errorcodes="VK_ERROR_OUT_OF_HOST_MEMORY,VK_ERROR_UNKNOWN,VK_ERROR_VALIDATION_FAILED">
    <proto><type>VkResult</type> <name>vkGetPipelinePropertiesEXT</name></proto>
    <param><type>VkDevice</type> <name>device</name></param>
    <param>const <type>VkPipelineInfoEXT</type>* <name>pPipelineInfo</name></param>
    <param noautovalidity="true"><type>VkPipelinePropertiesIdentifierEXT</type>* <name>pPipelineProperties</name></param>
</command>
<command successcodes="VK_SUCCESS" errorcodes="VK_ERROR_OUT_OF_HOST_MEMORY,VK_ERROR_UNKNOWN,VK_ERROR_VALIDATION_FAILED">
    <proto><type>VkResult</type> <name>vkGetPipelineProperties2EXT</name></proto>
    <param><type>VkDevice</type> <name>device</name></param>
    <param>const <type>VkPipelineInfoEXT</type>* <name>pPipelineInfo</name></param>
    <param><type>VkPipelinePropertiesEXT</type>* <name>pPipelineProperties</name></param>
</command>
```

**Extension registration (add, then deprecate the old entry point):**
```xml
<extension name="VK_EXT_pipeline_properties" >
    <require>
        <enum offset="2" extends="VkStructureType"                  name="VK_STRUCTURE_TYPE_PIPELINE_PROPERTIES_EXT"/>
        <type name="VkPipelinePropertiesEXT"/>
        <command name="vkGetPipelineProperties2EXT"/>
    </require>
    <deprecate explanationlink="deprecation-...">
        <command name="vkGetPipelinePropertiesEXT"/>
    </deprecate>
</extension>
```

This keeps the API consistent and extensible via `structextends`, and the use of the standard `pNext` chain makes explicit `validstructs` annotations unnecessary.

oddhack:
We discussed on today's Vulkan WG call. We are not willing to make this breaking change to a published API. The only way to do it would be as a new extension (or feature in a maintenance extension) which deprecates the entry point. It is not likely the WG would be interested in spending time on that, though.

mematthias:
Completely understandable, but unfortunate

PR #2607, opened 2025-11-02
Fix incorrect returnonly=true on VkPhysicalDeviceFragmentDensityMapLayeredFeaturesVALVE

mematthias:
`VkPhysicalDeviceFragmentDensityMapLayeredFeaturesVALVE` extends `VkDeviceCreateInfo` via `structextends`, but was incorrectly marked with `returnonly="true"`.
This would imply that the application can only query this feature but cannot enable or disable it, which is not the intended behavior.

This change removes the `returnonly="true"` attribute to allow the feature to be properly configured during device creation.


oddhack:
I asked the Valve contacts for this to take a look to verify the typo.

PR #2616, opened 2025-11-10
`-removeExtensions` are now applied to platform headers too

vinh-truong-sc:


vinh-truong-sc:
This PR fixes the following use case:
```
python3 ../scripts/genvk.py -registry vk.xml .... -removeExtensions=VK_NV_win32_keyed_mutex vulkan_win32.h
```

oddhack:
Thanks. Verified this has no impact on artifacts we normally produce.

PR #2617, opened 2025-11-14
Align pointer and pointer-to-pointer param/member names

mematthias:
This PR updates the naming of pointer and pointer-to-pointer parameters and members to improve consistency with Vulkans naming conventions and existing API patterns. Only parameter names are changed; no metadata, types, or semantics are modified.


r-potter:
Unfortunately we can't really do this without breaking existing application code.

r-potter:
Thanks for the contribution. The Working Group discussed this on 2025-11-19.

Unfortunately one of the constraints that we operate under is that it is not acceptable to cause previously valid application code to begin failing to compile on a header update, and so we won't take this particular MR.

PR #2624, opened 2025-11-29
Fixed VkPresent* bitmasks and enums link

mematthias:
VkFlags bitmasks are always linked to the respective flag bits enum using `requires`.


oddhack:
Approved as editorial - bitvalues is only supposed to be used for 64-bit enum types. This makes no difference to generated files, other than moving a couple of enum typedef declarations around.

PR #2643, opened 2025-12-29
Fix copy/paste error in `wsi.adoc`

EasyVulkan:
From the current description of `VkSurfacePresentModeCompatibilityKHR`:

> Otherwise, `presentModeCount` must point to a variable set by the application to the number of elements in the `pPresentModes` array,

Apparently a copy / paste error. `VkSurfacePresentModeCompatibilityKHR::presentModeCount` is not a pointer.

Change it to:

> Otherwise, `presentModeCount` must be set by the application to the number of elements in the `pPresentModes` array,


oddhack:
@cubanismo agreed this is a cut & paste error, probably from when we were getting the array query language more consistent. Please ask SI TSG to sign off.

chapters/VK_KHR_surface/wsi.adoc
```diff
@@ -3403,10 +3403,10 @@ endif::VK_EXT_surface_maintenance1[]
 If pname:pPresentModes is `NULL`, then the number of present modes that are
 compatible with the one specified in slink:VkSurfacePresentModeKHR is
 returned in pname:presentModeCount.
-Otherwise, pname:presentModeCount must: point to a variable set by the
-application to the number of elements in the pname:pPresentModes array, and
-on return the variable is overwritten with the number of values actually
-written to pname:pPresentModes.
+Otherwise, pname:presentModeCount must: be set by the application to the
+number of elements in the pname:pPresentModes array, and on return the
+variable is overwritten with the number of values actually written to
```
cubanismo: Maybe also drop "the variable" here since it is a product of the indirection language removed above it?

cubanismo:
SI TSG Discussion Today: Approved once CI and feedback above are resolved. Assigning @EasyVulkan to follow-up.

chapters/VK_KHR_surface/wsi.adoc
EasyVulkan: > Maybe also drop "the variable" here since it is a product of the indirection language removed above it?

Descriptions of following variables use the exactly same phrase _"and on return **the variable** is ..."_:
```
// In limits.adoc
	layeredApiCount
	copyDstLayoutCount
	copySrcLayoutCount

// In wsi.adoc
	presentModeCount
	presentationTimingCount
	timingCount
```
Remove "the variable" from all these descriptions?

cubanismo (approved):
Ah I didn't realize this wording was so widespread. Regardless, the new wording sounds better to me.

cubanismo:
Note: Ignore the copyright date check CI failures for now. Something isn't working right there. I manually verified the copyright was correctly updated for the files this change touches.

oddhack:
> Ah I didn't realize this wording was so widespread. Regardless, the new wording sounds better to me.

We tried to harmonize all the array query language a while back. In this case the pattern is ", and on return pname:{name of member} is overwritten...". The more common case is "pname:foo points to a variable ... and on return the variable is overwritten" but that's not for queries where the count is inside a structure.

As for CI that will continue to break until I push the next spec update from gitlab later today.

cubanismo:
Assigning to @oddhack. No objections to merging this from the SI TSG. Assuming CI passes now, Jon, up to you if you want to circulate for review with the main WG given the language was altered in other areas to keep it consistent, or merge as-is.

PR #2648, opened 2026-01-03
Add power 2.2 colorspace enum to VK_EXT_swapchain_colorspace

llyyr:
Primary motivation for this is Wayland platform, which recommends to use gamma22 for video and computer graphics for applications but there's no gamma2.2 colorspace to set via the Vulkan API.


cubanismo (changes requested):
Generally we don't add new enums to existing extensions. This would require a new extension. If you feel strongly that this is an exceptional case and must be done as part of VK_EXT_swapchain_colorspace, please explain why and we can discuss within Khronos.

llyyr:
> Generally we don't add new enums to existing extensions. This would require a new extension.

I wasn't aware of this. I'm completely fine to make this a new extension, like swapchain_colorspace2 or such. If that is fine, I'll open a new PR with that.

PR #2664, opened 2026-01-24
Update reference from srcBuffer to srcImage for CmdCopyImage

martty:


oddhack:
Good typo catch - signed off as editorial.

PR #2692, opened 2026-03-02
Add the VK_EXT_external_semaphore_drm_syncobj extension

### LINKED ISSUE #2473: Support for DRM `syncobj` FDs in `VkSemaphoreGetFdInfoKHR`
[DRM timeline sync objects](https://www.kernel.org/doc/html/v5.4/gpu/drm-mm.html?highlight=syncobj#drm-sync-objects) seem tailor-made for interop with Vulkan timeline semaphores. However, while it's possible to import these using [`vkSemaphoreFdKHR`](https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/vkGetSemaphoreFdKHR.html) and `VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_FD_BIT`, there are a few issues that make importing/exporting unergonomic:

 - When importing, the `SYNC_FD` type seems more correct, and it's not clear in the spec which should be used in this case (unless I missed it). However, `SYNC_FD` is restricted to temporary binary semaphore imports, and does not allow permanent timeline semaphore imports.
 - When exporting a semaphore as a FD using [`vkSemaphoreGetFdInfoKHR`](https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/VkSemaphoreGetFdInfoKHR.html), there's no way to distinguish what type of FD you're getting.

Should syncobjs be covered by an additional enum value in [`HandleTypeFlagBits`](https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/VkExternalSemaphoreHandleTypeFlagBits.html)? Or should the the spec/validation be relaxed to allow timeline semaphores to be imported/exported using `SYNC_FD`?

mahkoh:
This extension enables interop between timeline semaphores and Linux DRM synchronization objects (syncobj).

Closes #2473

### DIFF
--- a/chapters/capabilities.adoc
+++ b/chapters/capabilities.adoc
@@ -1641,6 +1641,12 @@ ifdef::VK_NV_external_sci_sync[]
     hardware engines including the CPU and software (intra-process and
     inter-process) operating domains and perform signal and wait operations.
 endif::VK_NV_external_sci_sync[]
+ifdef::VK_EXT_external_semaphore_drm_syncobj[]
+  * ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT specifies a
+    file descriptor handle to a Linux DRM synchronization object (syncobj).
+    This type must be used with timeline semaphores and Vulkan timeline
+    semaphore values corresponding to syncobj points.
+endif::VK_EXT_external_semaphore_drm_syncobj[]
 
 [NOTE]
 ====
@@ -1674,6 +1680,9 @@ endif::VK_FUCHSIA_external_semaphore[]
 ifdef::VK_NV_external_sci_sync[]
 | ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SCI_SYNC_OBJ_BIT_NV | No restriction | No restriction
 endif::VK_NV_external_sci_sync[]
+ifdef::VK_EXT_external_semaphore_drm_syncobj[]
+| ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT | No restriction | No restriction
+endif::VK_EXT_external_semaphore_drm_syncobj[]
 |====
 --
 
--- a/chapters/synchronization.adoc
+++ b/chapters/synchronization.adoc
@@ -3746,6 +3746,17 @@ ifdef::VK_NV_external_sci_sync[]
     slink:VkPhysicalDeviceExternalSciSyncFeaturesNV::pname:sciSyncExport>>
     features must: be enabled
 endif::VK_NV_external_sci_sync[]
+ifdef::VK_EXT_external_semaphore_drm_syncobj[]
+  * If pname:handleTypes includes
+    ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT, the
+    <<features-externalSemaphoreDrmSyncobj,
+    slink:VkPhysicalDeviceExternalSemaphoreDrmSyncobjFeaturesEXT::pname:externalSemaphoreDrmSyncobj>>
+    feature must: be enabled
+  * If pname:handleTypes includes
+    ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT, the
+    slink:VkSemaphoreTypeCreateInfo::pname:semaphoreType field must: be
+    ename:VK_SEMAPHORE_TYPE_TIMELINE
+endif::VK_EXT_external_semaphore_drm_syncobj[]
 ****
 
 include::{generated}/validity/structs/VkExportSemaphoreCreateInfo.adoc[]
@@ -5180,6 +5191,9 @@ The handle types supported by pname:handleType are:
 | Handle Type                                               | Transference | Permanence Supported
 | ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_FD_BIT | Reference    | Temporary,Permanent
 | ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SYNC_FD_BIT   | Copy         | Temporary
+ifdef::VK_EXT_external_semaphore_drm_syncobj[]
+| ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT | Reference | Permanent
+endif::VK_EXT_external_semaphore_drm_syncobj[]
 |====
 
 .Valid Usage
@@ -5213,6 +5227,17 @@ ifdef::VK_BASE_VERSION_1_2,VK_KHR_timeline_semaphore[]
     semaphore from which pname:fd was exported must: not be
     ename:VK_SEMAPHORE_TYPE_TIMELINE
 endif::VK_BASE_VERSION_1_2,VK_KHR_timeline_semaphore[]
+ifdef::VK_EXT_external_semaphore_drm_syncobj[]
+  * If pname:handleType is
+    ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT, the
+    <<features-externalSemaphoreDrmSyncobj,
+    slink:VkPhysicalDeviceExternalSemaphoreDrmSyncobjFeaturesEXT::pname:externalSemaphoreDrmSyncobj>>
+    feature must: be enabled
+  * If pname:handleType is
+    ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT, the
+    slink:VkSemaphoreTypeCreateInfo::pname:semaphoreType field must: be
+    ename:VK_SEMAPHORE_TYPE_TIMELINE
+endif::VK_EXT_external_semaphore_drm_syncobj[]
 ****
 
 If pname:handleType is ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SYNC_FD_BIT,
--- a/proposals/VK_EXT_external_semaphore_drm_syncobj.adoc
+++ b/proposals/VK_EXT_external_semaphore_drm_syncobj.adoc
@@ -1,0 +1,58 @@
+// Copyright 2021-2026 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+= VK_EXT_external_semaphore_drm_syncobj
+:toc: left
+:docs: https://docs.vulkan.org/spec/latest/
+:refpages: https://docs.vulkan.org/refpages/latest/refpages/source/
+:sectnums:
+
+This extension aims to add interoperability between timeline semaphores and
+Linux DRM synchronization objects (syncobjs).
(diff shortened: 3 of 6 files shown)


mahkoh:
Mesa implementation: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40171

xml/vk.xml
```diff
@@ -30347,6 +30347,13 @@ endif::VK_KHR_internally_synchronized_queues[]
                 <enum value="&quot;VK_SEC_extension_675&quot;"             name="VK_SEC_EXTENSION_675_EXTENSION_NAME"/>
             </require>
         </extension>
+        <extension name="VK_EXT_external_semaphore_drm_syncobj" number="100000" type="device" depends="VK_VERSION_1_2" author="EXT" contact="Julian Orth @mahkoh" supported="vulkan" nofeatures="true">
```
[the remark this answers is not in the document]
mahkoh: Is this something worth supporting?

chapters/synchronization.adoc
```diff
@@ -3740,6 +3740,13 @@ ifdef::VK_NV_external_sci_sync[]
     slink:VkPhysicalDeviceExternalSciSyncFeaturesNV::pname:sciSyncExport>>
     features must: be enabled
 endif::VK_NV_external_sci_sync[]
+ifdef::VK_EXT_external_semaphore_drm_syncobj[]
+  * [[VUID-VkExportSemaphoreCreateInfo-handleTypes-100002]]
+    If pname:handleTypes includes
+    ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT, the
+    slink:VkSemaphoreTypeCreateInfo::pname:semaphoreType field must: be
+    ename:VK_SEMAPHORE_TYPE_TIMELINE
```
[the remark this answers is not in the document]
mahkoh: Thanks, done.

xml/vk.xml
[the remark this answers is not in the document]
mahkoh: https://github.com/KhronosGroup/Vulkan-Docs/pull/2693

xml/vk.xml
```diff
@@ -30347,10 +30347,11 @@ endif::VK_KHR_internally_synchronized_queues[]
                 <enum value="&quot;VK_SEC_extension_675&quot;"             name="VK_SEC_EXTENSION_675_EXTENSION_NAME"/>
             </require>
         </extension>
-        <extension name="VK_EXT_extension_678" number="678" author="EXT" contact="Julian Orth @mahkoh" supported="disabled">
+        <extension name="VK_EXT_external_semaphore_drm_syncobj" number="678" type="device" depends="VK_VERSION_1_2" author="EXT" contact="Julian Orth @mahkoh" supported="vulkan" nofeatures="true">
```
cubanismo: This extension will have to have a feature bit. This is a newer requirement that most of the other external object specs predate. You can look at the VK_NV_external_memory_rdma spec for an example of a single-feature XML and spec update.

chapters/capabilities.adoc
```diff
@@ -1618,6 +1618,12 @@ ifdef::VK_NV_external_sci_sync[]
     hardware engines including the CPU and software (intra-process and
     inter-process) operating domains and perform signal and wait operations.
 endif::VK_NV_external_sci_sync[]
+ifdef::VK_EXT_external_semaphore_drm_syncobj[]
+  * ename:VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_DRM_SYNCOBJ_BIT_EXT specifies a
+    file descriptor handle to a Linux DRM synchronization object (syncobj).
```
cubanismo: I think this extension specifically assumes DRM timeline synchronization objects, but binary syncobjs are a thing too, so the spec should probably be specific here and elsewhere. You may want to consider including that in the name as well, but it might also be considered overly verbose. I could go either way personally.

cubanismo (changes requested):
Thanks for taking the time to move this extension forward @mahkoh! I've made a quick pass through and left some feedback, but this is looking pretty good. I'll ask the SI TSG to review as well. Please keep us updated on implementation progress.

Note this will also need CTS test coverage (https://github.com/KhronosGroup/VK-GL-CTS/tree/main/external/vulkancts) and validation layer coverage (https://github.com/KhronosGroup/Vulkan-ValidationLayers) to be merged. Let us know if you plan on working on those as well. Extending the existing timeline semaphore opaque FD coverage to run on DRM syncobj handle types would likely be sufficient for CTS.

chapters/capabilities.adoc
mahkoh: I don't believe the kernel distinguishes between binary and timeline syncobjs. The ioctl to create a syncobj does not have a way to allocate one or the other:

```c
struct drm_syncobj_create {
	__u32 handle;
#define DRM_SYNCOBJ_CREATE_SIGNALED (1 << 0)
	__u32 flags;
};
```

The kernel warns against using both binary and timeline ioctls on the same syncobj but I believe the following sentence from this PR already covers this:

```xml
                                                            Vulkan timeline
    semaphore values correspond to syncobj points.
```

r-potter:
@gfxstrand cc to provide feedback

mahkoh:
> Note this will also need CTS test coverage ([KhronosGroup/VK-GL-CTS@`main`/external/vulkancts](https://github.com/KhronosGroup/VK-GL-CTS/tree/main/external/vulkancts)) and validation layer coverage ([KhronosGroup/Vulkan-ValidationLayers](https://github.com/KhronosGroup/Vulkan-ValidationLayers)) to be merged. Let us know if you plan on working on those as well. Extending the existing timeline semaphore opaque FD coverage to run on DRM syncobj handle types would likely be sufficient for CTS.

They are linked above this comment now.

cubanismo:
@mahkoh While going through the release process in Khronos, it was pointed out this change is missing a proposal document, which is required for all new extensions. It doesn't have to be very long for a small extension like this, but please document the motivation behind the extension, any design alternatives considered, and any issues that came up during review (E.g., good to document why it makes sense to limit this to Vulkan timeline semaphores rather than adding it for binary semaphores as well).

As an example of the format/layout, here's a link to the proposal document for VK_EXT_swapchain_maintenance1:

https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_swapchain_maintenance1.html

mahkoh:
I've added a proposal document. I haven't considered any alternatives so I've left out the solution space section.

oddhack:

mahkoh:
@wusha01 The comments you've posted don't seem to be appearing anywhere.

proposals/VK_EXT_external_semaphore_drm_syncobj.adoc
```diff
@@ -0,0 +1,58 @@
...
+// SPDX-License-Identifier: CC-BY-4.0
+
+= VK_EXT_external_semaphore_drm_syncobj
+:toc: left
+:docs: https://docs.vulkan.org/spec/latest/
+:refpages: https://docs.vulkan.org/refpages/latest/refpages/source/
+:sectnums:
+
+This extension aims to add interoperability between timeline semaphores and
+syncobjs.
```
wusha01: nitpicking: 
maybe use the full name 'Linux DRM synchronization objects (syncobjs) " instead of the short form?

proposals/VK_EXT_external_semaphore_drm_syncobj.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+== Problem Statement
```
wusha01, line 14: Nitpicking:IMHO, the information in this session seems to fit better as a proposal instead of a problem statement. :) Particaluarly, the information about the syncobjs provides the justification why it is the chosen solution, while the information about 'OPAQUE_FD' seems like an alternative proposal, and the limitation comes along with it. 

For the problem statement, it may be sufficient to explain what is the problem the extension is intended to solve. For example:
Vulkan does not define an external sync handle type that is suited for sharing timeline semaphore outside of Vulkan.
The existing external semaphore handle types for cross-API sharing are limited to binary semaphore usage. This limits Vulkan's ability to use timeline semaphore as sync primitive for cross API and process sharing.

mahkoh: > Vulkan does not define an external sync handle type that is suited for sharing timeline semaphore outside of Vulkan.

I don't think that is true. Vulkan does have such types on windows. And I think opaque FDs can be shared with cuda on the nvidia driver.

I wrote this section to show specifically why there is value in supporting syncobj interop. If it is already assumed that all interop with any external synchronization primitive is valuable, then I'd rather just cut this section down to the following without changing the proposal section.

> Vulkan currently does not provide semaphore interoperability with syncobjs which are a synchronization primitive used on linux.

wusha01: Fair enough.
I like the trim down version much better :)

mahkoh: I can do that if it is the consensus. If I were to read the proposal document, I would prefer to know more about why the extension was added and what things I could do with it.

wusha01: The reason I suggested an alternative problem statement is that the original problem statement reads like “Vulkan lacks a sync handle type for syncobj” and I slightly disagree with it.

I think the actual problem is that we lack a mechanism to share timeline semaphore payloads between Vulkan and other APIs on Linux. Opaque_FD could be used in a fairly limited set of use cases, while syncobj fits better to that problem, because it is a Linux kernel primitive with the same semantics as a Vulkan timeline semaphore, and is widely supported by the Linux ecosys. Therefore exporting/importing timeline semaphore payloads through drm syncobj handle becomes the choice on Linux.

wusha01: Thanks very much for writing the proposal document, by the way.
If we don’t receive others input on this, I’m fine with using the original problem statement.

mahkoh: >The reason I suggested an alternative problem statement is that the original problem statement reads like “Vulkan lacks a sync handle type for syncobj” and I slightly disagree with it.

That was how I intended it to be read. The two usecases from the problem statement are the usecases I wanted to solve with this extension.

It being the first well-defined handle type for timeline semaphores on linux is only a side effect and having any other such type would not solve the usecases I had in mind, since they specifically use syncobjs.

cubanismo: Yes, I would have to agree the problem statement seems good given the extension specifically aims to solve interoperation gaps involving this one native primitive type.

proposals/VK_EXT_external_semaphore_drm_syncobj.adoc
mahkoh: Done

proposals/VK_EXT_external_semaphore_drm_syncobj.adoc
wusha01: Seems that we have reached a conclusion. 
Resolve the issue by keeping the original problem statement.

cubanismo:
@mahkoh: At this time the Vulkan Working Group is ready to proceed with the ratification process for this extension, which would grant it protection under the [Khronos IP Framework](https://www.khronos.org/members/ip-framework). We'd like to know if you have any objections to proceeding with that step.

mahkoh:
Sounds good to me.

PR #2698, opened 2026-03-12
doc: clarify VkStructureType naming derivation to avoid double underscores

yixun-chen:
The current specification states that VkStructureType names are derived by "prefixing each capital letter with _". If followed literally, a structure like VkImageCreateInfo (after stripping Vk) would become _Image_Create_Info, leading to VK_STRUCTURE_TYPE__IMAGE_CREATE_INFO with a double underscore.

Updated the description to "prefixing each capital letter except the first with _". This aligns the text with the actual naming convention used throughout the Vulkan API and removes any ambiguity for developers.


oddhack:
That's a good catch - thanks.

PR #2709, opened 2026-04-01
Add VK_EXT_ycbcr_3plane_16bit_lsb_formats extension

rmader:
This extension adds support for 10/12bit YCbCr formats used by software decoders like ffmpeg, dav1d and libvpx.

See also:
 - https://github.com/torvalds/linux/commit/2271e0a20ef795838527815e057f5af206b69c87
 - https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/34303
 - https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/35576
 - https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/9251
 - https://www.ffmpeg.org/doxygen/1.0/pixfmt_8h.html

---

This branch as well as the related Mesa and GTK4 MRs have been in draft status for a while and already got a lot of testing. With the recent Gstreamer 1.28 release (shipped in distros like Fedora 44 and Ubuntu 26.04) they can be tested and used - e.g. the default Gnome Video player will make use of the new formats when playing HDR videos (for various codecs, including AV1, HEVC, H266, Pro-Res and DNxHR).

@gfxstrand I'd be super happy if you could have a look at this and, if you find time, help me get this over the line :sweat_smile: Most importantly point me to where I should elaborate more.

### DIFF
--- a/appendices/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
+++ b/appendices/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
@@ -1,0 +1,27 @@
+// Copyright 2026 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+include::{generated}/meta/{refprefix}VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc[]
+
+=== Other Extension Metadata
+
+*Last Modified Date*::
+    2026-04-30
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Robert Mader, Collabora
+
+=== Description
+
+This extension adds some 10/12/14bit {YCbCr} formats that are in common use for
+video software encode and decode, but were not part of the
+`apiext:VK_KHR_sampler_ycbcr_conversion` extension.
+
+include::{generated}/interfaces/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc[]
+
+=== Version History
+
+  * Revision 1, 2026-04-30 (Robert Mader)
+  ** Initial draft
--- a/chapters/features.adoc
+++ b/chapters/features.adoc
@@ -6173,6 +6173,42 @@ endif::VK_BASE_VERSION_1_3[]
 --
 endif::VK_EXT_ycbcr_2plane_444_formats[]
 
+ifdef::VK_EXT_ycbcr_3plane_16bit_lsb_formats[]
+[open,refpage='VkPhysicalDeviceYcbcr3Plane16BitLsbFormatsFeaturesEXT',desc='Structure describing whether the implementation supports additional 3-plane 16bit LSB {YCbCr} formats',type='structs']
+--
+The sname:VkPhysicalDeviceYcbcr3Plane16BitLsbFormatsFeaturesEXT structure is
+defined as:
+
+include::{generated}/api/structs/VkPhysicalDeviceYcbcr3Plane16BitLsbFormatsFeaturesEXT.adoc[]
+
+This structure describes the following feature:
+
+  * pname:sType is a elink:VkStructureType value identifying this structure.
+  * pname:pNext is `NULL` or a pointer to a structure extending this
+    structure.
+  * [[features-ycbcr3plane16BitLsbFormats]] pname:ycbcr3plane16BitLsbFormats
+    indicates that the implementation supports querying format features for, and
+    using, the following  3-plane 16bit LSB {YCbCr} formats:
+  ** ename:VK_FORMAT_Z6R10_UNORM_EXT
+  ** ename:VK_FORMAT_Z4R12_UNORM_EXT
+  ** ename:VK_FORMAT_Z2R14_UNORM_EXT
+  ** ename:VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_420_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_422_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_444_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_Z4G12_Z4B12_Z4R12_3PLANE_420_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_Z4G12_Z4B12_Z4R12_3PLANE_422_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_Z4G12_Z4B12_Z4R12_3PLANE_444_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_Z2G14_Z2B14_Z2R14_3PLANE_420_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_Z2G14_Z2B14_Z2R14_3PLANE_422_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_Z2G14_Z2B14_Z2R14_3PLANE_444_UNORM_3PACK16_EXT
+
+:refpage: VkPhysicalDeviceYcbcr3Plane16BitLsbFormatsFeaturesEXT
+include::{chapters}/features.adoc[tag=features]
+
+include::{generated}/validity/structs/VkPhysicalDeviceYcbcr3Plane16BitLsbFormatsFeaturesEXT.adoc[]
+--
+endif::VK_EXT_ycbcr_3plane_16bit_lsb_formats[]
+
 ifdef::VK_EXT_color_write_enable[]
 [open,refpage='VkPhysicalDeviceColorWriteEnableFeaturesEXT',desc='Structure describing whether writes to color attachments can be enabled and disabled dynamically',type='structs']
 --
--- a/chapters/formats.adoc
+++ b/chapters/formats.adoc
@@ -1359,6 +1359,161 @@ ifdef::VK_BASE_VERSION_1_3,VK_EXT_ycbcr_2plane_444_formats[]
     ename:VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane, and
     ename:VK_IMAGE_ASPECT_PLANE_1_BIT for the BR plane.
 endif::VK_BASE_VERSION_1_3,VK_EXT_ycbcr_2plane_444_formats[]
+ifdef::VK_EXT_ycbcr_3plane_16bit_lsb_formats[]
+  * ename:VK_FORMAT_Z6R10_UNORM_EXT specifies a one-component, 16-bit
+    unsigned normalized format that has a single 10-bit R component in the
+    bottom 10 bits of a 16-bit word, with the top 6 bits unused and required
+    to be set to zero.
+  * ename:VK_FORMAT_Z4R12_UNORM_EXT specifies a one-component, 16-bit
+    unsigned normalized format that has a single 12-bit R component in the
+    bottom 12 bits of a 16-bit word, with the top 4 bits unused and required
+    to be set to zero.
+  * ename:VK_FORMAT_Z2R14_UNORM_EXT specifies a one-component, 16-bit
+    unsigned normalized format that has a single 14-bit R component in the
+    bottom 14 bits of a 16-bit word, with the top 2 bits unused and required
+    to be set to zero.
+  * ename:VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_420_UNORM_3PACK16_EXT specifies an
+    unsigned normalized _multi-planar format_ that has a 10-bit G component
+    in the bottom 10 bits of each 16-bit word of plane 0, a 10-bit B component
+    in the bottom 10 bits of each 16-bit word of plane 1, and a 10-bit R
+    component in the bottom 10 bits of each 16-bit word of plane 2, with the
+    top 6 bits of each word unused and required to be set to zero.
+    The horizontal and vertical dimensions of the R and B planes are halved
+    relative to the image dimensions, and each R and B component is shared
+    with the G components for which latexmath:[\left\lfloor i_G \times 0.5
+    \right\rfloor = i_B = i_R] and latexmath:[\left\lfloor j_G \times 0.5
+    \right\rfloor = j_B = j_R].
+    The location of each plane when this image is in linear layout can be
+    determined via flink:vkGetImageSubresourceLayout, using
+    ename:VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane,
+    ename:VK_IMAGE_ASPECT_PLANE_1_BIT for the B plane, and
+    ename:VK_IMAGE_ASPECT_PLANE_2_BIT for the R plane.
+    This format only supports images with a width and height that is a
+    multiple of two.
+  * ename:VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_422_UNORM_3PACK16_EXT specifies an
+    unsigned normalized _multi-planar format_ that has a 10-bit G component
+    in the bottom 10 bits of each 16-bit word of plane 0, a 10-bit B component
+    in the bottom 10 bits of each 16-bit word of plane 1, and a 10-bit R
+    component in the bottom 10 bits of each 16-bit word of plane 2, with the
+    top 6 bits of each word unused and required to be set to zero.
+    The horizontal dimension of the R and B plane is halved relative to the
+    image dimensions, and each R and B value is shared with the G components
+    for which latexmath:[\left\lfloor i_G \times 0.5 \right\rfloor = i_B =
+    i_R].
+    The location of each plane when this image is in linear layout can be
+    determined via flink:vkGetImageSubresourceLayout, using
+    ename:VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane,
+    ename:VK_IMAGE_ASPECT_PLANE_1_BIT for the B plane, and
+    ename:VK_IMAGE_ASPECT_PLANE_2_BIT for the R plane.
+    This format only supports images with a width that is a multiple of two.
+  * ename:VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_444_UNORM_3PACK16_EXT specifies an
+    unsigned normalized _multi-planar format_ that has a 10-bit G component
+    in the bottom 10 bits of each 16-bit word of plane 0, a 10-bit B component
+    in the bottom 10 bits of each 16-bit word of plane 1, and a 10-bit R
+    component in the bottom 10 bits of each 16-bit word of plane 2, with the
+    top 6 bits of each word unused and required to be set to zero.
+    Each plane has the same dimensions and each R, G, and B component
+    contributes to a single texel.
+    The location of each plane when this image is in linear layout can be
+    determined via flink:vkGetImageSubresourceLayout, using
+    ename:VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane,
+    ename:VK_IMAGE_ASPECT_PLANE_1_BIT for the B plane, and
+    ename:VK_IMAGE_ASPECT_PLANE_2_BIT for the R plane.
+  * ename:VK_FORMAT_Z4G12_Z4B12_Z4R12_3PLANE_420_UNORM_3PACK16_EXT specifies an
+    unsigned normalized _multi-planar format_ that has a 12-bit G component
+    in the bottom 12 bits of each 16-bit word of plane 0, a 12-bit B component
+    in the bottom 12 bits of each 16-bit word of plane 1, and a 12-bit R
+    component in the bottom 12 bits of each 16-bit word of plane 2, with the
+    top 4 bits of each word unused and required to be set to zero.
+    The horizontal and vertical dimensions of the R and B planes are halved
+    relative to the image dimensions, and each R and B component is shared
+    with the G components for which latexmath:[\left\lfloor i_G \times 0.5
+    \right\rfloor = i_B = i_R] and latexmath:[\left\lfloor j_G \times 0.5
+    \right\rfloor = j_B = j_R].
+    The location of each plane when this image is in linear layout can be
+    determined via flink:vkGetImageSubresourceLayout, using
+    ename:VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane,
+    ename:VK_IMAGE_ASPECT_PLANE_1_BIT for the B plane, and
+    ename:VK_IMAGE_ASPECT_PLANE_2_BIT for the R plane.
+    This format only supports images with a width and height that is a
+    multiple of two.
+  * ename:VK_FORMAT_Z4G12_Z4B12_Z4R12_3PLANE_422_UNORM_3PACK16_EXT specifies an
+    unsigned normalized _multi-planar format_ that has a 12-bit G component
+    in the bottom 12 bits of each 16-bit word of plane 0, a 12-bit B component
+    in the bottom 12 bits of each 16-bit word of plane 1, and a 12-bit R
+    component in the bottom 12 bits of each 16-bit word of plane 2, with the
+    top 4 bits of each word unused and required to be set to zero.
+    The horizontal dimension of the R and B plane is halved relative to the
+    image dimensions, and each R and B value is shared with the G components
+    for which latexmath:[\left\lfloor i_G \times 0.5 \right\rfloor = i_B =
+    i_R].
+    The location of each plane when this image is in linear layout can be
+    determined via flink:vkGetImageSubresourceLayout, using
+    ename:VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane,
+    ename:VK_IMAGE_ASPECT_PLANE_1_BIT for the B plane, and
+    ename:VK_IMAGE_ASPECT_PLANE_2_BIT for the R plane.
+    This format only supports images with a width that is a multiple of two.
+  * ename:VK_FORMAT_Z4G12_Z4B12_Z4R12_3PLANE_444_UNORM_3PACK16_EXT specifies an
+    unsigned normalized _multi-planar format_ that has a 12-bit G component
+    in the bottom 12 bits of each 16-bit word of plane 0, a 12-bit B component
+    in the bottom 12 bits of each 16-bit word of plane 1, and a 12-bit R
+    component in the bottom 12 bits of each 16-bit word of plane 2, with the
+    top 4 bits of each word unused and required to be set to zero.
+    Each plane has the same dimensions and each R, G, and B component
+    contributes to a single texel.
+    The location of each plane when this image is in linear layout can be
+    determined via flink:vkGetImageSubresourceLayout, using
+    ename:VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane,
+    ename:VK_IMAGE_ASPECT_PLANE_1_BIT for the B plane, and
+    ename:VK_IMAGE_ASPECT_PLANE_2_BIT for the R plane.
+  * ename:VK_FORMAT_Z2G14_Z2B14_Z2R14_3PLANE_420_UNORM_3PACK16_EXT specifies an
+    unsigned normalized _multi-planar format_ that has a 14-bit G component
+    in the bottom 14 bits of each 16-bit word of plane 0, a 14-bit B component
+    in the bottom 14 bits of each 16-bit word of plane 1, and a 14-bit R
+    component in the bottom 14 bits of each 16-bit word of plane 2, with the
+    top 2 bits of each word unused and required to be set to zero.
+    The horizontal and vertical dimensions of the R and B planes are halved
+    relative to the image dimensions, and each R and B component is shared
+    with the G components for which latexmath:[\left\lfloor i_G \times 0.5
+    \right\rfloor = i_B = i_R] and latexmath:[\left\lfloor j_G \times 0.5
+    \right\rfloor = j_B = j_R].
+    The location of each plane when this image is in linear layout can be
+    determined via flink:vkGetImageSubresourceLayout, using
+    ename:VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane,
+    ename:VK_IMAGE_ASPECT_PLANE_1_BIT for the B plane, and
+    ename:VK_IMAGE_ASPECT_PLANE_2_BIT for the R plane.
+    This format only supports images with a width and height that is a
+    multiple of two.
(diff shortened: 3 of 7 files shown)


chapters/formats.adoc
```diff
@@ -2081,6 +2081,17 @@ ifdef::VK_KHR_maintenance5,VK_BASE_VERSION_1_3[and]
     ename:VK_FORMAT_G16_B16R16_2PLANE_444_UNORM
 ****
 endif::VK_EXT_ycbcr_2plane_444_formats[]
+ifdef::VK_EXT_ycbcr_3plane_16bit_lsb_formats[]
+  * [[VUID-vkGetPhysicalDeviceFormatProperties-None-?????]]
```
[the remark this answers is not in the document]
oddhack: VUIDs are assigned once an extension is merged to main branch. You should not have any VUID- anchors at all at this stage, so please take those out.

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
```diff
@@ -0,0 +1,39 @@
...
+:sectnums:
+
+This extension adds support for 10/12bit {YCbCr} formats used by software
+decoders like ffmpeg, dav1d and libvpx.
+
+== Problem Statement
+
+Many popular software video decoders - such as ffmpeg/libav, dav1d and libvpx -
+use formats for 10 and 12 bit content that are not yet supported in Vulkan, but
+are for example in Linux DRM and Mesa GLES. These formats long established and
```
r-potter:
```suggestion
are supported by Linux DRM and Mesa GLES. These formats are long established and
```

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+Supporting these formats directly in drivers has various advantages for apps
+and system compositors, as it allows to avoid unnecessary copies and/or quality
+loss. Further more it can also simplify code-sharing with code-paths for
+hardware decoders.
+
+Unfortunately it is not possible to use the existing 16 bit formats such as
+VK_FORMAT_G16_B16_R16_3PLANE_420_UNORM as - unlike most formats used by hardware
+decoders - the alignment is LSB, not MSB. The bit-values thus need to get
+shifted by 6 or 4 bit.
```
r-potter:
```suggestion
shifted by 6 or 4 bits.
```

appendices/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
```diff
@@ -0,0 +1,27 @@
...
+*Last Modified Date*::
+    2025-04-01
+*IP Status*::
+    No known IP claims.
+*Contributors*::
+  - Robert Mader, Collabora
+
+=== Description
+
+This extension adds some {YCbCr} formats that are in common use for video
```
r-potter:
```suggestion
This extension adds some 10/12bit {YCbCr} formats that are in common use for video
```

oddhack:
@rmader please rebase on github main branch, which should fix the CI CTS framework issue, and address the other issues raised by CI around copyright dates and naming.

chapters/formats.adoc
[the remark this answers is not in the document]
rmader: Ah, makes sense - removed!

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
rmader: Done

rmader:
> @rmader please rebase on github main branch, which should fix the CI CTS framework issue, and address the other issues raised by CI around copyright dates and naming.

Thanks for the review! I *think* I addressed all issues - and hope the change to append `_EXT` to all formats is correct. The goal of course being that they'll get promoted at some point going forward.

chapters/features.adoc
```diff
@@ -6033,6 +6033,36 @@ endif::VK_BASE_VERSION_1_3[]
...
+    structure.
+  * [[features-ycbcr3plane16BitLsbFormats]] pname:ycbcr3plane16BitLsbFormats
+    indicates that the implementation supports the following 3-plane 16bit LSB
+    {YCbCr} formats:
+  ** ename:VK_FORMAT_X6G10_X6B10_X6R10_3PLANE_420_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_X6G10_X6B10_X6R10_3PLANE_422_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_X6G10_X6B10_X6R10_3PLANE_444_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_X4G12_X4B12_X4R12_3PLANE_420_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_X4G12_X4B12_X4R12_3PLANE_422_UNORM_3PACK16_EXT
+  ** ename:VK_FORMAT_X4G12_X4B12_X4R12_3PLANE_444_UNORM_3PACK16_EXT
```
r-potter: Just noting that `VK_EXT_ycbcr_2plane_444_formats` used a slightly different pattern here:

````
  * [[features-ycbcr2plane444Formats]] pname:ycbcr2plane444Formats indicates
    that the implementation supports querying format features for, and
    using, the following 2-plane 444 {YCbCr} formats:
  ** ename:VK_FORMAT_G8_B8R8_2PLANE_444_UNORM
  ** ename:VK_FORMAT_G10X6_B10X6R10X6_2PLANE_444_UNORM_3PACK16
  ** ename:VK_FORMAT_G12X4_B12X4R12X4_2PLANE_444_UNORM_3PACK16
  ** ename:VK_FORMAT_G16_B16R16_2PLANE_444_UNORM
````

In other words  `VK_EXT_ycbcr_2plane_444_formats` adds the enums, and says that you can query for them, but did not mandate than any given implementation supports all 4 of them. Did you consider that approach, and is there a motivation for doing it this way instead?

Comment (rmader) [reply chapters/features.adoc]: Oh, indeed, that's actually better / totally sufficient. Will update.

Comment (rmader) [reply chapters/features.adoc]: Done!

**Comment by rmader** [comment]
I added one more change for completeness - 14bit formats. While those are very uncommon AFAIK, [they are supported by ffmpeg](https://www.ffmpeg.org/doxygen/1.0/pixfmt_8h.html). Thus let's include them while on it, so we don't have to revisit this in a couple of years.

File: xml/vk.xml [hunk 6]
```diff
@@ -32962,6 +32979,78 @@ endif::VK_KHR_internally_synchronized_queues[]
         <format name="VK_FORMAT_R8_SFLOAT_FPENCODING_FLOAT8E5M2_ARM" class="8-bit" blockSize="1" texelsPerBlock="1">
             <component name="R" bits="8" numericFormat="SFLOAT"/>
         </format>
+        <format name="VK_FORMAT_X6G10_X6B10_X6R10_3PLANE_420_UNORM_3PACK16_EXT" class="10-bit 3-plane LSB" blockSize="6" texelsPerBlock="1" chroma="420">
+            <component name="G" bits="16" numericFormat="UNORM" planeIndex="0"/>
+            <component name="B" bits="16" numericFormat="UNORM" planeIndex="1"/>
+            <component name="R" bits="16" numericFormat="UNORM" planeIndex="2"/>
+            <plane index="0" widthDivisor="1" heightDivisor="1" compatible="VK_FORMAT_R16_UNORM"/>
+            <plane index="1" widthDivisor="2" heightDivisor="2" compatible="VK_FORMAT_R16_UNORM"/>
+            <plane index="2" widthDivisor="2" heightDivisor="2" compatible="VK_FORMAT_R16_UNORM"/>
```
gfxstrand: I think we also need `VK_FORMAT_X6R10_UNORM`, `VK_FORMAT_X4R12_UNORM` and `VK_FORMAT_X2R16_UNORM` formats which get referenced as the per-plane formats here.

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
```diff
@@ -0,0 +1,40 @@
...
+VK_FORMAT_G16_B16_R16_3PLANE_420_UNORM as - unlike most formats used by hardware
+decoders - the alignment is LSB, not MSB. The bit-values thus need to get
+shifted by 6, 4 or 2 bits.
+
+== Proposal
+
+Add the formats in question. For any drivers already supporting YCbCr formats
+via shaders this should be straight forward. Values in the 0.0 - 1.0 range just
+need to get multiplied by 65535.0 / 1023.0 (for 10 bit), 65535.0 / 4095.0 (for
+12 bit) or 65535.0 / 16383.0 (for 14 bit).
```
gfxstrand, line 38: This is a tricky assertion to make. Right now, we assume that the X bits are garbage and (mostly) ignored by sampling. I say "mostly" because it's kind of okay if they're not since they're the low bits and any garbage in those bits will show up as noise which likely isn't perceptible to the human eye. With theese formats, however, your assertion that they're easy for shaders by doing a bit of multiplication assumes that the top bits are always zero. While this is fairly easy for software to guarantee when outputting such an image, it's very hard for the driver to defend against. We could make it invalid usage or unknown results if you have non-zero bits in the high bits when sampling but we'd need to be very explicit about that in the API.

gfxstrand (commented):
Sorry this took me nearly a month to get to. Between a Khronos meeting and vacation, it kind of got lost in my e-mail.

gfxstrand:
@fluppeteer should probably also give this a read.

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
rmader: Indeed. So on the DRM side we solved that by using `z` instead of `x`: https://github.com/torvalds/linux/blob/master/include/uapi/drm/drm_fourcc.h#L403-L405

We could do the same here by turning
```
VK_FORMAT_X6G10_X6B10_X6R10_3PLANE_420_UNORM_3PACK16_EXT
```

into

```
VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_420_UNORM_3PACK16_EXT
```

which would hopefully make this clear. WDYT?

xml/vk.xml
rmader: Yeah, I was wondering about that. I had it in a previous draft, however in my example implementation in Mesa - https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/35576 - it turned out to be much easier to just stick to `VK_FORMAT_R16_UNORM` so all drivers supporting those format would support the YCbCr ones as well.

Given that there's no clear independent use-case for these formats, I figured it would only unnecessarily complicate implementations. If there's a consent that we should have them, I'm surely happy to add them though, would just like to be sure that it's really needed :/

gfxstrand: We need to make the spec clear. If that means Mesa drivers do a bit of internal mapping, then so be it.

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
gfxstrand: We could do that. Since no one will ever be writing to these formats, it's not like we would need hardware to support that. I'd feel a little better about that than saying it's ignored. Again, I think this is where @fluppeteer will have good opinions. As might @spencer-lunarg since he did the format table in the first place.

rmader: > Since no one will ever be writing to these formats

For completeness: these formats are also the native/optimal input formats for common software encoders, meaning there is a chance that at some point we'd want native driver support for writing into shared buffers that can directly be fed into those sw-encoders. Not super likely, just to keep in mind.

chapters/resources.adoc
```diff
@@ -6728,6 +6754,19 @@ ifdef::VK_BASE_VERSION_1_3[Vulkan 1.3 is not supported and]
     ename:VK_FORMAT_G12X4_B12X4R12X4_2PLANE_444_UNORM_3PACK16, or
     ename:VK_FORMAT_G16_B16R16_2PLANE_444_UNORM
 endif::VK_EXT_ycbcr_2plane_444_formats[]
+ifdef::VK_EXT_ycbcr_3plane_16bit_lsb_formats[]
+    If the <<features-ycbcr3plane16BitLsbFormats,pname:ycbcr3plane16BitLsbFormats>>
```
oddhack:
```suggestion
  * If the <<features-ycbcr3plane16BitLsbFormats,pname:ycbcr3plane16BitLsbFormats>>
```

I think this is missing a list item bullet to separate it from the previous VU.

chapters/resources.adoc
```diff
@@ -1323,6 +1323,19 @@ ifdef::VK_BASE_VERSION_1_3[Vulkan 1.3 is not supported and]
...
```
oddhack:
```suggestion
 * If the <<features-ycbcr3plane16BitLsbFormats,pname:ycbcr3plane16BitLsbFormats>>
```

Same missing bullet.

chapters/formats.adoc
```diff
@@ -2081,6 +2081,19 @@ ifdef::VK_KHR_maintenance5,VK_BASE_VERSION_1_3[and]
     ename:VK_FORMAT_G16_B16R16_2PLANE_444_UNORM
 ****
 endif::VK_EXT_ycbcr_2plane_444_formats[]
+ifdef::VK_EXT_ycbcr_3plane_16bit_lsb_formats[]
+    If the <<features-ycbcr3plane16BitLsbFormats,pname:ycbcr3plane16BitLsbFormats>>
```
oddhack:
```suggestion
 * If the <<features-ycbcr3plane16BitLsbFormats,pname:ycbcr3plane16BitLsbFormats>>
```

chapters/formats.adoc
```diff
@@ -2447,6 +2460,19 @@ ifdef::VK_KHR_maintenance5,VK_BASE_VERSION_1_3[and]
...
```
oddhack:
```suggestion
 * If the <<features-ycbcr3plane16BitLsbFormats,pname:ycbcr3plane16BitLsbFormats>>
```

chapters/formats.adoc
```diff
@@ -2447,6 +2460,19 @@ ifdef::VK_KHR_maintenance5,VK_BASE_VERSION_1_3[and]
     ename:VK_FORMAT_G16_B16R16_2PLANE_444_UNORM
 ****
```
oddhack: The '****' marks the end of the VU block so needs to be moved after the new VU you are adding immediately below.

chapters/resources.adoc
```diff
@@ -3067,6 +3080,19 @@ ifdef::VK_BASE_VERSION_1_3[Vulkan 1.3 is not supported and]
...
```
oddhack:
```suggestion
 * If the <<features-ycbcr3plane16BitLsbFormats,pname:ycbcr3plane16BitLsbFormats>>
```

oddhack:
I got tired with adding explanatory comments, but there are what appear to be a bunch of minor markup errors in the new VUs which I have suggested fixes for.

oddhack:
(Sorry, accidentally clicked the comment & close button).

chapters/resources.adoc
rmader: Done

chapters/formats.adoc
rmader: So this required me to move the
```
.Valid Usage
****
```
block outside of the `ifdef::VK_EXT_ycbcr_2plane_444_formats[]` block. There don't seem to be many cases where the VU list could potentially be empty / every item is guarded by an `ifdef`, however there's one such case here: https://github.com/KhronosGroup/Vulkan-Docs/blob/main/chapters/features.adoc?plain=1#L1893-L1907

So I hope this is ok here as well.

xml/vk.xml
rmader: Ok, added (and use) `VK_FORMAT_Z6R10_UNORM`, `VK_FORMAT_Z4R12_UNORM` and `VK_FORMAT_Z2R14_UNORM` now.

See discussion below regarding `Z` vs `X`.

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
rmader: Did this for now (`X` -> `Z`).

oddhack:
Github is borked ATM (presumably because it is a day ending in 'y') and won't let me comment on the diff for some reason, but features.adoc line 1845 introduced some leading white space before the '    ifdef::' which should be removed.

rmader:
> Github is borked ATM (presumably because it is a day ending in 'y') and won't let me comment on the diff for some reason, but features.adoc line 1845 introduced some leading white space before the ' ifdef::' which should be removed.

Whops, fixed. Also should have addressed all other issues from the last CI run (and ran the corresponding scripts locally to validate that)

rmader:
Sorry for the noise, just some more minor fixes.

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
fluppeteer: Sorry for the delay - I tried to write this a couple of weeks ago and GitHub ate it.

I don't love Z as a choice for the same depth-related reasons as @gfxstrand (and arguably should we decide to use WXYZ in a format at any point), but it may be the least bad option - N for "null" may be an alternative, but that may conflict with the common writing of "nnnn" to mean a series of bits. As you say, this feels like a 16-bit format with a different decode path for scaling, since we're saying the "must be zero" bits are actually significant and just expected to be zero; I'm not 100% clear that our hardware, at least, would be able to support this, since our hardware matrix transform has limited inputs. At least we avoided using Z for depth. (Fortunately a recent KDFS update can express mandatory 0/1 bits, I hope as is needed here.)

For reference it would be nice to include these formats in the 56.1 descriptions. C.f.:

VK_FORMAT_G10X6_B10X6R10X6_2PLANE_420_UNORM_3PACK16 specifies an unsigned normalized multi-planar format that has a 10-bit G component in the top 10 bits of each 16-bit word of plane 0, and a two-component, 32-bit BR plane 1 consisting of a 10-bit B component in the top 10 bits of the word in bytes 0..1, and a 10-bit R component in the top 10 bits of the word in bytes 2..3, with the bottom 6 bits of each word unused. The horizontal and vertical dimensions of the BR plane are halved relative to the image dimensions, and each R and B value is shared with the G components for which ⌊iG×0.5⌋=[iB]=[iR] and ⌊jG×0.5⌋=[jB]=[jR]. The location of each plane when this image is in linear layout can be determined via vkGetImageSubresourceLayout, using VK_IMAGE_ASPECT_PLANE_0_BIT for the G plane, and VK_IMAGE_ASPECT_PLANE_1_BIT for the BR plane. This format only supports images with a width and height that is a multiple of two.

I don't think I see that in this PR? (Unless they're autogenerated these days.)

fluppeteer (commented):
Feedback from two weeks ago...

proposals/VK_EXT_ycbcr_3plane_16bit_lsb_formats.adoc
rmader: > I don't think I see that in this PR? (Unless they're autogenerated these days.)

Ouch, I'm sorry - this was indeed missing and arguably a very important part :sweat_smile: 

[Added that now](https://github.com/KhronosGroup/Vulkan-Docs/compare/f5bec6c23f2f6736ea3bf41156a73a5c94ce9103..143fe5fe7f4efc130bea3fc2342d522e706805eb).

> I don't love Z as a choice for the same depth-related reasons as @gfxstrand (and arguably should we decide to use WXYZ in a format at any point), but it may be the least bad option

Thanks! IIUC that means unless somebody else voices resistance we can go forward with that.

> I'm not 100% clear that our hardware, at least, would be able to support this

For the record: even if the formats can only be supported via a shader fallback path it would still be a useful for multimedia apps, freeing them from having to carry custom shaders themselves.

chapters/formats.adoc
rmader: Taking the freedom to close this.

xml/vk.xml
```diff
@@ -33505,6 +33525,87 @@ endif::VK_KHR_internally_synchronized_queues[]
...
+        <format name="VK_FORMAT_Z6R10_UNORM_EXT" class="16-bit" blockSize="2" texelsPerBlock="1">
+            <component name="R" bits="10" numericFormat="UNORM"/>
+        </format>
+        <format name="VK_FORMAT_Z4R12_UNORM_EXT" class="16-bit" blockSize="2" texelsPerBlock="1">
+            <component name="R" bits="12" numericFormat="UNORM"/>
+        </format>
+        <format name="VK_FORMAT_Z2R14_UNORM_EXT" class="16-bit" blockSize="2" texelsPerBlock="1">
+            <component name="R" bits="14" numericFormat="UNORM"/>
+        </format>
+        <format name="VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_420_UNORM_3PACK16_EXT" class="10-bit 3-plane LSB" blockSize="6" texelsPerBlock="1" chroma="420">
```
rmader: One open question from my side: is it ok if all three 10-bit formats share `class="10-bit 3-plane LSB"` - or do they need to have a own class each, with the chroma included/repeated?

spencer-lunarg: These `class` are just here for VUs like 

image

(you can search `vkuFormatCompatibilityClass` in VVL to find them all)

I think the safest thing is to just make a new class, unless you know there are going to be CTS to make sure you can go between other formats

rmader: I see, thanks! I added the corresponding chromas to the class names now, following the example of other similar formats. I.e. each format has a dedicated class now.

rmader:
I took the freedom to re-request reviews, hope that's ok :see_no_evil: 

To quickly re-iterate why I'd love to see this land and therefor be very thankful for further reviews/acks: at least on Linux + Mesa the corresponding GL and DRM formats are already used by default by e.g. Gstreamer + GTK4 based video apps, including the default Gnome video player and various other apps in the ecosystem.

Having proper format representations and driver support allows better performance for software decoded video on most devices with unified memory. Most importantly: if buffers used by the software decoder are allocated as dmabuf (via udmabuf or dma heaps), the buffers can be imported and used by the GPU directly (via driver-internal shaders).

Not having these Vulkan formats currently forces GTK4 - which uses Vulkan by default - to fall back to a slower and overall suboptimal GLES path - and I expect other multimedia apps planning to switch to VK to run into similar issues going forward.

tl;dr: landing this extension would help close a feature gap between VK and GLES for multimedia apps (at least in the FDO / Linux world) :)

fluppeteer:
Sorry (again) for the delay. The updated descriptions look fine - thank you.

If I'm hesitant it's only because we kind of have a Vulkan policy not to introduce features that developers couldn't implement just as well themselves. In this case, that means someone can actually support this efficiently in hardware, especially since the fall-back path (especially with sampling) is quite painful code, and developers might want to short-circuit it (and may also want to be precise about chroma resampling choices). Unless the IHV's software emulation is guaranteed to be as fast as anything the developer could do (and it's added just to make the ecosystem more consistent), we tend not to like adding software fallbacks; it means developers get unexpected performance impacts because they're expecting all format support to be fairly similar in performance. I've not thought through the interactions where some paths are hardware accelerated and others aren't (when it comes to things like dynamic state), but I suspect there may be dragons.

I say that partly because I've confirmed that we (Imagination) would have a problem: we have specific hardware for pulling out the 10- and 12-bit subsets, and our colour transform matrices don't use a conventional float encoding - we can't just scale the values up in the matrix. For us, if we supported this, it _would_ be a software fallback, at least for the foreseeable future (and, as an IP vendor, our design sales have a very long tail). That probably means developers who care about our hardware would need to implement the fallback themselves anyway (or someone could implement a layer to catch it). I'm not sure how much that helps - if it's a GTK thing, the workaround might be possible (and necessary) in Vulkan orthogonally to the existence off this extension.

I'd welcome comments from other IHVs, though. Obviously if everyone else can do this efficiently in hardware and we're an outlier, that's different from it being a special code path for everyone.

oddhack:
FYI the "CTS framework tests" CI failure is spurious - there's an upstream problem in the repo that script is pulled from which is getting dealt with separately.

rmader:
> If I'm hesitant it's only because we kind of have a Vulkan policy not to introduce features that developers couldn't implement just as well themselves. In this case, that means someone can actually support this efficiently in hardware, especially since the fall-back path (especially with sampling) is quite painful code, and developers might want to short-circuit it (and may also want to be precise about chroma resampling choices). Unless the IHV's software emulation is guaranteed to be as fast as anything the developer could do (and it's added just to make the ecosystem more consistent), we tend not to like adding software fallbacks; it means developers get unexpected performance impacts because they're expecting all format support to be fairly similar in performance. I've not thought through the interactions where some paths are hardware accelerated and others aren't (when it comes to things like dynamic state), but I suspect there may be dragons.

Thanks, this is a very good point. A couple of comments

1. I have not yet confirmed that any existing GPU hardware can accelerate any of these formats further than what is already possible with shaders within an app or in a fallback path. There are no hardware implementations for the corresponding GL formats yet, even though they also only got added relatively recently and there might be yet-unknown potential. In that regard your hesitation is definitely justified.
2. What *is* known is that certain display engines support some of these formats. Notably the format used for usual HDR10 videos (`VK_FORMAT_Z6G10_Z6B10_Z6R10_3PLANE_420_UNORM_3PACK16_EXT` which maps to `DRM_FORMAT_S010` on Linux) is supported on e.g. the Raspberry Pi 5. That means that at least in the Wayland ecosystem having the format supported by the driver via a fallback path has a big practical advantage for both apps and compositors: they can easily advertise support for the format (in the best case archiving a zero-copy path from CPU to display engine without using the GPU at all, otherwise using the driver fallback) without having to duplicate the corresponding shaders between every app/toolkit (GTK4, QT, Chromium etc.) and Wayland compositor.
3. Even without display engine support it is often preferable to offload video composition from apps to system/Wayland compositors for performance reasons. Doing so reduces the worst case number of composition steps from two - one in the app, one in the compositor - to just the later. Therefor widespread support for these formats in compositors is desirable on most/all hardware - and outside of the control of individual apps. Many Wayland compositors (e.g. Mutter, Sway, Weston) already support the formats using GL(ES). Without this new extension each of them will need custom shaders equivalent to what would otherwise live in the driver - which in case of Mesa is shared between all platforms and, as of recently, even between Vulkan and GL.
4. To my knowledge existing software fallbacks - say the one for NV12 in Mesa - are considered quite fast / close to what's possible. I might be missing something, but I'm not aware of any existing apps preferring custom NV12 shaders for performance reasons. For example the shaders in Gstreamer and Firefox are essentially copies of what is found in Mesa and mostly exist for compatibility/historical reasons. Some apps like `mpv` apparently prefer their own, slower shaders in order to have tighter control over quality, however.

tl;dr: even without hardware support this extension can provide non-trivial quality-of-life improvements for multimedia ecosystems - for the increasingly common use-case of HDR video playback. IMHO that justifies the extension. Nonetheless I agree that it falls into a grey zone until at least one of the formats can be further accelerated on at least one existing GPU.

rmader:
P.S. regarding
> That probably means developers who care about our hardware would need to implement the fallback themselves anyway (or someone could implement a layer to catch it). I'm not sure how much that helps - if it's a GTK thing, the workaround might be possible (and necessary) in Vulkan orthogonally to the existence off this extension.

In case of GTK the plan is to rely on either having the driver handle the formats (which would be the case for all Mesa Vulkan drivers going forward) or fall back to either GL or software conversion, resulting in lower performance. I.e. the GTK devs would strongly prefer to not have to introduce YCbCr->RGB shaders in their Vulkan backend. I expect the same to apply to Chromium and similar apps.

oddhack:

rmader:
Rebased and ran `normalalize_xml.py` - no other changes.

rmader:
Hi - gentle ping here (sorry for the noise). IIUC the main concern raised is that so far it's unclear which GPUs are actually able to provide acceleration that goes further than shader emulation. 

Apart from that question, does anyone have further concerns so far - and does anyone see the question above as a strong blocker (instead of a question of weighting practical benefits vs the desire to keep the spec clean)?

It'd be super helpful to me to know if there's strong resistance here, whether I should try to better explain my point of why I think it's worth landing - or if we should just give it some more time for people to have another look :)

rmader:
Hi - sorry for the noise, however I'd be really happy if we could get to a decisions here - preferably to land it, of course ;)

To reiterate:
1. This extensions is only about optional formats - nothing in it is mandatory.
2. Most of the formats are well established in software de-/encoders and have been used for many years.
3. There is an existing ecosystem using these formats with GL (Mesa/Wayland/DRM) for which not having these formats in Vulkan is a regression / additional hurdle.
4. There is existing hardware supporting some of these formats - even though so far we only know about display engines, not GPUs yet.

IMHO this should be sufficient reasons to allow the extension to exists - is there any clear resistance from anyone against it? If not, is there anyone allowed and willing to make the decision? :grimacing: 

Best regards

fluppeteer:
Sorry yet again for the delay. While, as I said, there's historically been a bit of a tendency to be wary of adding features to Vulkan that _nobody's_ hardware can accelerate (that is, anything that the user could do as well as the IHVs), I buy the argument that display subsystems may support it - and of course it may help to make Vulkan "a joy to use".

While it would be inconvenient for Imagination to support this because of the need to handle both a hardware and software decode path, nothing says we have to expose an EXT (other than market pressures), so I'm certainly not going to stand in the way (any more than I already have by being slow).

I say go for it unless anyone else raises major objections. :-)

r-potter:
Apologies for the delay in answering on hardware support/EXT vs. MESA status. This has appeared on the Working Group agenda several times recently. This topic has generated a range of views on the tradeoff between developer ease of use vs. only exposing what exists as hardware. The principle blocker has just been getting champions of both positions present in the same meeting during the summer vacation period. We plan to try again next week.

oddhack:
>  We plan to try again next week.

@rmader headsup - we discussed today but did not reach a conclusion, yet. I believe we're waiting on someone who just got back from vacation to decide on the EXT vs. MESA.

r-potter:
The WG discussed at length today. We concluded that EXT is fine for this with MESA plus one additional implementation expected. As usual with EXTs, it's not anticipated that all vendors will implement.

For next steps, @gfxstrand would like to do another round of review. We will also need CTS tests, which I believe @gfxstrand planned to discuss with @rmader.

zlatinski (commented):
The additional formats look acceptable from NVIDIA’s perspective. These newly proposed formats would not be used directly by the hardware for sampling or encoding/decoding operations; however, they could be used with compute filters and transfer operations, as outlined below.

rmader:
Short update from my side: @gfxstrand asked me to look into writing tests and expanding the proposal statement a bit. Unfortunately I didn't get around to that in the last weeks and now I'll be AFK  for two month - I plan to get to back this ASAP after I'm back.

PR #2720, opened 2026-04-26
Change links in proposals to fix errors and be easier to convert for inclusion in the docs site (internal issue 4795)

oddhack:
- Update template.adoc with the new style
- Use 'xref:' instead of 'link:' for targets hosted in the Vulkan Docs site (APIs, extensions, and other spec anchors)
- Add '{refpages}' attribute and link to API refpages in the Vulkan Docs site, rather than into the monolithic HTML spec extension appendix
- Fix numerous typos that resulted in dead links
- Add CI check for old-style links
- Tweak Antora conversion script to rewrite xrefs into Antora conventions when building the site

In addition to manual review, I generated independent HTML versions of the proposals locally (using 'make proposals') and ran a link checker over them, which did not report broken links.


oddhack:
@tobski you might take a look at the descriptor_heap proposal doc because it's the one that had the most minor typos affecting crosslinks. The fixes were obvious and got bundled in with the change from linking to anchors within the spec to linking to the corresponding refpage, and I did run through quickly and verify a bunch of them in the github rendering of the proposal as well as the docs site build.

Tobski (approved):
LGTM as best I can tell.

PR #2763, opened 2026-06-09
Mark superseded VK_KHR_surface commands as legacy

HildarTheDorf:
Similar logic to legacy-gpdp2, only for the VK_KHR_surface extension.

(I started off wanting to write a bpa check for it, but since the legacy checks are auto-generated from vk.xml, this seems the correct place)

### DIFF
--- a/appendices/legacy.adoc
+++ b/appendices/legacy.adoc
@@ -98,6 +98,25 @@ provided in the pname:pNext chain of slink:VkDeviceCreateInfo instead of
 using slink:VkDeviceCreateInfo::pname:pEnabledFeatures.
 endif::VK_BASE_VERSION_1_1[]
 
+ifdef::VK_KHR_get_surface_capabilities2[]
+[[legacy-gpdsc2]]
+=== Physical Device Surface Queries: Superseded via version 2
+
+Some parts of apiext:VK_KHR_surface were superseded by
+apiext:VK_KHR_get_surface_capabilities2, which introduced new versions
+of some physical device surface query functions.
+These provide the same functionality as the VK_KHR_surface functionality
+but with greater extensibility.
+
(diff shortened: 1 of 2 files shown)


HildarTheDorf:
I'm guessing EXT should also be obsoleted by KHR, but that feels like something that should be done in a big "mark all EXTs that have KHR variants as legacy" PR?

oddhack:
@cubanismo this is superficially an SI PR, but the underlying obsoletion issue can be brought up on the maintenance call.

cubanismo:
> I'm guessing EXT should also be obsoleted by KHR

You mean vkGetPhysicalDeviceSurfaceCapabilities2EXT()? vkGetPhysicalDeviceSurfaceCapabilities2KHR() is not a promoted version of that function, and the EXT version exposes functionality the KHR does not, so it should not deprecate it.

VK_KHR_get_surface_capabilities isn't a real extension name, so you should probably pick a different title for the commit/PR, but otherwise, I think this looks OK.

r-potter:
2026-06-10 maintenance call: technically looks fine, but legacy/deprecation changes require main WG approval. Scheduled for next meeting.

HildarTheDorf:
Fixed: Spelling error, PR/commit title
Unsure if this fail is related or not: https://github.com/KhronosGroup/Vulkan-Docs/actions/runs/27239585222/job/80462982256

oddhack:
> Fixed: Spelling error, PR/commit title Unsure if this fail is related or not: https://github.com/KhronosGroup/Vulkan-Docs/actions/runs/27239585222/job/80462982256

It is not related. There's an error in a script in the repository that test comes from, and it is affecting all our CI runs until it's fixed.

PR #2785, opened 2026-08-08
Remove redundant `len="1"` attribute

holtowd:
vkGetDeviceSubpassShadingMaxWorkgroupSizeHUAWEI::pMaxWorkgroupSize used `len="1"` attribute and was the only place where you could encounter such value. 1 is allowed value but itself is redundant and as mentioned in `registery.adoc` should be used for nested pointers like `len="<outer_len>,1` for array of pointers to single values

I would ping HUAWEI. But I don't know how 😅


oddhack:
Agreed. This does not result in any changes to generated files aside from not pluralizing VUID-vkGetDeviceSubpassShadingMaxWorkgroupSizeHUAWEI-pMaxWorkgroupSize-parameter, and is consistent with the text describing the parameter, so accepting as editorial.

PR #2787, opened 2026-08-11
Improved VK_EXT_descriptor_heap align limits

mematthias:
The alignment limits must be less than or equal to the corresponding maximum size limits.
This is now represented by `min`.

### DIFF
--- a/chapters/limits.adoc
+++ b/chapters/limits.adoc
@@ -7488,9 +7488,9 @@ ifdef::VK_EXT_descriptor_heap[]
 | pname:samplerDescriptorSize                 | - | 32                                                         | max
 | pname:imageDescriptorSize                   | - | 64                                                         | max
 | pname:bufferDescriptorSize                  | - | 128                                                        | max
-| pname:samplerDescriptorAlignment            | - | 32                                                         | max
-| pname:imageDescriptorAlignment              | - | 64                                                         | max
-| pname:bufferDescriptorAlignment             | - | 128                                                        | max
+| pname:samplerDescriptorAlignment            | - | [eq]#min(32, pname:samplerDescriptorSize)#                 | max
+| pname:imageDescriptorAlignment              | - | [eq]#min(64, pname:imageDescriptorSize)#                   | max
+| pname:bufferDescriptorAlignment             | - | [eq]#min(128, pname:bufferDescriptorSize)#                 | max
 | pname:maxPushDataSize                       | - | 256                                                        | min
 | pname:maxDescriptorHeapEmbeddedSamplers     | - | 2032                                                       | min
 | pname:samplerYcbcrConversionCount           | - | 3                                                          | max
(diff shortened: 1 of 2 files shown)


oddhack:
Could you identify the spec language behind your comment "The alignment limits must be less than or equal to the corresponding maximum size limits."? We weren't able to resolve this in real time on the teleconference today and people are not sure about this change.

spencer-lunarg:
@mematthias for context, it is allowed for drivers to have a `bufferDescriptorSize` of 24 but have an alignment of `32` (and those 8 bytes can contain extra stuff)

It is already invalid in the spec to have that flipped, where the alignment is less than the size

mematthias:
The fields of `VkPhysicalDeviceDescriptorHeapPropertiesEXT` specify the constraints:
* `samplerDescriptorSize` specifies the size of descriptors returned by `vkWriteSamplerDescriptorsEXT`. Must be a power-of-two value.
* `imageDescriptorSize` specifies the maximum size of descriptors for an image or texel buffer written by `vkWriteResourceDescriptorsEXT`. Must be a power-of-two value.
* `bufferDescriptorSize` specifies the maximum size of descriptors for an address range written by `vkWriteResourceDescriptorsEXT`. Must be a power-of-two value.
* `samplerDescriptorAlignment` indicates the required alignment of sampler descriptors within a sampler heap. It must be a power-of-two value, and less than or equal to `samplerDescriptorSize`.
* `imageDescriptorAlignment` indicates the required alignment of image and texel buffer descriptors within a resource heap. It must be a power-of-two value, and less than or equal to `imageDescriptorSize`.
* `bufferDescriptorAlignment` indicates the required alignment of unformatted buffers and acceleration structure descriptors within a resource heap. It must be a power-of-two value, and less than or equal to `bufferDescriptorSize`.

@spencer-lunarg According to the constraints, `bufferDescriptorSize` cannot be 24 bytes because it is not a power-of-two.

If the `alignment` value is allowed to be greater than the `size` value, the description must be changed accordingly.

spencer-lunarg:
> @spencer-lunarg According to the constraints, bufferDescriptorSize cannot be 24 bytes because it is not a power-of-two.

Sorry, you are right, but the "precise" size from `vkGetPhysicalDeviceDescriptorSizeEXT` could be 24

spencer-lunarg:
@HansKristian-Work I think i messed this up from what you said on the call, the spec does say `bufferDescriptorAlignment` must be less than or equal to `bufferDescriptorSize`.

mematthias:
Given the constraints, it could even be shortened further; so that it is no longer `min(<N>, size)`, but simply `size`.

HansKristian-Work:
Right, it's the individual descriptor types as queried by GetDescriptorSizeEXT that can return "odd" sizes ... Missed the requirement that bufferDescriptorSize and friends had to be POT. The PR LGTM after more spec review.

HansKristian-Work:
> Given the constraints, it could even be shortened further; so that it is no longer min(<N>, size), but simply size.

I think the min makes sense since it makes the spec constraint clearer.

spencer-lunarg:
followup - we updated the CTS to match these limit and everyone passed

PR #2789, opened 2026-08-12
Fix xref to <<SetAndBindingDecorations>>

oddhack:
This anchor is in the proposal document, not the specification. I'm not sure if this replacement xref is the right place, however.

@Tobski PTAL. I'd like to get this sorted before the next spec update.


Tobski (approved):
LGTM

PR #2791, opened 2026-08-15
Misc and small python improvements

holtowd:
Tbh, I am not even sure it deserves any kind of review so feel to close right away without hesitation.

For me it sped up parsing of the registry `from vulkan_object import get_vulkan_object;vk = get_vulkan_object()` from ~\~840ms to \~810ms~. Probably the least concerning thing of scripts of the codebase... I am stupid, I can't do more (at least for now) and just throwing away is kinda awkward for me so just close it if you want

### DIFF
--- a/scripts/reg.py
+++ b/scripts/reg.py
@@ -272,6 +272,7 @@ def mergeInternalFeatures(tree, apiName):
             tree.remove(internal_feature)
 
 
+
 def stripNonmatchingAPIs(tree, apiName, actuallyDelete = True):
     """Remove tree Elements with 'api' attributes matching apiName.
 
@@ -293,7 +294,7 @@ def stripNonmatchingAPIs(tree, apiName, actuallyDelete = True):
             if apiNameMatch(apiName, api):
                 # Add child to the queue
                 stack.append(child)
-            elif not apiNameMatch(apiName, api):
+            else:
                 # Child does not match requested api. Remove it.
                 if actuallyDelete:
                     parent.remove(child)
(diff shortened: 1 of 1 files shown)


scripts/reg.py
```diff
@@ -1012,26 +1010,20 @@ def addFormatCondition(format_name, feature_name):
             self.addElementInfo(spirv, spirvInfo, 'spirvcapability', self.spirvcapdict)
 
         for format in self.reg.findall('formats/format'):
-            condition = None
+            condition = format_condition.get(format_name)
```
gpx1000: format_name looks defined below this line.  Is that a bug?

scripts/vulkan_object.py
```diff
@@ -10,15 +10,15 @@
 from dataclasses import dataclass, field
 from enum import Enum, auto
 
-@dataclass
+@dataclass(slots=True)
```
spencer-lunarg: I've looked at this in the past, the issue is sometime people want to append extra fields here and make this object mutable, to my knowledge, this will prevent it

I've also done some tests and found the memory usage saved from this was only like 5% (when tried a while ago), the 


> from ~840ms to ~810ms

reflects this and I don't think it is worth breaking the ability to update/add fields (which I know some tools depend on)

holtowd: Ok, thanks

scripts/reg.py
holtowd: Yeah condition is meant to be after

oddhack:
This seems OK to me. I'd want to run some simple generated file comparisons but it's all very straightforward cleanup.

PR #2818, opened 2026-09-21
Fix typos in `memory.adoc` chapter

Rageking8:


chapters/memory.adoc
```diff
@@ -4493,7 +4493,7 @@ imported is being used.
     import, as reported by slink:VkExternalImageFormatProperties or
     slink:VkExternalBufferProperties
   * [[VUID-VkImportScreenBufferInfoQNX-buffer-08967]]
-    pname:buffer is not `NULL`, it must: be a pointer to
+    If pname:buffer is not `NULL`, it must: be a pointer to
```
oddhack:
```suggestion
    If pname:buffer is not `NULL`, it must: be a pointer to a
```

oddhack (changes requested):
LGTM aside from the add-on suggestion. I had to think about the 'vowel sound' of Android for a bit :-)

